PR #170, opened 2015-07-26
Adding LOTW sent and recieved field to the ADIF import

core-dd:
Hope that minor addition works out of the box.

73s de M0INN


magicbug:
Thanks for the patch, looks like Graham (W5ISP) merged :)

PR #179, opened 2016-01-18
Bug updates

teabreakninja:
ADIF log import - logbook_model didn't check for existence of 'freq' before trying to use it.

PHP v5.3 compatibility - qso.php array declarations


magicbug:
Thanks for doing this @teabreakninja

PR #184, opened 2016-02-12
Basic DXCC, bug fixes, and ADIF updload

teabreakninja:
- Basic DXCC chart (work in progress here).
- When adding a new QSO, there is an ajax call to clublog.org to get country. This now uses dxcc info etc.
- Allow upload .adif files (not just .adi) (fldigi uses .adif).
- When performing a QSO search, if you click the callsign for details the result is not opening in modal window as does on main page - Now fixed.
- Show eQSL upload and download status in main log screen.


magicbug:
Merged it up doesn't look like anything that nasty. Few extra features that would be good

- If eQSL details haven't been provided within Cloudlog don't display it on the logbook views
- On one of my QSOs for example this satellite QSO I get the eQSL icons, I don't use eQSL so these won't be set as sent or received.
eqslicons
- DXCC Stats is looking good I see the band rows twice?
dxcc
- Couple of duplicate listings and an extra column (guessing 70cms?) 
dxcctwice
 - I think a little automated script is needed to run through the log and check the callsign against the DXCC and make sure its valid for myself i've probably a few without the field filled in or with it mixed case while I've been developing it over the years.

magicbug:
I've added some code which only shows the eQSL rows if the eQSL username is filled in.

COL_EQSL_QSL_RCVD & COL_EQSL_QSL_SENT are set to "N" even if the user isn't using the eQSL feature so doing a check to make sure its not empty might not be enough.

magicbug:
Remembering with DXCC Bands, that theres also 4m inbetween 6/2 etc which might need taken into account.

teabreakninja:
The point on the username and eQSL - I'd thought about doing that but clean forgot to code it.

The DXCC view is a rough first draft.

I'm nearly done on the 'update' controller. This pulls the cty.xml from clublog and extracts the three sections into new tables. We can use these tables as part of the country/dxcc/cqz lookup. I'll also write a script to update existing log entries with this info.

On the update point - migrations. The migration table entry is currently zero and there's 5 updates in the migration directory. I've created 3 new tables and have them labelled 7,8 & 9 so how can I include them in the next update without a fail due to the existing migrations already complete? A manual update to the database is needed, or start from one again?

magicbug:
No problem on username, added it for both LOTW & eQSL..

I'm thinking the arrows should also be used for standard QSLs rather than the image icons just for the sake of consistency.

Excellent on the update, was going to say if not I can write this as I'd expect it mostly affects my log :) 

Info about migrations is at https://ellislab.com/codeigniter/user-guide/libraries/migration.html

magicbug:
There is a migrate controller.. should be a good starting point.

PR #187, opened 2016-02-15
DXCC awards has links to show stations worked on that band and country

teabreakninja:


magicbug:
Seems to be an issue when I click on the links i get "Unable to load the requested file: awards/dxcc/details.php"

PR #189, opened 2016-02-17
Update dxcc

teabreakninja:
Pulls the clublog DXCC file and inserts all three sections into new tables


magicbug:
Seems to function OK, although I had to create the status.html file on my server due to permissions :+1:  

I'll add a menu option later for "updates" in the Admin drop down later.

teabreakninja:
I had real hassle trying to provide update feedback without doing it the status.html way, this ended up being the easiest option. When I get some free time I'll revisit and see if I can do it a cleaner way.

magicbug:
No problem, was easily adjusted with a file and permissions :-)

On 18 February 2016 at 14:49, Andy <notifications@github.com> wrote:

> I had real hassle trying to provide update feedback without doing it the
> status.html way, this ended up being the easiest option. When I get some
> free time I'll revisit and see if I can do it a cleaner way.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/magicbug/Cloudlog/pull/189#issuecomment-185752585>.
>



-- 
Peter Goodhall

magicbug:
Either way thanks for putting some work into the DXCC side of Cloudlog

teabreakninja:
Yup, I'd say so. The format wasn't similar to the clublog download. I'm about to push a update to check the existing log entries and once we're happy with that, we can migrate a table drop of the old ones :D

magicbug:
Before deleted the "Add QSO Logic" might need a tweak to use those new
tables as if Clublog API isn't available it resorts back to using the local
database.

On 22 February 2016 at 08:12, Andy <notifications@github.com> wrote:

> Yup, I'd say so. The format wasn't similar to the clublog download. I'm
> about to push a update to check the existing log entries and once we're
> happy with that, we can migrate a table drop of the old ones :D
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/magicbug/Cloudlog/pull/189#issuecomment-187063715>.
>



-- 
Peter Goodhall

teabreakninja:
Is there any need to use the Clublog API? If there's a way to test the version of the file before downloading, then you can see if an update is needed. The cty.xml has a date in the xml declaration, but this would require you to download the file anyway!

PR #190, opened 2016-02-18
Import more ADIF fields

csete:
While importing the ADIF file you created for me few weeks ago I have noticed that some of the field were not imported, so I made these changes. Feel free to merge it or ignore it :-)


magicbug:
Happy to merge them up, sure someone will find it interesting :+1:

PR #192, opened 2016-02-22
Check existing log entries against the new DXCC/Country list

teabreakninja:
Should really only need to run this the once...


magicbug:
Havent really had time to debug but quick test against my log

A Database Error Occurred
Error Number: 1064

You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IVOIRE', COL_DXCC='428' where COL_PRIMARY_KEY=4145' at line 1

update TABLE_HRD_CONTACTS_V01 set COL_COUNTRY = 'COTE D'IVOIRE', COL_DXCC='428' where COL_PRIMARY_KEY=4145

Filename: /home/peter/websites/logbooks.2e0sql.co.uk/2e0sql/models/logbook_model.php

Line Number: 1004

PR #193, opened 2016-02-22
addslashes to escape country names with quotes :/

teabreakninja:


magicbug:
https://logbooks.2e0sql.co.uk/2e0sql/index.php/awards/dxcc

Just run this against my live log, you'll see theres quite a few discrepancies including callsigns being assigned to the wrong DXCC.

magicbug:
For the sake of testing this is my log might be useful for testing.

[TABLE_HRD_CONTACTS_V01.zip](https://github.com/magicbug/Cloudlog/files/141435/TABLE_HRD_CONTACTS_V01.zip)

teabreakninja:
Seems there were optional start and end dates in the prefix entries of the xml file. I've added these to the schema, updated the import and tested against your log as well as my own. All seems ok.

Useless trivia - San Marino, prefix T7, previously had prefixes 9A and M1. One for the pub quiz ;)

PR #194, opened 2016-02-23
DXCC match on date as well as callsign. Table schema updates

teabreakninja:


teabreakninja:
I've created a stored procedure version of this (to try and improve performance) but it's failing with one entry in your log as there's an error with the ClubLog data: There's two entries for SZ-Greece, both have different start dates, but no end date. I've raised a ticket with ClubLog, so we'll see how that goes:

        <prefix record='7900'>
                <call>SZ</call>
                <entity>GREECE</entity>
                <adif>236</adif>
                <cqz>20</cqz>
                <cont>EU</cont>
                <long>23.7</long>
                <lat>38</lat>
                <start>1972-01-01T00:00:00+00:00</start>
        </prefix>
        <prefix record='7901'>
                <call>SZ</call>
                <entity>GREECE</entity>
                <adif>236</adif>
                <cqz>20</cqz>
                <cont>EU</cont>
                <long>23.7</long>
                <lat>38</lat>
                <start>1971-01-01T00:00:00+00:00</start>
        </prefix>

magicbug:
If the callsigns just "SZ" it will be a duff callsign.

magicbug:
Ah OK, Yes the likes of SZ1A etc should be valid

teabreakninja:
Got word back, there's a new update to the list from ClubLog...

magicbug:
OK Merged, seems better although I can see that some of my contacts have been associated with Czechoslovakia which was which I'm guessing should be Slovakia as I've not had any QSOs before 2002.

Also theres cases of two instances of the DXCC with a different case (https://logbooks.2e0sql.co.uk/2e0sql/index.php/awards/dxcc)

teabreakninja:
OK, give me examples of the contacts and I'll manually work out the DXCC. Also, what's the two instances? (My eyesight must be failing me!)

magicbug:
the one i can find at the moment is French guiana and French Guiana there was another but cant see it for the life of me will check again this eve! :)

PR #197, opened 2016-04-08
Changed the table_hrd_contacts_v01 table name to uppercase in log.sql

DineshCyanam:
Cloudlog is expecting an uppercase table name


magicbug:
Thanks @DineshCyanam merged.

PR #203, opened 2017-02-18
Refactored frequency class

RobertKaldenbach:
I rewrote this class to be more efficient and also removed a duplicate code for the 70cm band.  The logic is to have the mode default to SSB if none is selected. For the 30m band I have added a SSB value that has the same value as CW so that it would default to the same frequency like the previous code.  Let me know if you would like any further changes before being merged into the master.
Thanks!
Robert Kaldenbach
KD2MSR


magicbug:
@Manawyrm @teabreakninja Do you guys have any thoughts on this being merged

teabreakninja:
Seems all good. Much easier on the eye ;)

PR #219, opened 2018-11-25
Move from GoogleMaps to OpenStreetMap

teabreakninja:


magicbug:
@teabreakninja Thanks for this Andy, I certainly think its probably time to separate from Google Maps, I know its been mentioned a few times to me and also removes peoples concerns about paying Google.

@Manawyrm Tobis couldn't do me a massive favour in double checking the maps are all OK before I merge the code base?

Guess next step is to look at different colour pins for bands.

Manawyrm:
Yes, this looks really neat. Was on my Todo-List, too, so this saves me a lot of trouble 😄

I‘ll get the PR running tomorrow morning and report back :)

Manawyrm:
I just tested the PR. :) 

@teabreakninja Looks really good on the front page, but there's some more maps in 
application/views/social/map.php 
application/views/view_log/qso.php
which are still using GMaps... 

It looks neat: 
screenshot of OSM

There seems to be another (unrelated issue) with the map on the frontpage. I think it's supposed to show all contacts, but it's currently only showing one. I think this has worked in the past, so I might have been the one who has broken that... Will report back on that (in another issue/PR). 

After fixing the other 2 map displays, this can be merged I think :)

magicbug:
Map looks good on staging, nicer on the eye than Google's!

magicbug:
Looks fab, shall we move map discussions now I'll merge this.

I have a feeling "social" can actually be removed, it was an early thing I made for an event I was doing just to display contacts, I wanted it to update every x seconds but never did it.

Andy thanks for the hardwork!

PR #224, opened 2018-11-28
Create updates folder

Manawyrm:
Hi,

this PR adds the "updates" folder to the git repository.
Without this folder being present, clicking on the DXCC update button causes a cascade failure, resulting in a crash and an empty DXCC table. If you then import an ADIF file, you completely ruin your database :-) 
Luckily, I had a recent backup at hand... 

EDIT: thinking about it, I added an additional check to the Update controller, to create the folder if it doesn't already exist. Doesn't hurt, but could help. 

Best regards,
Tobias

### DIFF
--- a/application/controllers/Update.php
+++ b/application/controllers/Update.php
@@ -169,6 +169,11 @@ class Update extends CI_Controller {
 		}
 		gzclose($gz);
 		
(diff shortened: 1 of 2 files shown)


magicbug:
Haha, I had that experience first hand

PR #230, opened 2018-12-13
Maps: Add some base maps and a fullscreen control

g0wfv:
In an effort to add some layers so we can display other pins, I at least got basemaps and a fullscreen control working.

I am however struggling to make the current code in leafembed.js  suitable for multiple, selectable pin layers. (well, I can get multiple pin layers working on a standalone map, but I can't get them to integrate into the existing code!)


magicbug:
Feel free to adjust the code to make it work, I'll probably be doing that this weekend as I've now got the code to do grid overly so I'll probably hold back this merge

magicbug:
Quick thing to mention of it's a is file to a library try and keep it local rather than linking to a remote cdn

magicbug:
I'm going to close this for now, while its a good idea, it needs implemented a bit differently to keep already existing core features.

PR #231, opened 2018-12-13
Dashboard: Change map to countries worked

g0wfv:
Keep this one in the back pocket for when we get multiple pin layers working.

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -78,48 +78,34 @@ class Dashboard extends CI_Controller {
 		$mon = date('Y-m-d', $raw);
 		$sun = date('Y-m-d', strtotime('Sunday this week'));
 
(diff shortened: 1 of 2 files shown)


magicbug:
While I know some might prefer the switch I personally prefer to see all my QSOs from the period allowed than just countries.

Thus but reluctant to bin the option think this would be better intergrating into a layer that can be turned on

g0wfv:
Hence "keep it in the back pocket" ...  Just so we have the code to hand one layers are sorted.

Currently though, the code only lends itself to one set of markers, and isn't easily changed to accommodate a multiple layers .....

Will probably need some help to do that one (but if you're tackling grids this weekend ......?)

g0wfv:
Just thinking some more ....

Perhaps this would be better under the statistics page anyway ....

I've since added custom pins that denote differing confirmed status too ....

PR #232, opened 2018-12-14
LoTW: Prevent error auto pull after initial pull

g0wfv:
This temporarily prevents an error (failed to open stream: HTTP request failed! HTTP/1.1 400 Bad Request) from occurting during auto pull of LoTW QSLs.

I will investigate further, but for now the offending part of the URL generation has been commented out; it's probably an easy fix along the lines of date format?


g0wfv:
It was the date format!  Now fixed.

PR #233, opened 2018-12-14
Fixed adding non 'SSB','DATA','CW' qso

teabreakninja:
Was erroring out when trying to add a qso with 'FM' mode. On investigation most modes were failing...


magicbug:
Merged wonder if merging https://github.com/magicbug/Cloudlog/pull/203 caused this issue?

magicbug:
It's worth noting that's for logging purposes changing anything data to mode Data can effect awards

Issue #229, opened 2018-12-07
Add QSO: QSO details remain after deletion of callsign field.


g0wfv:
**Describe the bug**
If you start to enter a QSO by typing in the callsign, certain field will be populated by lookup and you can also add other fields manually.

If you then decide not to log the QSO (you didn;t make the contact perhaps?) and delete the callsign, the previously entered callsign's details (Name and Locator eg.) are still populated and will not update on entry of the next callsign. 

**To Reproduce**
1. Log a QSO with the name field populated.
2. Add a second QSO with the name field populated.
3. Type the first callsign into the callsign field (name field - and others) populated by lookup_.
4. Remove the callsign from the callsign field.
5. Type the second callsign into the callsign field (previously lookup data persists).

**Expected behavior**
The QSO data fields (except static data not pertaining to an individual station) should be removed on deletion of the callsign field as though you were adding a "fresh" QSO.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser Chrome
 - Version 70.0.3538.77 (Official Build) (64-bit)

**Additional context**
I'm looking into it ... (bare with me might not be within my skills to sort out the logic in PHP!)

magicbug:
Indeed, although pressing reset also clears all fields, could use the JavaScript keyup and watch it the input does go fully blank

g0wfv:
Javascript, PHP, it's all "new" to me ..... I am learning though :-)

g0wfv:
... but not quick enough!

Spent the last few hours scratching my head over this one (on and off!)

keyup does indeed seem to be the way to go (after introducing a delay to wait for typing to stop else you get way too many db lookups when they're not required!)

However, I just can't get the QSO list to update back to the full list when the callsign input goes empty (ie with no filter) despite trying various ways to pass the '%' catch all through to Logbook.php or other workaround hacks.

magicbug:
If I remember right, the last 16 qsos was hard coded and not loaded by JS, so it would need a similar function to the search to make it return.

g0wfv:
I'm making some progress now.

Trying to pass a % back through was a red herring (and possibly being stopped by the webserver for security reasons anyway?)

[...]

PR #235, opened 2018-12-16
Change from separate files to add post and live QSOs

g0wfv:
This PR moves QSO logging (live and post) to a single PHP file whilst maintaining current functionality through 2 menu items.  (I think in the two file version, the files had become divergent and post was lacking some of the updates that had been made to live (and possibly vice versa?)

This also addresses #229 to some degree (although I'm sure the logic can be improved on!)

I encountered merge conflicts whilst trying to apply @teabreakninja's #233, so I applied it manually, please check it's correct (it seems to be working here at least!)

Overall, please check this entire PR before applying to maser (I'm sure you will!); I've notbeen able to test some of the functions like CAT radio data during live logging (I am running on Linux and don't use the software it references!)  My approach may be a little simplistic, hence the need to test it first!

Other than that, the actual adding of QSOs seems to be fine, again, not extensively tested, but tested none-the-less!


g0wfv:
Ugh! Conflicts!

PR #236, opened 2018-12-16
Change from separate files to add post and live QSOs (in lieu of #235 which had conflicts)

g0wfv:
OK, I don't know how this will go; I've gone about compiling my commits in a different way ....

Rather than add all my commits (which GitHub seems to take cumulatively rather than sequentially which created a conflict along the way), I've reverted @teabreakninja's #233 to take the conflicting file back to a stage comparable to before I started making my changes, and then added my commits, reincorporating his change manually along the way .....

This PR moves QSO logging (live and post) to a single PHP file whilst maintaining current functionality through 2 menu items. (I think in the two file version, the files had become divergent and post was lacking some of the updates that had been made to live (and possibly vice versa?)

This also addresses #229 to some degree (although I'm sure the logic can be improved on!)

Overall, please check this entire PR before applying to maser (I'm sure you will!); I've not been able to test some of the functions like CAT radio data during live logging (I am running on Linux and don't use the software it references!) My approach may be a little simplistic, hence the need to test it first!

Other than that, the actual adding of QSOs seems to be fine, again, not extensively tested, but tested none-the-less!

Please check!


magicbug:
Will check it in the week, I'm planning to start cat app for Linux now, be Hamlib and ultimately support unlimited radios.

Idea being could connect all your rigs to a raspberry pi.

That said easy to send dummy data

g0wfv:
hamlib is the way to go .... via rigctld :-)

It wouldn't even need a dedicated app or everything connected to a dedicated raspberry pi; it could be accomplished by a PHP socket call to rigctld (running wherever, even on localhost!) to query the mode/freq as appropriate.  This would cover all setups both remote with rigs connected to a dedicated raspberry pi or locally connected to a Linux host (which is how I'd run BTW!)

I'm have no experience under Windows, but there is a version available from Sourceforge ([https://sourceforge.net/projects/hamlib/files/hamlib/3.3/](https://sourceforge.net/projects/hamlib/files/hamlib/3.3/)) which may mitigate the need for Omni-Rig (you could always support both?!)

Manawyrm:
> PHP socket call to rigctld
hm, yes, but I'd rather not have my rigctld open to the internet. Ideally the connection would be the other way around, e.g. a small local client program polls the data from rigctld and posts them to Cloudlog via HTTPS POST (or similar). 

I've hacked together a small PHP script that does this for my usage with DF0YOTA the other day in order to properly log accurate frequencies. I'm running that script locally on my Linux workstation, Cloudlog is running on my webhosting server in the datacenter.

magicbug:
Excellent!

magicbug:
Are you by chance prepared to share it, just so I can include as an example for now.

g0wfv:
Ahh, see?  I hadn't counted on folks hosting their Cloudlog instance off-site!  What we need is something that gets the frequency from the client side to rigctld  (assuming both are within the same LAN!) rather than server-side .....

magicbug:
I've made this an issue https://github.com/magicbug/Cloudlog/issues/238 rather than discuss it on this PR.

magicbug:
As this requires a bit of testing, I'll hold off till later in the week to test it all as need to make sure radio interfacing works etc.

Manawyrm:
I just installed this on my staging instance and it works nicely for me. 
Radio interfacing works fine (as far as I can tell, only tested with my implementation :smile: )

A bit of real nitpicking: 
```
.input_time {
    width: 36px !important;
}
```
looks nicer on my machine. with 32px part of the time is cut off.

magicbug:
Just tested it my side and works OK with all my CAT and SatPC32 interfaces so i'm happy to merge it

PR #240, opened 2018-12-17
Fix the broken link and add some shameless self promotion :)

Manawyrm:
The Link to the cloudlog-rigctl-interface was the wrong one... Also I've added my homepage to the list on the bottom, hope that's ok :)


magicbug:
Perfectly fine, I didn't know what it was to include!

PR #252, opened 2019-02-25
Changed charset to utf8mb4

Manawyrm:
This changes the database charset to multi-byte capable UTF8. Some tables were latin1, some were (2-byte) UTF-8. 
This allows for special characters in the notes and log. 
I had an issue with a russian OM, whose name would be corrupted and not saved properly. 

As this change is handled by MySQL internally, we can just switch the charset without any "special" work. The HTML output of Cloudlog is UTF8, already anyway.


magicbug:
Seems reasonable, I need to see what the merge conflict is before I can merge it later it's moaning about station_profile. also, I've created a change log at [https://github.com/magicbug/Cloudlog/wiki/Change-Log](https://github.com/magicbug/Cloudlog/wiki/Change-Log) to try and make it easier to track what's happening or changed. especially for older installers.

Manawyrm:
Sorry, that's my fault. I'll fix this!

PR #256, opened 2019-02-25
Fixed default frequencies

Manawyrm:
These frequencies were all missing a zero and were one order of magnitude too small.


magicbug:
Thanks for sorting those, just shows no one was on microwaves :)

Issue #249, opened 2019-02-24
Discussion: satellite/relay frequency offsets


Manawyrm:
While talking to some other OMs on the new satellite, my online logbook came into discussion. 
The frequency input I had done was talked about and someone brought up that it is custom to log the uplink/TX frequency in the logbook. 
This seems to be reinforced by the existence of the COL_FREQ_RX column in the database, which is currently unused. 

With repeaters you often have the -7.6MHz offset, in case of the es'hail it's a +8089,5MHz offset.. 
Any ideas on how/if this could be integrated into the UI? 
Either seperate frequency input fields (those could be filled by the API/CAT integration), or (crude!) maybe just a list of birds, if "sat_name" == es'hail-2 -> offset +8089,5M. 

Is that a useful feature in general? How do you log sat/repeater contacts?

magicbug:
I actually send it on my app from SatPC32 it's stored in the database just need to add it to the frontend

magicbug:
In the QSO area under "Station" there's the frequency input box, I'm proposing we add Frequency RX below to store the RX this can then be populated manually or by using the API cat system.

![capture](https://user-images.githubusercontent.com/84308/53349695-947fa300-3915-11e9-92df-b71de32d2421.PNG)

Manawyrm:
Good idea. I think I'll have a go at that this evening.

magicbug:
Think in the database for CAT/Radios the downlink is stored in a field called "downlink" or something close to it! Just need a little extra js to add it.

magicbug:
While thinking about it, I can't remember if there is also fields for band and mode for RX however perhaps should populate those

Manawyrm:
Ok, so there's a field for RX band, not for RX mode. That probably makes sense :) 
I'll implemented that later.

Another thing I'd like to add is a better frequency parser for the input field.
"3700" -> 3.700 MHz 
"2350210000" -> 2.35021 GHz
"2.350e6" -> 2.35 GHz
"3.7M" -> 3.700 MHz
What do you think about that? Should be easy to do and not break any existing mechanisms...

magicbug:
Yes would like to show the frequencies better a long number is a bit overwhelming to most, it would also look nicer on the display.

magicbug:
Merged the changes they look great, also note the change to big int, to handle the larger numbers.

[...]

PR #257, opened 2019-02-25
Add downlink frequency handling

Manawyrm:
This PR does a lot of things at once: 

Implements #249 (not perfect yet, will do another PR later) 
Remove the old "manual" QSO interface, that's already handled by the same code as the live QSOs
Remember manually entered frequencies for new QSOs, as long as no band/mode switch was done
Fix the database layout for the CAT table to also allow for the 3cm band frequencies


magicbug:
Excellent, guessing you didn't spot any major issues? agree that manual function needed to be removed, also makes things a bit neater long term.

Manawyrm:
No, it all worked pretty well. I have to say, the Cloudlog codebase is fun to work on. I'll have to look into the Codeigniter framework for my private projects as well, it's really nice. :)

magicbug:
Its one of my favourite frameworks as it keeps out the way, I use it a lot for private stuff and commercial 👍

PR #258, opened 2019-02-25
Request CAT information as JSON

Manawyrm:
Here's already the next PR. 
I have changed the way the CAT information is requested from the server. Instead of 4 (well, with the DL frequency 5) seperate HTTP requests every second, now only 1 has to be executed and replies with a JSON array containing the neccesarry information.

Should be faster and neater this way :+1:

### DIFF
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -61,6 +61,18 @@
 			
 	}
 	
+	function json($id)
+	{
+		echo json_encode(array(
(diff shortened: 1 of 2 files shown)


magicbug:
Yay!, if your bored the other lookups for name, qth, locator needs some love, it's rather slow, mostly as for some reason I thought it was a fab idea to check them one by one (stupid)

PR #262, opened 2019-02-25
Cleanup DL/UL handling, add microwave sat modes

Manawyrm:
One last PR for this evening, I'll go to sleep now :) 

This fixes an embarrassing little mixup between UL and DL frequencies, adds all the sat-modes and requests both UL and DL frequency in one DB query. 

alt

### DIFF
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -63,21 +63,18 @@
 	
 	function json($id)
 	{
+		$frequency = $this->frequency($id); 
+
 		echo json_encode(array(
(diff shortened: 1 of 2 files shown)


magicbug:
Haha, I'd recommend storing Es'Hail-2 contacts in Cloudlog as QO-100 as that's what ARRL LOTW accepts for uploading QSOs but that's just my personal opinion :)

Manawyrm:
Good point, I was thinking if QO-100 or es'hail was the better choice... This is a good argument :D

Issue #288, opened 2019-06-14
[FL-DIGI] Wrong Frequency on ADIF-Import


dg9vh:
Hi,

when importing following ADIF-file, the Frequency is imported incorrect for example for first entry as 70.41500 that is mapped as 4m QSO.

```
File: export.adi
<ADIF_VER:5>2.2.7
<PROGRAMID:6>fldigi
<PROGRAMVERSION:6>4.1.03
<EOH>
<FREQ:8>7.041500<CALL:6>GI4SZW<MODE:3>PSK<SUBMODE:5>PSK31<NAME:6>Seamus<QSO_DATE:8>20190603<QSO_DATE_OFF:8>20190603<TIME_ON:6>190100<RST_RCVD:3>599<RST_SENT:3>599<EOR>
<FREQ:8>7.044980<CALL:7>GB19CGW<MODE:4>RTTY<NAME:4>Rsgb<QSO_DATE:8>20190603<QSO_DATE_OFF:8>20190603<TIME_ON:6>192800<RST_RCVD:3>599<RST_SENT:3>599<EOR>
<FREQ:8>7.040677<CALL:6>LY2BKQ<MODE:3>PSK<SUBMODE:5>PSK31<NAME:8>Anatolij<QSO_DATE:8>20190610<QSO_DATE_OFF:8>20190610<TIME_ON:6>184500<RST_RCVD:3>599<RST_SENT:3>599<EOR>
<FREQ:8>7.041223<CALL:8>SM5LIB/P<MODE:3>PSK<SUBMODE:5>PSK31<NAME:4>Igge<QSO_DATE:8>20190610<QSO_DATE_OFF:8>20190610<TIME_ON:6>185500<RST_RCVD:3>599<RST_SENT:3>599<EOR>
```

How could this be solved? Should band-data be exported with to get frequency correct?

Kim
DG9VH

magicbug:
Seems we have an issue with the import of the frequency for example WSJT-X 40m - freq is stored as 7074987 so 7.074.987 which cloudlog handles fine but coughing on the FLdigi

This is handled by https://github.com/magicbug/Cloudlog/blob/master/application/models/Logbook_model.php in the import() function I'll take a look tomorrow and try make some adjustments

Peter.

magicbug:
Converting frequency to the band is handled by https://github.com/magicbug/Cloudlog/blob/master/application/libraries/Frequency.php

dg9vh:
[...]

PR #290, opened 2019-06-15
Various tweaks required to import an N1MM+ ADIF

M0LTE:


magicbug:
Thanks Tom will get that checked and merged, need to check the #288 issue next as it seems band isn't right which is odd as I was heavily using fldigi before ft8 etc so it would have been working!

PR #296, opened 2019-06-16
 Added more Band-Slots on DXCC-Statistics

dg9vh:
Microwave-Bands were missing on the statistics-page


magicbug:
Merged, I'm wondering whether some options need to be made available for what you want to track or not show bands where theres no qsos

dg9vh:
Maybe we can modify the statistics in that way, that it shows only used band slots and not all of them... This could be done by creating the header-line dynamically based on a select distinct col_band from %logbook_table% group by col_band maybe?

PR #297, opened 2019-06-17
Make Awards - DXCC to show only worked band-slots

dg9vh:


dg9vh:
As discussed in https://github.com/magicbug/Cloudlog/pull/296 this now only shows slots on DXCC-Awards-page, that really were worked.

PR #298, opened 2019-06-17
removed hardcoded table name

Manawyrm:
The hardcoded table named caused an exception for me

(I really need to get the multiple logbooks stuff mainlined/upstreamed).


Manawyrm:
(also I've now fixed a notice that occures when running PHP with higher debug levels)

PR #300, opened 2019-06-17
Modified Javascript to update RST on mode change via cat-control

dg9vh:
This is a bug-fix for cat-control based setting of correct RST.


magicbug:
@dg9vh Just to be clear that mode change based on the CAT is updating? if so could the js for RSTs not be triggered when that changes?

dg9vh:
Correct. Mode change works...

dg9vh:
No problem, I count on your conscientious working methods :-)

PR #302, opened 2019-06-21
Masking QSO-Time on dashboard with **:** of not logged in

dg9vh:
To avoid third parties to get to much infos on QSOs.

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -80,7 +80,7 @@ $autoload['helper'] = array('url');
 |
 */
 
-$autoload['config'] = array();
+$autoload['config'] = array('cloudlog');
 
 
 /*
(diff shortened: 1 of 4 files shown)


dg9vh:
If you think, it would be better to make this switchable for the user, you could add some preferences within the config?

magicbug:
Yes I like the idea of making it optional, this could be true false activated within the /application/config/cloudlog.php file if you don't know how that works check the CodeIgniter docs

magicbug:
I've also wondered on the dashboard whether we could hide signal reports

dg9vh:
> I've also wondered on the dashboard whether we could hide signal reports

Hi, I made the thing now configurable like the way you wanted it :-)

Hiding the reports is not so necessary I think - hiding the time would prohibit SWLs just to grab QSOs from your log to fetch QSLs cards from you (or your QSO-partner, if it is a rare DXPedition). Reports are not the target of this I think.

magicbug:
I actually did another commit and completely removed the date for not signed in purely to keep the interface looking tidy.

dg9vh:
Hi, no Problem. Looks better now :-)
73 de Kim

PR #306, opened 2019-06-24
Possibility to add DARC-DOK in existing Database-Field COL_DARC_DOK

dg9vh:
For further (award-)tracking possibilities (and because of the fact, that this data is imported into the database via ADIF already), I added the DOK-Field for adding/editing and showing qso-details.


dg9vh:
CR done with some little more (see comment of commit).

PR #308, opened 2019-06-25
replaced 0 with Ø in callsigns

dg9vh:
... to make the Display of callsigns more beautiful and less ambiguous


dg9vh:
I wanted to change the callsigns in the backend, only did a str_replace on the (hopefully all?) places, where it is shown on the front-end...  I would revert the changes in lines 56, 59 and 62 that where done accidently :-) sorry.. I fixed it just now.

dg9vh:
It's on my side :-)

PR #309, opened 2019-06-26
Added new Awards - DOK

dg9vh:
This gives us a list of worked DOKs (counted only, if DXCC is 230 = Germany)


dg9vh:
Thank you :-)

PR #339, opened 2019-08-29
navbar to center, awards table header to sticky

teabreakninja:
Fixed the navbar to center

Made the awards table header sticky, to make viewable throughout scroll

### DIFF
--- a/application/views/interface_assets/header.php
+++ b/application/views/interface_assets/header.php
@@ -22,7 +22,7 @@
   </head>
   <body>
 
-
+<div class="container" id="nav-container">
 <nav class="navbar fixed-top navbar-expand-lg navbar-light bg-light">
 		<a class="navbar-brand" href="<?php echo site_url(); ?>">Cloudlog</a>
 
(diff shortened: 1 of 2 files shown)


dg9vh:
After getting this into production, I recognize a problem with the z-axis and the layout of the nav-menu.

image

as You can see on the screenshot, the map-layer is overlaying the menu-structure. This also happens with other components like Table-headers and so on...

dg9vh:
I think, it has something to do with sections in "card"-css-class.

magicbug:
Yes, also removes the grey background going all the way along, so had to revert the pull for the top elements the table for dxcc heading however functions fine!

dg9vh:
Yes, now it's working ok - maybe we get the nav-bar nice again later :-)

magicbug:
Its actually simple case of the container div needed to be inside the <nav> tags just about to commit the change

magicbug:
Resolved and commited

dg9vh:
looks fine now :-)

PR #340, opened 2019-09-01
fixed logical error on eqsl-selection.

dg9vh:
This fixes the Error based on https://github.com/magicbug/Cloudlog/commit/09a46b2b66ffde8f2b4aeda0fa66e9cbd8870dc5


magicbug:
The reason for this line

`$this->db->or_where('COL_EQSL_QSL_SENT IS NULL');`

was the fact if it was null and it wasn't checked it caused issues. you might want to sanity check with a test.

dg9vh:
I tested my new variant with success - it only delivered new qsos, the one marked as invalid was ignored... so all working perfect here with my variant... your variant before delivered all >10000 QSOs to be sent to eQSL that dropped me with the mentioned error at the commit from you...

PR #348, opened 2019-09-07
Link to the eQSL image in log

myzinsky:
A link to the eQSL card image is inserted to the log. In a future commit, the image will be cached when the button is clicked and then showed from the cache instead of forwarding to the eQSL website.


magicbug:
But yes agree with @dg9vh due to API requirements storing the image locally is a must to stop abuse, plus I think roughly you'd not be opening more than 3 a minute

myzinsky:
Okay, i will change it. I’ll keep updating here

myzinsky:
Instead of the letter symbol we could set the link also to the received arrow? What do you think?

dg9vh:
Think, this is a good idea from ux view.

PR #354, opened 2019-09-27
Qslprint is considering only the QSOs of the active station profile now.

dg9vh:
After making cloudlog able to work with different logs on different station profiles, the print-export-function now only considers the qsos of the active station.


magicbug:
My only suggestion would be to get station_callsign from the station_profile table I'll be shifting all my fields to that table

dg9vh:
Would be no problem, just a join on sql 😊

dg9vh:
I modified it to get callsign from profile...

PR #355, opened 2019-10-02
mark requested paper qsls (send and receive) with yellow maker

dg9vh:
Would look like this:
image


dg9vh:
Effort is to see at first look if you marked a qso requesting qsl or not (for getting control, if there are cards for printing or if you forgot to set the marker right in the qso-logging-tab).

Issue #359, opened 2019-10-05
enlarge COL_MODE from varchar(10) to varchar(12)


dg9vh:
If you take a look at http://www.adif.org/304/ADIF_304.htm#Mode_Enumeration you'll see, that "DIGITALVOICE" is a valid mode in ADIF-Specification with length of 12 chars, COL_MODE only allows varchar(10), so we need 2 additional characters there to produce valid logs in ADIF.

magicbug:
Oddly I hit that issue today I just made a change request on the PR sir once change is done i'll add it.

You might need to add similar for sub_mode

magicbug:
@dg9vh Kim Thanks its now merged :) do you think COL_SUB_MODE needs similar?

magicbug:
Submodes 25 so should be fine! closing this issue

PR #360, opened 2019-10-05
set size from varchar(10) to varchar(12)

dg9vh:
see http://www.adif.org/304/ADIF_304.htm#Mode_Enumeration and https://github.com/magicbug/Cloudlog/issues/359


dg9vh:
Done

magicbug:
Merged :)

PR #368, opened 2019-10-08
Use Club Log CDN for cty.xml

g7vjr:
Change the URL for collecting Club Log's cty.xml definitions to use https://cdn.clublog.org (Content Delivery Network, faster). Remove reference to secure.clublog.org and update string "Clublog" to be "Club Log".

Getting my hands dirty for the first time so very minor stuff!


magicbug:
Nothing wrong with tweaking minor stuff :)

PR #373, opened 2019-10-13
Added indicators, if callsign/dxcc is worked on band and mode

dg9vh:
like the locator-input the callsign-input is colored in red or green
the country-label is enabled with a label that tells us about worked/not wkd

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -221,6 +221,74 @@ class Logbook extends CI_Controller {
 		return;
 	}
 
+	function jsonlookupdxcc($country, $type, $band, $mode) {
+		
+		$return = [
+			"workedBefore" => false,
+		];
+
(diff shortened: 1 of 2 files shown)


magicbug:
I'm guessing this works for satellite too?

dg9vh:
I think - if the locator-code did work before *g* I just copied and pasted it from there and had also the satellite-passages within it... It would be nice, if you could test it on your live-system when active :-)

magicbug:
Seems to work, for New DXCC can we change the label colour so instead of it being grey its something else to match 

You can find css https://getbootstrap.com/docs/4.3/components/badge/

dg9vh:
How about using Success / Danger (green/red)?

dg9vh:
done in https://github.com/magicbug/Cloudlog/pull/374

Issue #358, opened 2019-10-04
Cloudlog 2.0 - Multi-Client, Multi-Logbook, Moving to Database Settings taking things forward


dg9vh:
**Problem**
I am often asked about the client capability of cloudlog. Actual developments are in the direction to have one log that could be filled by different users in the system and/or one user with different station-profiles that is a perfect solution for multiop-Stations like clubstations where one callsign is operated by different users for example.

What users also asked is to have one instance of installation for different users with their own log.

**Solution**
We could create a new field in the logging-table called "client" or similar that refers to one dedicated client instance. So we can manage all log-entries as usual in one logging-table but could filter in general on that client id.

The client-id should be maintained within an own table like station_profile only by administrators. We should also consider about a relation-table client_user that would map a user to one or more clients and vice versa (n:m-relation).

The log-view so should present only those logentries that matches the client-ids that the user is mapped to (but also those QSOs of other users mapped to this client-id).

dm-o:
Yes - would be a very nice feature for our DARC chapter to give every member a personal logbook as a free service.

magicbug:
While I totally get people to want to start offering Cloudlog as a SaaS to others, it wasn't really its fundamental plan at the early stage and to get it to that stage you'd need to consider the following

- Implement a logbook table which assigns to station profiles, logbook being whats assigned to a user.
- Logbook ID in the QSO table
- Table to link Users to Logbooks
- Table to Link Users from their Logbooks to Station Profiles
- eQSL/Clublog/LOTW settings all need split off and stored differently to support "club members" callsigns.
- QRZ/HamQTH logins need to be shifted to the database (This needs done anyway) but encrypted along with the other logbook services.
- Any config item in config/cloudlog.php needs moved to the database and collected on login.

For good database design, it requires a bit more than just a column in the log ID, it will be something that I start looking into, but functions will require a bit more than you imagine.

dg9vh:
Hi Peter,
you are right - my idea was very incomplete and only a first shoot within 5 minutes when typing the issue :-)

I think this is more than a point of coding, more than a strategical decision, what way cloudlog should go.

This decision is not done by my person, I only could give some impulses from feedback from our users out there.

So if we decide to go the way to make cloudlog completely usable for different client situations, we are good positioned, if we first go to have a concept to do this before coding into the blue.

As you mentioned, this would be a lot of work but it could also be a big effort for the users.

[...]

PR #383, opened 2019-10-27
Multi client

dh9dat:
Hi Peter,
first of all, many thanks for creating, maintaining and sharing CloudLog as an OpenSource-Project. We're currently evaluating it for our local radio club (DARC O54), and during the evaluation, we've noticed that it would be nice to add multi-client functionality. This would make it possible to provide a web-based logbook for our members, providing CloudLog as a kind of SaaS (or maybe better "LaaS", Logbook as a service).
So, I tried to make the necessary changes to make CloudLog able to handle connections from different users at the same time, who are working with the same profile (think of club stations) or with different station profiles.
I've tested it a bit, and as far as I can see it is working as expected:
- I've added a column "user_station_id" to the users table to store the active station id for this user which is updated when the user changes the station profile,
- I've changed the logbook views in Logbook and QSO to show only the QSOs of the active station,
- I've added an onchange event to the station option list in the QSO view which sets the new statio 
 id as active and reloads the page to show the last QSOs etc. for the new station,
- oh, and since I'm using Eclipse with the DBeaver-Plugin, I've added their special files etc. to .gitignore since I obviously don't want to check them in

If you like, please take a look at my changes, I'm looking forward to any feedback from your side.

Have a nice hackin' and vy 73 from Harald, DH9DAT

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -41,8 +41,19 @@ class Logbook extends CI_Controller {
 		} else {
 				$data['qra'] = "none";
 		}
-
-
+		$CI =& get_instance();
+		$CI->load->model('Stations');
+		$station_id = $CI->Stations->find_active();
+        $this->db->where('station_id',$station_id);
+        $query = $this->db->get('station_profile');
+        if($query->num_rows() >= 1) {
+            foreach ($query->result() as $row)
+            {
+                $data['title']  = "Logbook of station '" . $row->station_profile_name ."'";
+            }
+        } else {
+            $data['title'] = "Logbook";
+        }
 
 		// load the view
 		$data['page_title'] = "Logbook";
--- a/application/models/User_model.php
+++ b/application/models/User_model.php
@@ -241,7 +241,8 @@ class User_Model extends CI_Model {
 			'user_eqsl_qth_nickname' => $u->row()->user_eqsl_qth_nickname,
 			'user_hash'		 => $this->_hash($u->row()->user_id."-".$u->row()->user_type),
 			'radio' => isset($_COOKIE["radio"])?$_COOKIE["radio"]:"",
-			'station_profile_id' => isset($_COOKIE["station_profile_id"])?$_COOKIE["station_profile_id"]:""
+			//'station_profile_id' => isset($_COOKIE["station_profile_id"])?$_COOKIE["station_profile_id"]:""
+		    'station_profile_id' => $u->row()->user_station_id
 		);
 
 		$this->session->set_userdata($userdata);
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@
 .idea/*
 .DS_Store
 sync.sh
+/.buildpath
+/.project
+/.dbeaver
--- a/Scripts/.gitignore
+++ b/Scripts/.gitignore
@@ -1,0 +1,1 @@
+/Script.sql
--- a/Scripts/add_column_user.sql
+++ b/Scripts/add_column_user.sql
@@ -1,0 +1,1 @@
+alter table users add column user_station_id integer not null default 0;
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -22,7 +22,24 @@ class QSO extends CI_Controller {
 		$data['notice'] = false;
 		$data['stations'] = $this->stations->all();
 		$data['radios'] = $this->cat->radios();
-		$data['query'] = $this->logbook_model->last_custom('5');
+		//if (isset($_COOKIE['station_profile_id'])) {
+		//    $station_profile_id = $_COOKIE['station_profile_id'];
+		//    $data['active_station'] = $this->stations->profile($station_profile_id);
+		//} else {
+		//    $station_profile_id = 0;
+		//}
+		//$station_profile_id = isset($_COOKIE['station_profile_id']) ? $_COOKIE['station_profile_id'] : 0;
+		$station_profile_id = $this->stations->find_active();
+		$station_changed = false;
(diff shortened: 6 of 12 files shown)


magicbug:
Hi Harald,

Thanks for the PR, indeed multi-client is something that Cloudlog does require, however, implementation is something I hadn't gotten around to, I've not had time to check your PR just yet but will do this evening.

Couple of points for multi-client would be

- Need to retain a user being able to have more than one station profile, for example, I have around 15 profiles.
- The idea with station profiles is that within "QSOs" when it returns previous QSOs it returns all logged with that station that could be from another station profile that belongs to the login.
- Logbook/Dashboard I always kept that area designed to show all QSOs from station profiles, obviously, that gets more complex if you have more users with their own logbooks and in some respects might not be right to show those without being logged in.
- The search was across all station profiles
- Analytics, Awards was for the active station profile.

There is some discussion on this inside Issue #358

I'll take a look at the PR later and provide some feedback

dh9dat:
Hi Peter,

many thanks for your quick response, really haven't expected it so fast, :-). I'm very interested in your thoughts about my changes (btw., I've just found a bug after activating it on a server with a bit more QSOs in the database than my test server holds: It seems that the page counter on the logbook doesn't care about the active station, but still counts all QSOs in the database - haven't noticed that, sorry...).
But that's of course only a minor point, I've read the points you mentioned and also took a look at issue #358 (didn't know that there was a discussion already, I just started by cloning the repo and digging into the code, being more or less an educational hobby for myself).
Concerning your points, my approach was to just add a column to the user table which stores the active station id for this user instead of setting a cookie, and then show only the QSOs for this station profile in all views.
This gives the user the ability to choose any station profile as the active one, and even switch between profiles during a QSO (for example, log an interesting DX contact for the clubstation and also for your private callsign).
The search is still across all profiles (haven't touched that code), and Analytics, Awards etc. are also unchanged since they are already station-specific.
The logbook and the dashboard are changed, since they are now only displaying the QSOs for the active station profile, no matter which user logged them.
But of course, there's at least one disadvantage with this approach: As the user can choose any profile from the database he/she can also choose a "foreign" one, may it be intentionally or by accident.
So, you're right, if we want to implement this safely we would also need to implement some kind of rights management to ensure that no user can switch to a foreign profile but only to the own ones.
And, eQSL, Clublog, QRZ, LOTW and all the other user-specific things must be considered, too (just had a request from one of our members if it would be possible to show more than 25 QSOs on a page, preferably making this configurable for each user...).

Again, many thanks for your quick response and your time to check my addons,

vy 73 from Germany,

Harald

application/controllers/Logbook.php
```diff
@@ -41,8 +41,19 @@ function index()
 		} else {
 				$data['qra'] = "none";
 		}
-
-
+		$CI =& get_instance();
+		$CI->load->model('Stations');
+		$station_id = $CI->Stations->find_active();
+        $this->db->where('station_id',$station_id);
```
Maescool: Minor: indent wrong

application/models/User_model.php
```diff
@@ -241,7 +241,8 @@ function update_session($id) {
 			'user_eqsl_qth_nickname' => $u->row()->user_eqsl_qth_nickname,
 			'user_hash'		 => $this->_hash($u->row()->user_id."-".$u->row()->user_type),
 			'radio' => isset($_COOKIE["radio"])?$_COOKIE["radio"]:"",
-			'station_profile_id' => isset($_COOKIE["station_profile_id"])?$_COOKIE["station_profile_id"]:""
+			//'station_profile_id' => isset($_COOKIE["station_profile_id"])?$_COOKIE["station_profile_id"]:""
+		    'station_profile_id' => $u->row()->user_station_id
```
Maescool: Minor: indent

dh9dat: Hello Maescool,

thanks for reviewing my code, and many thanks for your comments!
First of all, sorry for the wrong indentation - that's something I hadn't noticed because I'm developing with Eclipse, and Eclipse doesn't make a difference between spaces and tabs. And since I'm normally using spaces while Peter uses tab stops it looked okay in my editor view (and still does). I will fix this, and use tabs for indentation from now on.
Concerning the logbook code for the station profile, you're right, the model is the right place for this - this was my first approach, and since it worked I haven't spend more time to design rules, which is of course something that should be fixed.
Also, the submit event in the onchange handler was my first (working) idea, I'm sure there's a better solution for this. 
But, as said, it's only a first approach, and I thought it would be good to upload it to see what other developers who are a bit more familiar with the code think about it so they can tell me if I'm on the right way.
And, as we see, this has happened, and it seems that I'm not completely wrong, :-).
Let's see what Peter thinks about my approach, his first comments gave me a lot of stuff to think about, :-).

magicbug:
Haven't been a handle to pull and have a nose, but I did dump some ideas on multi-client in this thread https://github.com/magicbug/Cloudlog/issues/358 appreciate any feedback.

Also just because I noticed it, any database additions need to be migration files, info about these can be found in the Codeigniter documentation, better than SQL import files as its automated :)

dh9dat:
Hello Peter,
first of all, thanks for the tip with the migration files, I'll take a look into the Codeigniter documentation - I'm not so familiar with this framework (to be honest, it's my first time), so I didn't know that there's an option for this.
I've also read your PDF in the other thread (thanks for uploading it), and I will comment your thoughts there to avoid spreading things across multiple threads.

magicbug:
I've started work on this in the Cloudlog-v2 branch currently just redesigning the login/register pages so that the dashboard is only seen after login is complete, I'll add a public dashboard for users to display based on a token or similar.

PR #394, opened 2019-12-07
Improved Countries Breakdown

dg9vh:
by showing confirmed (by paper-QSL) and needed confirmed.

image

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -47,7 +47,9 @@ class Dashboard extends CI_Controller {
 		$data['total_digi'] = $this->logbook_model->total_digi();
 		
 		$data['total_countrys'] = $this->logbook_model->total_countrys();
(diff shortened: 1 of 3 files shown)


magicbug:
I'm concerned that this would cause confusion because is not obvious that the bracketed number is for paper and generally isn't an important stat well in my opinion at least.

dg9vh:
So would it be better to remove the number in brackets and rename "confirmed" to "Confirmed(Paper-QSL)"?

I think the number of confirmed countries would be valuable... we could divide it also into 3 numbers (Confirmed Paper/EQSL/LoTW)

Would this be an idea?

dg9vh:
I added mentioned thing. Now it looks like this:
image

So needed is as it was before, only the row "confirmed" has been added now...

dg9vh:
Hi Peter, how about die PR now? Or would you commit it, if we make it configurable-visible via config-file?

PR #398, opened 2019-12-22
Added ESC-Key for resetting all QSO-Fields (quick-delete at qso-entry)

dg9vh:
Enables ESC-Key in QSO-Window to reset all entered fields (including callsign) for quick delete after frequency change for example.


magicbug:
Thanks for this, I've actually held off merging while I'm trying to get V2 codebase justed for allowing more than one user on the cloudlog-v2 branch

dg9vh:
Hi and no problem :-)

PR #405, opened 2020-01-23
Allow for spaces in eQSL QTH nicknames

### LINKED ISSUE #404: eSQL import broken if QTH nickname contains a space
**Describe the bug**
At first, I thought this would be #377, because the behaviour was the same. Export to eQSL works, but pulling from eQSL only gave a blank white screen and no QSOs imported. With a bit of echo-debugging I found out, that eQSL was returning status code 400. Using the Cloudlog-generated URL in the browser gets a valid 200 response.

**To Reproduce**
1. On eQSL choose a QTH Nickname containing a space, e.g. `Home QTH`
2. Set this nickname on your station profile
3. Supply eQSL credentials on user profile
4. Trigger an eQSL import with the pull method
5. See white page

**Expected behavior**
The QSOs should get imported. Errors during the import should be reported.

**Additional context**
In a URL, the space should be represented by a `+` and not as a literal space.

abrain:
Fixes #404 

This made it work for me.

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -143,7 +143,7 @@ class eqsl extends CI_Controller {
 			$eqsl_url .= "&Password=" . $data['user_eqsl_password'];
 			
 			$eqsl_url .= "&RcvdSince=" . $eqsl_last_qsl_date;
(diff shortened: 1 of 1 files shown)


magicbug:
Thanks

PR #413, opened 2020-02-10
Fix "non zero" date error

ofadam:
I've tested these edits and it seems to fix the error one gets when running Cloudlog with MySQL 8.0.


magicbug:
@ofadam Hi Adam, thanks for this PR, I'm not having a good health day so won't get round to merging/checking till tomorrow but thanks for taking time to fix the zero date issue.

dkozinn:
I pulled the latest version, removed the config changes for mysql, and tested and it works.

PR #435, opened 2020-03-13
Markers

mbridak:
Add Mode markers.


nr0q:
I've got some more icon files I'm working on, I'll add them here as I finish them. I'd like to get more of the common ADIF modes done before we look at merging this branch back in. Maybe even get bands added in too.

magicbug:
Thanks for the PR, currently unable to merge as I get asset 404s

- Failed to load resource: the server responded with a status of 404 (Not Found) marker-shadow.png:1 
- Failed to load resource: the server responded with a status of 404 (Not Found) icon-fm.png:1 
- Failed to load resource: the server responded with a status of 404 (Not Found) icon-ssb.png:1 

Will need to check this code further.

magicbug:
> I've got some more icon files I'm working on, I'll add them here as I finish them. I'd like to get more of the common ADIF modes done before we look at merging this branch back in. Maybe even get bands added in too.

@nr0q excellent, I'd suggest perhaps just small dots with different colours as not everyone will want big markers small points would allow maximum visibility of the map

mbridak:
Yeah, I have to figure out how to expose the codeigniter base URL to javascript. I tried a javascript only approach with window.location.origin and a portion of window.location.pathname but this obviously does not work since the path will change depending on if it's installed in the servers root or a subfolder. Anyhoo I'll try and pop down the codeigniters mvc rabbit hole and find where to change/add it when i get home.

mbridak:
I worked out problem with the icons.

A quick comment about colored band indicators. I think there would be waaaay to many band color/mode combinations.

If you want band/mode, I'd simplify the icons reducing mode indicators to something like:
P - Phone
C - CW
D - Digital
S - Satellite

log

PR #461, opened 2020-04-13
Added posibility to manually set dxcc and cq zone in qso entry window.

AndreasK79:


magicbug:
I'm sadly going to reject the PR, while it's probably useful to have those fields changeable I think on the main QSO panel isn't the place I've tried to keep that as simple as possible

- DXCC Drop down could be placed somewhere else perhaps general, also on change really a check needs to be done to see if it's needed/worked and the relevant badges/colours set, if DXCC is changed but Gridsquare is populated it shouldn't change map location either.
- IOTA change to select is fine, however the population of the field from QRZ/Callbook no longer functions.
- CQ no objection to it being an exposed field but again not in the current position could be under General and might be worth exposing ITU zone as well.

AndreasK79:
I understand. I will make some changes and fixes, and make further commits to this PR.

As for ITU zone, I had that in mind, but as it was not set in dxcc_entities, I chose not to include it. I looked into the code, and the code sets it, but I could not find the ITU zone in the cty.xml... If @g7vjr can include it, I can add it.

AndreasK79:
@magicbug Can you take another look now? I think I managed to fix everything.

AndreasK79:
@magicbug Have you forgotten about this? Afaik I have fixed all the issues you pointed out.

magicbug:
I had will double check it later 😊

magicbug:
@AndreasK79 Yep that looks OK to me merge approved

PR #475, opened 2020-04-24
Added statistics for number of days with qso each year.

AndreasK79:


magicbug:
@AndreasK79 just a comment, i'm guessing it should be telling me how many days I've been active but on my live copy its only a couple of days which isn't correct.

Are you using TIME_ON as the count or some other method?

AndreasK79:
@magicbug I counted distinct col_qso_date, grouped on year. It seemed ok when I tested it, but I can always change to col_time_on if that is better.

PR #487, opened 2020-05-02
Ignored QSLs are greyed arrowed...

dg9vh:
Should look like this now:
image


myzinsky:
I strongly support this idea!

PR #490, opened 2020-05-08
Auto grid locator

stratoss:
If the QTH locator is empty, and the location field is set, trying to automatically determine the QTHLOC using OpenStreetMap's Nominatim API.

### DIFF
--- a/application/config/cloudlog.php
+++ b/application/config/cloudlog.php
@@ -139,4 +139,4 @@ $config['map_6digit_grids'] = FALSE;
 |
 */
 
-$config['qso_auto_qth'] = TRUE;
+$config['qso_auto_qth'] = FALSE;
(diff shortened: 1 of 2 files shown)


magicbug:
Thanks for submitting the PR @stratoss its a nice idea to be able to figure it out from the persons location, I'd suggest its controllable with an on / off, could be simply done via the /application/config/cloudlog.php with a true/false.

For myself, I chase squares on VHF+ and I'd rather have no square populated to reduce risk of miss logging which could affect chasing the square.

stratoss:
@magicbug thanks for the proposal. Indeed such automation needs to be controllable. I've added the qso_auth_qth setting in the file mentioned, defaults to true.

BTW: Thanks for the great application, and making it OSS.

73 de LZ3ST

magicbug:
Thanks will check it tomorrow and then should be good for merge, 

Most welcome for Cloudlog, it's getting there and being OSS means anyone can make changes they want.

PR #499, opened 2020-05-17
Configurable modes

dg9vh:
This PR makes mode-list editable and prefilled with all actually listed modes from ADIF-Spec. in Edit mode and prefilled with those as before.

Now Users can activate/deactivate modes from being shown within the Mode-selection in new-QSO-logging-windows.


magicbug:
This looks interesting, I think one thing that needs to improve is comments so that it's obvious what the controllers/models tasks are this is something that needs to be improved across the board within the Cloudlog codebase going forward.

Will give it a test and some feedback this evening!

dg9vh:
Nice to hear :-) 

Just go testing - we can do some good documentation on it... but I think, it would be usable for novices :-)

AndreasK79:
Excellent. I was thinking about something like this, since the modes were locked down with the change to a dropdown that I did. 

Several modes are hardcoded in javascript with default reports. Should this somehow be implemented with a default report, so that the javascript fetches from this?

dg9vh:
I actually had problems within the frequency-library to get the infos from the modes-model... no idea how to solve this... It could be an idea to put default-reports also into a column in the database...

magicbug:
One other thing that would be an improvement to the system is modes and sub modes as that keeps coming up

dg9vh:
If I get an idea on how to realize it, we can fill both fields in adif with correct values but only have the modes in one list to choose from...

For example: if you choose PSK31 it would resolve to mode = PSK and submode = PSK31.

On database, we can add a column parent mode containing the main-mode... if empty, we know, that mode = main-mode so no submode is to be filled... or we can have 2 columns "mode" and "submode"... but that means, rebuilding the table from scratch :-)
Would be not a big problem... would think about it... on how to do it... and on how to map it into corresponding fields.

dg9vh:
I added support for sub-modes on different levels...

dg9vh:
Some further additions of submode ... I think, I now catched all..

dg9vh:
@magicbug what's about merging this PR into the master?

dg9vh:
@magicbug Hi Peter, what's open on this pull-request that hides you from merging?

ofadam:
I tried this commit out and it seems to work well – I can not only see the submode when applicable on log screens, but can also see the submodes listed as nested items when editing a QSO. I couldn't find any issues. Nicely done!

dg9vh:
hi @magicbug what's exactly the problem with merging this feature into the main-code? We are idling around about for 3 months now without any specific problem mentioned on this PR but with no moving...

Please merge it into the code, it locks serveral other developments here on a stale state.

dg9vh:
mni tnx fer merging :-)

magicbug:
If you can just drive it a brief overview in the wiki for people that would be useful.

dg9vh:
Already started at https://github.com/magicbug/Cloudlog/wiki/QSO-Modes

PR #500, opened 2020-05-17
Update satellite_data.json

Ciemon:
Added data for all listed satellites


magicbug:
You did good, needed a decimal removed and APRS sats changed from FM to PKT but apart from that was spot on!

PR #508, opened 2020-05-21
Alphabetized satellite_data

mwimages:
I sorted this data alphabetically by satellite name for easier selection in mobile lists/pickers.


magicbug:
Thanks

PR #518, opened 2020-05-25
Added Castle on The Air (COTA) Award Field

myzinsky:
Since 2014, COTA-Team-Germany is awarding the Castle on the Air (COTA) Diploma
Programme together with the DARC-OV Stiftland U23 association. More than 3000
diplomas have been sent free of charge to over 30 countries during this period.

This commit adds COTA to cloudlog as award and also in the station profile.


myzinsky:
Okay that could be. So the question is if we need COTA_REF and MY_COTA_REF in the ADIF export. 

By checking the COTA website, we actually do not need it: 

https://www.cotagroup.org/cotagroup/service/logs-einreichen/?lang=en

The COTA Reference must be in the logfile name. I dont think it should be in the ADI file itself.

magicbug:
I don't have a problem adding it, would just need to use APP_CLOUDLOG_ COTA_REF & APP_CLOUDLOG_ MY_COTA_REF as the tag name

myzinsky:
Should I change it only in the ADIF export or should I change it everywhere consistently?

myzinsky:
Hi @magicbug, what would be your preference?

myzinsky:
I think I have to redo it because of conflicts. I also want to bring COTA on the table of the ADIF standardisation team.

magicbug:
Sorry this got caught up in a bad health period and got pushed down my list, yes lots of have changed in the codebase this this PR so merge wont be happy.

We can still add it to Cloudlog I'd suggest we make showing it user definable and till ADIF add it to the spec it will be user_defined within the ADIF spec.

myzinsky:
Discussion is ongoing here: https://groups.io/g/adifdev/topic/76658151

magicbug:
I'll let you restart the PR when ready @myzinsky but for now will close

daccle:
I possibly solved by #688 as COTA seem to follow the MY_SIG and MY_SIG_INFO pattern as, POTA and WWFF.

PR #519, opened 2020-05-25
reworked QSL-Indicators and print-functions 

dg9vh:
reworked QSL-Indicators and print-functions to not only use "R" but also
look for "Q" (Queued)... this is logically better if you initiate QSL-sending
by your own... R is for requested by other station to send...

Now we are back to ADIF-Specs-logic :-)


magicbug:
Will double check in the morning but looks fine for merge thanks do try to keep Cloudlog to the adif spec sheet where possible

Issue #524, opened 2020-05-26
Navigation bar not fully accessible on mobile device


dl6nm:
If I open Cloudlog on my mobile device, not all navigation bar entries/elements are visible and accessible. For example I'd like to logout but I cannot reach the "Logout" entry in "Logged in as DL6NM". Also the "Admin" entries aren't fully visible and also not accessible. Neither scrolling down the page nor scaling the content are a solution to work around.

**Steps to reproduce the behaviour:**
1. Take a mobile device
2. Click on the navigation bar icon in the upper right corner
3. Select 'Admin' or 'Logged in as ...'
4. Try to select the last item

**Expected behaviour**
As a user of Cloudlog I'd like to access all navigation bar items.

**Screenshots**
![IMG_0506](https://user-images.githubusercontent.com/1536542/82950173-2527b180-9fa5-11ea-8024-aaf10e63371d.jpg)

**Smartphone (please complete the following information):**
 - Device: iPhoneSE
 - OS: iOS 13.4.1
 - Browser: Safari, Chrome

magicbug:
Now resolved.

PR #525, opened 2020-05-27
Changed navbar to be sticky instead of fixed

dl6nm:
Fixes problem with non visible and accessible links in navbar #524


magicbug:
Not PR merged but change made

PR #549, opened 2020-06-26
Fixed confusing behavior...

dg9vh:
... when clicking on the "Mark imported/exported QSOs as uploaded to LoTW"... orifinally clicking on the export-label sets the marker on import, which was a bug by doublicated IDs on the HTML-entities.


AndreasK79:
@dg9vh Oops. Copy/paste is not always good. I thought I fixed all the references. Thanks for fixing them.

PR #568, opened 2020-08-06
Bugfixes

teertz:
QSO bugfixes (duplicates issue fix, timezone issue fix, do not call "Reload" popup if data is not changed)
Map coordinates update
Add Transmit Power to the API

Requires SQL update:
`ALTER TABLE cat ADD tx_power varchar(255);
 ALTER TABLE cat ADD rx_frequency varchar(255);`

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -469,7 +469,7 @@ class API extends CI_Controller {
 		$this->cat->update($obj);
 
 		/* Set transmit power */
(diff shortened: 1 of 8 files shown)


magicbug:
Hi @teertz thanks for these, I'll double check things and make sure its OK to merge, any SQL changes are handled via migrations for reference but I'll get that added before committing to master.

magicbug:
I'm sadly rejecting this PR, because theres too many changes to allow it within one commit and its hard to fully track whats being changed.

PR #579, opened 2020-08-19
Split the query for the dxcc summary in two. Some configurations of m…

AndreasK79:
…ysql didn't like this type of query.

@magicbug Can you give this a try?

### DIFF
--- a/application/controllers/Awards.php
+++ b/application/controllers/Awards.php
@@ -128,6 +128,7 @@ class Awards extends CI_Controller {
 
 		$dxcclist = $this->dxcc->fetchdxcc($postdata);
         $data['dxcc_array'] = $this->dxcc->get_dxcc_array($dxcclist, $bands, $postdata);
+        $data['dxcc_summary'] = $this->dxcc->get_dxcc_summary($bands);
 
 		// Render Page
 		$data['page_title'] = "Awards - DXCC";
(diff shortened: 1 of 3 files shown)


magicbug:
This looks better, WAS will need similar treatment shall commit this and see if it barfs on production

magicbug:
Looks like its still unhappy on production (guessing MySQL server versions differ on this machine)

A Database Error Occurred
Error Number: 1140

In aggregated query without GROUP BY, expression #1 of SELECT list contains nonaggregated column 'logbook_2m0sql_master.thcv.COL_BAND'; this is incompatible with sql_mode=only_full_group_by

SELECT thcv.col_band, count(distinct thcv.col_dxcc) as count FROM TABLE_HRD_CONTACTS_V01 thcv where station_id = 1 and thcv.col_prop_mode ='SAT'

Filename: models/Dxcc.php

Line Number: 499

magicbug:
This seems to actually be to do with this

SELECT thcv.col_band, count(distinct thcv.col_dxcc) as count FROM TABLE_HRD_CONTACTS_V01 thcv where station_id = 1 and thcv.col_prop_mode ='SAT'

thcv is of course not the name of the table name for myself

magicbug:
Scrap the above, brains not functioning :p too many pain meds...

But def not working here sadly

AndreasK79:
@magicbug It's because another column is included with the count. I removed it, but seems I need to make all modifications again on another branch and make a new PR. Pushing to the existing branch and trying to make a PR, doesn't seem to work.

magicbug:
Think it got caught when I reverted the WAS changes from Git sorry about that.

PR #580, opened 2020-08-19
Add summary table to WAS Awards area

AndreasK79:
…y in two. Some configurations of mysql didn't like this type of query.


AndreasK79:
@magicbug This PR has the necessary modifications done to the WAS summary. I also removed band from the query, as it is not needed after the rewrite.

magicbug:
@AndreasK79 DXCC merged OK without errors on production, so I think once you make the changes to this one to fix the conflicts it should be good to merge.

AndreasK79:
@magicbug Merge conflict is resolved now.

PR #594, opened 2020-09-01
Lotw sync

magicbug:
Getting ready to include Logbook of the World Sync to the master branch of Cloudlog, still has some outstanding tasks but wanted to make sure it would merge!

outstanding items

- [x] Only upload QSOS between the cert created and expired data nothing else
- [x] Add US county field to the Station Profiles area and to the TQ8 export
- [x] Remove the QSO LIMIT on the SQL call to stop lidness as it needs to be ALL for the real world.


magicbug:
Hi @AndreasK79 and @dg9vh wanted to make you both aware this is coming soon. I need to remove some development testing bits

AndreasK79:
@magicbug Great work Peter. I'm curious about how this is done, so I'll see if I can dig into the code when I have some time.

PR #595, opened 2020-09-01
Configurable updates folder

mattmelling:
I am in the process of packaging Cloudlog for [NixOS](https://nixos.org/). Nix requires that packages are installed in to a read-only store, meaning it is not possible to download updates to the `./updates` folder. This pull request introduces a `datadir` configuration option which allows us to redirect the updates folder elsewhere - in my specific case to a systemd tmpfiles location.

If `datadir` is not defined, updates will be simply be written to the normal `./updates` folder.

I've purposefully named this `datadir` as we could apply similar treatment to the `uploads` folder as well, however I've not looked at this in any detail. (If anyone has sample files for the ADIF/LOTW/etc imports I'd be happy to take a look at similar changes for those features.)

73,

Matt M7AQT

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -19,6 +19,8 @@ $config['directory'] = "logbook";
 
 $config['callbook'] = "hamqth"; // Options are hamqth or qrz
 
+$config['datadir'] = null; // default to install directory
+
 /*
 |--------------------------------------------------------------------------
 | Logbook Options
--- a/application/controllers/Update.php
+++ b/application/controllers/Update.php
@@ -18,6 +18,19 @@ class Update extends CI_Controller {
 	}
 
     /*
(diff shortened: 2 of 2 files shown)


AndreasK79:
@magicbug any reason this is not merged yet? Seems perfectly fine to me.

application/config/config.sample.php
```diff
@@ -19,6 +19,8 @@
 
 $config['callbook'] = "hamqth"; // Options are hamqth or qrz
 
+$config['datadir'] = null; // default to install directory
```
domnantas, line 22: `install/config/config.php` should be updated with this too. The `/install` route uses it as a blueprint for `application/config/config.php`.

I'm not exactly sure why these files are duplicated, I'd say either `application/config/config.sample.php` or `install/config/config.php` should be kept

PR #603, opened 2020-09-03
Fixed QSO Tab Problems in Safari

myzinsky:
In Safari on macOS the tabs are not shown corretly. This commit changes the
padding of the QSO-Tabs such that they are shown correctly in the Safari
browser.

Related discussion in the Forum
https://forum.cloudlog.co.uk/d/45-qso-panel-tabs-changing-text-to-icons/6


magicbug:
Looks fine on Chrome too so gets the nod and merged

PR #614, opened 2020-09-11
WAS javascript dialog qso

AndreasK79:


magicbug:
Thanks this is on my things to merge list today!

magicbug:
@AndreasK79 Looks great, love the popups way nicer and the way you changed the marking QSLs.

magicbug:
Approved, guess next jobs putting the new popup for edit on the logbook areas.

AndreasK79:
@magicbug Glad you liked it. Yes, it has to be implemented several places. I know DXCC And CQ awars needs to have the same rework. The logbook area is also a place. After that, I'm not sure, but when all is done, the double up of files can be removed.

PR #615, opened 2020-09-13
Measurement base moved to station_profile

AndreasK79:


magicbug:
As per the issue this is user profile choice rather than per station profile, which means then it can then just be grabbed from the session data rather than a DB call each time.

AndreasK79:
Haha. You are right. I see that now. My head was so locked on station profile, I didn't think about user profile. It's back to the drawing board then.

AndreasK79:
@magicbug Can you take another look? I hope I managed to change all the places that were necessary.

magicbug:
Thats looking better, I'm thinking its probably wise to keep the cloudlog.php value as the default so that if your not logged in that value is used could catch any errors where people haven't yet filled in the field yet.

AndreasK79:
Ok. I did this to set a default value:
`'user_measurement_base varchar(1) DEFAULT "K"',`

You mean I should change it and check for a value in config if user_measurement_base is null?

dg9vh:
He thought it would be good to have the value set in cloudlog.php as a fall-back-value if there is no value set in the database - and choosing "K" as default in the database differs from the default-value in cloudlog.php... I do not think, that this would be a good way... It would be better to read the value from cloudlog.php (because it is a CONSTANT-value, it would be available I think within the config-set) and write this into the database if a new station-profile would be created. So one can set his defaults within the cloudlog.php and tune it within the database, if necessary.

AndreasK79:
Config is now fallback. As I have added the field to the user profile as a dropdown, it will be set when creating a new profile.

AndreasK79:
When creating a new user, the select is set according to what is set in the config-file.
@magicbug I think that should cover things.

magicbug:
Yep looks fine, just fixed the issue with the migration class name but shall merge thanks

PR #617, opened 2020-09-16
Changed the dropdown selects to look the same (custom-select).

AndreasK79:


AndreasK79:
This is in user profile.

PR #620, opened 2020-09-16
Added some more date-formats and added date-format-chooser into user-…

dg9vh:
…add-dialog.

### DIFF
--- a/application/controllers/User.php
+++ b/application/controllers/User.php
@@ -62,7 +62,7 @@ class User extends CI_Controller {
 		}
 		else
 		{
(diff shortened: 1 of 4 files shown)


magicbug:
When adding fields need to make sure logic supports them on the backend, added the missing code to save date format when creating a new user.

dg9vh:
ok, mni tnx :-) I did not checked it - I thought in origin the author of the code only missed putting it into add.php :-)

dg9vh:
So I think, now it is ready to be merged...

magicbug:
Yep cleaned some other parts up in the code but ready to merge now

PR #624, opened 2020-09-18
Fixed display QSO for WAS-details

AndreasK79:
Changed so that bootstrapdialog is included on every page. Will start implementation on bootstrapdialog for other awards.


AndreasK79:
@magicbug I've implemented bootstrapdialog for all awards with this PR.

PR #625, opened 2020-09-21
Current streak is added.

AndreasK79:
Current streak has been added. It also checks if you you have a streak that will break if you don't make a QSO today.


M0TZO:
Looking forward to see this :)

PR #626, opened 2020-09-22
Changed to h2 for title, and table-sm for table to compacts tables a …

AndreasK79:
…bit.

### DIFF
--- a/application/views/awards/cq/index.php
+++ b/application/views/awards/cq/index.php
@@ -1,6 +1,6 @@
 <div class="container">
 
-  <h1><?php echo $page_title?></h1>
+  <h2><?php echo $page_title?></h2>
 
   <?php $this->load->view("awards/nav_bar")?>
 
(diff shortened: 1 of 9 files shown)


magicbug:
In terms of HTML Schema the usage of H1 is correct in that the headings dictate the purpose of the content on the page, I'd be interested to know the reasoning to want to change it to h2 when there's no other headings on the page?

magicbug:
If its to just bring it in line with the other pages that fine although really those should be changed to H1s really, being there's no further headings on most pages.

AndreasK79:
My reasoning is that I find the heading is too big, and to get more space for the table. That way you'll fit more on the page, and get a better overview on the page.

magicbug:
OK, no real objection, sizing etc can always be controlled with CSS they aren't fixed, I'll merge, we shall use H2 for page headings as the standard.

PR #627, opened 2020-09-22
Datatable added to tables in DXCC, IOTA and VUCC (in band page).

AndreasK79:
The footer had to be reworked a little bit, but better to move the included JS to the top of the file. Script tags were added so that it would have a open and close tag for script in each page. I didn't do it in a good way when bootstrapdialog was added in the awards section.


magicbug (commented):
Within the footer can datatables.min.js & datatables.min.css placed in the correct /assets folders and not via CDNs all assets should be local, this helps for those running them on internal networks.

AndreasK79:
Yeah. I can do that, but can I just download the files from the CDN and put them in the correct folders?
I used the builder here https://datatables.net/download/#bs4/jq-3.3.1/dt-1.10.22/fc-3.3.1/fh-3.1.7/r-2.2.6/sc-2.0.3
They are automatically minified and joined together, if not, I need to look more into which files are included. The downloadable zip file from that page, contained a bunch of files.

PR #628, opened 2020-09-23
New feature: theme is user selectable.

AndreasK79:
I've included the tweaked dark mode now. I fixed the theme here and there, but there might be other places that needs to be fixed as well. I think this needs to be a work in progress, and users can come with suggestions they feel are better.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 51;
+$config['migration_version'] = 52;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


magicbug:
In its current form, this can't be accepted as it causes a fundamental break in the style sheets for users who haven't already selected a style sheet, remember when things like this are added it needs to be mindful of existing users.

This will have two cases

- People who aren't logged in, this must be the default standard CSS style sheet, in till some global configuration options are added
- Have a system in place where it will automatically revert to the default stylesheet or the column in the table is prepopulated with the default css sheet.

magicbug:
Follow up for existing users, this can be using the migration

similar to 050 migration where `DEFAULT \'d/m/y\'` we can set this to the standard non dark style sheet which will fix one issue, however for non logged in users it needs to revert to standard for now.

AndreasK79:
I didn't think about non logged in users. but if you look at the commit, I have the default in the migration file.
'user_stylesheet varchar(255) default "bootstrap.min.css"',

Any suggestion on how to fix this for non logged in users?

magicbug:
Ah brain fart and missed that, just doing some code tweaks for the rest

magicbug:
There we go 👍🏻

AndreasK79:
Thanks. I didn't think about an easy if-check :)

PR #629, opened 2020-09-24
Added bootstrapdialog in dashboard, logbook and search. I think that …

AndreasK79:
…should cover it.

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -556,7 +556,7 @@ class Logbook extends CI_Controller {
 		if ($query->num_rows() > 0)
 		{
 			$data['results'] = $query;
-			$this->load->view('view_log/partial/log.php', $data);
+			$this->load->view('view_log/partial/log_ajax.php', $data);
 		} else {
 			$this->load->model('search');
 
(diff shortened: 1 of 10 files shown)


magicbug:
Looks good, only thing missing is in the QSO area where when you type callsign and it shows previous QSOs the callsigns are clickable, shows Fancybox still now minus the map

AndreasK79:
That's one that I missed. I will fix that one.

AndreasK79:
@magicbug Hopefully that was the last one.

magicbug:
Found one last item /index.php/search/filter the advanced search popups!

AndreasK79:
I see. That one is filled in via javascript. That's the reason I missed it. Will see if I can fix that one.

AndreasK79:
@magicbug and done... ?

AndreasK79:
@magicbug I just thought of one thing I did, that perhaps should be removed. In footer.php, in qso_save, I run a location.reload(). The reason I did this, was so that changes to a QSO would be reflected in the award table when QSO was saved, and dialog was closed. 

This line of code is in the codebase, but since this PR is about implementing Bootstrapdialog to other parts of Cloudlog, I think a commit to this branch is perhaps the best.

What are your thoughts on this Peter?

magicbug:
Location reload seems a good thing to have to see the changes

PR #649, opened 2020-10-09
Added button to activate/deactivate in mode list.

AndreasK79:
This is done dynamically with javascript.


magicbug:
This is much nicer with it not requiring page reloading :)

If your ever bored, be good to swap the "Create new mode" to a popup box and button at the top.

Merging this as is however

AndreasK79:
I had something like that in mind, but I also wanted to add Datatable to the mode-table. As Datatable is now included everywhere, I can add Datatable other places.

PR #666, opened 2020-10-18
Added QRZ.com Icon after Callsign in Log

myzinsky:
This is not really a PR, its more a feature request with some code example. But I think its a useful feature. I'm open also to bring it to some other positions. Let me know how you want to have it ;-)


magicbug:
While I don't personally need this feature, I'm wondering about its placement as an option to show it, think it would need to have its own column and option for either QRZ or HamQTH.

Think this would also need to be user-definable to whether its shown or not

myzinsky:
Yes I totally agree with you. I will work on that. Do you have any preferences for icons? :-)

PR #668, opened 2020-10-19
Datatables csv export

AndreasK79:
Also a little cleanup in awards-files at the same time as I was modifying files anyway.


magicbug:
@AndreasK79 Looks great, not sure how much effort it would be at a later point but with CSV say VUCC if its unconfirmed would be ace if the worked callsign could be added for chasing cards 👍🏻 

Merging as is though

AndreasK79:
@magicbug It's difficult to add this to the export itself, as it just takes what is presented in the table and exports. I guess with some modifications, this can be added to the VUCC table. I'll make a note of it.

magicbug:
That would work :)

PR #670, opened 2020-10-21
Normalise styles of admin pages

### LINKED ISSUE #652: Suggestion - Normalise heading styles
**Is your feature request related to a problem? Please describe.**
There are three or four styles used on the admin pages when it comes to headings and sub-text. These are the main ones:

1. One card with small header text and then small description text. E.g Radio Interfaces page
2. As above but the first line of the description text is a large font. E.g. API page
3. Multiple cards, each with bold header text and small description text. A big page title above all cards. E.g ADIF page

**Describe the solution you'd like**
I'd like to clean these up a bit but wanted to check the preferred style before starting any work. We could pick one of the three main styles already in use or define something new. 

I think I like a variation of the third style, maybe with a slightly smaller page title font and non-bold headers for each card. The page title could be set from the $page_title property, just like the browser tab title, and any card titles can be set by hand (if there are several) or can use the same $page_title content again.

poll-busily:
Resolves #652 , at least for pages under the Admin menu drop-down. This PR uses the LotW page style (H2 headings, title taken from $data['page_title']) with non-bold card headings and plain card description text. I tweaked the card descriptions for certain pages. For the Updates page I added a card to wrap the content as there wasn't one already.

### DIFF
--- a/application/controllers/Adif.php
+++ b/application/controllers/Adif.php
@@ -139,7 +139,7 @@ class adif extends CI_Controller {
 	public function index() {
 		$this->load->model('stations');
 
-		$data['page_title'] = "ADIF Import";
+		$data['page_title'] = "ADIF Import / Export";
 		$data['max_upload'] = ini_get('upload_max_filesize');
 
 		$data['station_profile'] = $this->stations->all();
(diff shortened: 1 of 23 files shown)


magicbug:
@poll-busily Paul huge thanks for taking the time to go through this and normalise it

PR #671, opened 2020-10-21
Put LOTW uploads in uploads/lotw

jancona:
I ran into an issue where LOTW uploads didn't work because Apache didn't have write permission to the top-level `Cloudlog` directory. While giving it write permission fixed my issue, it seemed better to write the temporary upload file in the `uploads/` directory, hence this PR.


M0LTE:
I've cherry-picked and applied this and it works great

PR #676, opened 2020-10-27
Normalise buttons

poll-busily:
Well, yep, this is a PR that is all about buttons! It does two or three buttony things.

### Sets button colours
- Blue Solid = Add something, activate something, or perform any other primary action.
- Blue Outline = Edit, change or update something.
- Red = Delete something or make an irreversable change.
- Grey = Deactivate something.
- Yellow = Reset something or clear a field (no permanent delete).
- Green = Search.

### Adds/modifies button images
- For example, adding plus symbols to buttons that add new items
- Added a couple of upload/download cloud icons on buttons relating to cloud sync
- Fixed a couple of cases where button icons were no longer available

### Fixes button layout and alignment
- E.g. Centre-align buttons that are in table cells so that buttons with different text lengths don't stand out so much

### DIFF
--- a/application/views/accumulate/index.php
+++ b/application/views/accumulate/index.php
@@ -83,7 +83,7 @@
                 <!-- Button (Double) -->
                 <div class="form-group row">
                     <div class="col-md-10">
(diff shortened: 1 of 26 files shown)


magicbug:
Thanks, Paul for taking the time to standardize it all, you clearly have some patience :) 

I've added guidelines to https://github.com/magicbug/Cloudlog/wiki/Development-Notes feel free to update this should make things more straightforward going forward.

PR #677, opened 2020-10-28
New feature: QSL card image upload

AndreasK79:


magicbug (changes requested):
PRs for migration and QSO popup window will need resolved

Image storage for eQSL I store images in /images/eqsl_card_images not sure if we need to make some changes to make sure everything goes into assets rather than in different areas

AndreasK79:
@magicbug Should be good to go now.

As for the directory, it's up to you. I just went by what you mentioned :)

magicbug:
Yeah tricky isn't it! i think moving to assets makes more sense not sure why i didnt brain fart i think! will have a play with this

magicbug:
Looks good, I'd suggest being able to upload the front and back at the same time.

Not sure if its possible on upload to return me back to the QSO popup feels a bit clunky

magicbug (approved):

AndreasK79:
As I went for the carousel, I ripped out the code for uploading 2 images at once. I originally had a type in qsl_images to indicate front or back, but that didn't make sense any more. 

I did try to get it to work with ajax and js, but I couldn't get it to work. I went for the regular route for now, but I agree that it would be better to just return back to the QSO popup.

magicbug:
Carousel is OK, but having to dig back to the same QSO to then upload the other part of the card is too clunky for my liking so need to figure out a solution to that.

magicbug:
Not sure if we can use https://github.com/CreativeDream/jquery.filer however that includes multi upload and even cropping so perhaps worth a look

magicbug:
Sorry linked to the out of date library meant this one https://innostudio.de/fileuploader/ same as they list it looks quite decent actually.

AndreasK79:
I agree about uploading 2 images at once. I had it working, but I ripped it out. It shouldn't be too hard to put back in.

I'll take a look at the library, but for now, I'll just add back 2 image upload.

AndreasK79:
@magicbug Give it a go now. 2 image upload re-added.

AndreasK79:
The fileuploader library isn't free as far as I can tell. Only free for trying, and the files are encrypted according to the website. Is this really something we should use?

AndreasK79:
DropzoneJS and Cropper.js are both MIT. Perhaps this might be something?
https://itnext.io/integrating-dropzone-with-javascript-image-cropper-optimise-image-upload-e22b12ac0d8a

magicbug:
I totally missed it wasn't free! yes the two you suggest look nice

AndreasK79:
I made some progress with JS uploading. The files are uploaded, so I need to do some updating in the dialog, or perhaps I should just load the dialog again instead?

This is not with DropzoneJS and Cropper.js. I will take a look at that later.

AndreasK79:
@magicbug I think everything should be working now with javascript. Could you give it a go and test it out?

magicbug:
Thats better 👍🏻 shall merge it in!

Issue #686, opened 2020-11-07
Please don't load pooper.min.js from Cloudflare.com


schnitzelbert:
I just found out that Cloudlog is loading its popper.min.js from Cloudflare.

IMHO this should be locally hosted to avoid privacy issues.

Thank you!

magicbug:
PRs accepted, however as you would have seen most things are there's just a few bits lurking that need changed locally.

It's on the to do.

AndreasK79:
PR #689 added.

PR #689, opened 2020-11-08
Javascript: changed to local copies of files.

AndreasK79:


daccle:
Looks good to me. BTW. Is there a build tool to keep track of those external libs and to keep them updated?

AndreasK79:
I'm not aware of any build tools for Cloudlog.

daccle:
This should fix #686 

If the there is no tooling, how does the project keep track of newer versions of this JS stuff?

AndreasK79:
That is a good question. I think it is done manually, but @magicbug might have a system himself?

magicbug:
Manually by me.

daccle:
BTW: you were missing the items in application/views/qso/edit_ajax.php:15

AndreasK79:
@daccle You are so right. Thanks for the heads up. Will look into it when I find some time.

daccle:
I can take care of that, if you don't mind.

AndreasK79:
Be my guest. The files should be there, just a matter of changing to local path.
It's the same thing in edit.php. I think that file is no longer in use, but can't say for sure. I think I changed over to edit_ajax everywhere.

Issue #687, opened 2020-11-07
ADIF Export misses Comments/TX-Power 


schnitzelbert:
Having a look into the exported ADIF files shows, that the comment field is missing and TX-Power is missing.

It would be great to have a full featured ADIF export with all existing datas.

daccle:
Can you give an example for given entities?

PR #691, opened 2020-11-09
Adds TX_PWR and COMMENT to adif export

daccle:
This adds TX_PWR and COMMENT to the adif export as requested in  #687.


daccle:
The next time that I will touch  `application/views/adif/data/exportall.php` I will write a helper function like adif_field_helper, like this

PSEUDOCODE
```
function adif_field_helper(column, adif_field_name):
    if column not empty:
        field_len = len(column)
        print <uppercase(adif_field_name):field_len:str(column)
```

And then I will generate the ADIF export iterating over the array of column/field_names.

PR #694, opened 2020-11-11
Added support for lookups in services for more detailed information when logged in

jsb2092:
Created function to do lookups in the model to avoid repeated code.  Names are now only shown when logged in as well.


magicbug:
While I understand the full name is available, the reason I never included it was due to the simple fact is that someone's surname isn't relevant for a QSO and its worth remembering although this is ham radio in Europe your still storing personal identifiable information.

For that fact, unless someone can give me a valid reason for it, full names won't be included.

Adding Grid and stuff from QRZ on imports a good idea, although need to make sure if QRZ doesn't respond that it cleanly errors out and QSOs still get imported.

jsb2092:
I'll change the full name to be an option in config.php.  Off by default. I like have the full name in my log book because having a lot of John's in my logbook isn't very helpful to me.  

I understand gdpr would prevent me from storing it if the data were stored in Europe.  Though my understanding is it doesn't apply to person use.  For example if I have desktop logging software and it looks up in qrz the name (HRD and MacLggerDx do this), I believe it's fine.  This being said I think an option is the right choice.  

I will check about the durability of updating on import and get back to you shortly.

jsb2092:
I added a flag to check for using full_name in the config which should satisfy the requirement here.  I also added exception handling to the import.

jsb2092:
The qrz/hamqth lookup now only happens on import if name or gridsquare are not set.  If they are, there is no reason to look them up.  This also means they can be overridden to not do the lookup and set whatever the user doing the import wants. 

All lookups to QRZ now also have the config setting to use_fullname, which is set to false by default, and must be explicitly turned on in the config.php file.  This should prevent people from getting this is a hosted environment (such as yours), but those who would like full names can turn it on.  

I hope this addresses your concerns.

73,
w2jsb

magicbug:
Looking great, sorry for not interacting over the weekend busy, chaos etc.. I'll check it over soon but config option looks ok 👍🏻

jsb2092:
No worries - I found an fixed a couple other things related to that (some of which I may have caused), so the delay wasn't the worst things that's happened.

magicbug:
Just a random thought while having a cup of tea, always wondered whether supporting QRZ and HamQTH at the same time might be an idea, for example if QRZ doesn't find a match Cloudlog would check HamQTH

jsb2092:
That is certainly something I could add (probably without too much difficultly).  I also want to clean up some of it, because the import logic is separate from the manual enter piece, and it'd be nice if was all the same workflow, which would make the failover much easier.

I like the idea though, so I'll come up with a way to allow the user to configure fail over and priority in the UI too.

73,
w2jsb

magicbug:
I have only been considering as I've noticed when doing satellite ops, some people I work in Europe arent on QRZ but have full details on HamQTH

jsb2092:
makes sense.  I've also noticed that there are a good number of people (especially in England) who aren't on QRZ but are on HamQTH.  It is funny (in a good way) that the more I use CloudLog, the more I like it, but certainly appreciate you being open to making it better!

magicbug:
OK so as this changes a view parts can you list which areas of Cloudlog has been changed and have either been rested or need testing, this will allow us to merge it.

jsb2092:
As of now, the only view parts that have changed are that names don't show up at all when you aren't logged in, and full names should show up when adding a QSO or in the logbook when the option is set in the config.php file (there is a sample in the install directory).

I would like to get this merged before doing the failover option, however if you'd like me to do that first, I am certainly willing.

AndreasK79:
@magicbug could you revisit this and see what's missing to get it merged? Is it just the merge-conflict?

install/config/config.php
```diff
@@ -40,10 +40,21 @@
...
+$config['use_fullname'] = false;
+
+
+/*
+|--------------------------------------------------------------------------
+| Are we running this on a personal server?  If we are, GDPR laws
+| prevent us from storing personal information.
+|
+*/
+$config['personal'] = false;
```
AndreasK79, line 57: @jsb2092 is this something that never got implemented? I can't se it is used anywhere.

AndreasK79:
@magicbug I've tested this out, and it works as it should. It's just an expansion of what is already there, so I don't see any problems in merging this. Agree?

install/config/config.php
jsb2092: I use it on my hosted server... It's never been merged.

magicbug:
I'm guessing that without $config['personal'] = false; in config file its not going to cause any issues?

jsb2092:
AFAIK it shouldn't cause any issues.  Stored in the DB would be good.  I don't recall that existing when I wrote it, but I could be wrong.

magicbug:
We should probably add something to the docs to say this functionality is there.

PR #696, opened 2020-11-13
Lengthen COL_CONTEST_ID for long contest names

ofadam:


magicbug:
Which ones aren't compatible for adif spec contest_id I believe fields long enough I'll double check this all when at a pc httpss://adif.org/311/ADIF_311.htm#Contest_ID

ofadam:
It turns out that Log4OM has much longer IDs listed by default and it somehow was written into the ADIF for me – this seems to be just an individual instance, so I'll close this ticket. The max on the ADIF spec appears to be 22 characters, within the limit specified in Cloudlog setup.

PR #699, opened 2020-11-14
[Adif export] Refactor the view and added a library to make the adif-…

AndreasK79:
…line. Also a bug fixed in MY_STATE. It was exporting COL_STATE, not COL_MY_STATE


daccle:
Hi Andreas! 
Great that you implemented my ideas so quickly. I am not 100% sure about the abstraction here, since the adifhelper is only specific to exportall.php and there are other types of adif export (like for clublog). As my personal preference I would avoid specific functionality to a a general helper function.
But we could improve that and for the time being your solution is much cleaner than the code before. So I would support to merge your pull request. And maybe we can work on a more general solution.
Best, Daniel

AndreasK79:
Hi David.

When you mentioned about a rewrite, my mind started working. It was fun to implement. 
I'm pretty sure this can be used other places as well, but perhaps need a little tweaking.
I know I wrote a function in logbook_model to create an ADIF-line for the upload for the QRZ logbook.

For now, it makes editing the ADIF-export a much more pleasant experience.

daccle:
Hi @AndreasK79!

Don't get me wrong! I like your work. I've just worked with it, when adding SIG and SIG_INFO and it is better than before. I just want to avoid to pile up tech debt, so for the moment the solution is a big improvement and IMHO should be merged. I don't have a good idea on how to improve this in a more "general" way right now.

Best, 
Daniel

PR #701, opened 2020-11-16
Change to show submode if available - otherwise mode

ofadam:
This change matches the initial dashboard view.


dg9vh (approved):
Looks good so far.

Issue #695, opened 2020-11-13
Edit QSO: Station "Transmit Power (W)" is not viewable or editable


TrystanDavies:
Description:
Once a QSO is entered, the Station "Transmit Power (W)" is not viewable or editable if the Edit QSO option is selected. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to gear icon next to a QSO
2. Select Edit QSO from the dropdown
3. Look for Transmit Power (W)

**Expected behaviour**
It should be possible to edit a QSO and enter the transmit power. 

**Desktop**
 - OS: Fedora 33
 - Browser Chrome
 - Version 85.0.4183.102 (Official Build) (64-bit)

Best regards,
Trystan G0KAY

daccle:
I've opened a pull request to fix this.

TrystanDavies:
Thanks Daccle

daccle:
You're welcome.
73, DC2ZP

PR #703, opened 2020-11-16
adds TX_PWR to qso edit dialog

daccle:
This adds the tx power to the edit dialog. This should fix: #695


magicbug (changes requested):
While it adds the input box, it forgets to add the business logic to save any changes to the field, this will need to be addressed before it can be merged.

daccle:
Will do the needed edits.

daccle:
I hope that I have found the right place.

magicbug:
Yep that looks good to me

Issue #704, opened 2020-11-16
Add POTA (Parks on the Air) to the awards section


lasantiagoa:
Incidentally, the SIG_INFO and MY_SIG_INFO fields would need to be used when exporting those log entries.
SIG_INFO (the park where the other station was operating)
MY_SIG_INFO (the park reference from where I am operating)

That way the log can be exported and sent to POTA for processing.

daccle:
Don't forget about MY_SIG and SIG. MY_SIG and MY_SIG_INFO fields are available since #688 is merged. See my comment in #700 to enable Park2Park.

daccle:
I would suggest to close this issue and start a pull request to add SIG and SIG_INFO.

daccle:
This still doesn't offer POTA in the AWARDS section, but since there are a couple of funny things based on SIG/SIG_INFO we should aim for a more general solution, since adding POTA, COTA and like WWFF on their own would crowd IMHO the AWARDS section

magicbug:
Biggest issue is theres so many awards now that it could end up with cloudlog having 20 fields just for different award types which gets a bit much, issue is I don't really know the best way to handle it.

daccle:
I am looking into the awards which are based on SIG/SIG_INFO. I have no solution so far, but I don't want to clutter CloudLog. So either there is a solution based on a config setting or we parse the content of SIG to generate a list of possible awards.

For the moment I would to see this issue here to be closed. I will come up with a proposal for a more general solution.

dg9vh:
How about having just a few fields (I think 3 should be ok) where we have a listbox for label (to choose the award programme) and a textfield to enter the value for the number? Or to make it flexible: have an award table where we can enter n awards for each qso and display them into a table in the QSO-edit-dialog? so we can have 1 to n relations for qso to awards and vice versa... by this way it would also be very easy to have all QSOs for a specific award...

dg9vh:
I my eyes after thinking about it, the second way (with having an extra table building up the n:m-relation for qso:awards) would be the best I think.

daccle:
For moment I know of five awards, which do use SIG/SIG_INFO:
* COTA
* POTA
* TOTA (Toilets on the Air) (which is a bit specific...)
* WWFF
* GMA (German Mountain Award)

[...]

Issue #700, opened 2020-11-16
Feature request: support for Parks on the Air


ryandzink:
**Describe the solution you'd like**
Would it be possible to add Parks on the Air/POTA to as an awards category/field, similar to SOTA references? I think something like "POTA Number" as an input field on the QSO entry field similar to the SOTA reference would work for this.

**Describe alternatives you've considered**
Currently I am just adding POTA details to the notes in the QSO but it would be nice to have some similar Awards features like SOTA/IOTA etc.

ryandzink:
I'm going to close this as it seems to be a duplicate of or very similar to #688 . POTA uses the same MY_SIG_INFO field, with the addition of a SIG_INFO field for Park-to-Park QSOs.

daccle:
Well, only parts of the issue are solved in #688 The enable Park2Park contacts, we would to have complementary to MY_SIG and MY_SIG_INFO, fields for SIG and SIG_INFO as well. They would need to added to the QSO dialog as well. I am willing to add this, but before I fiddle with the QSO dialog I would to have the consent of @magicbug himself.

PR #705, opened 2020-11-16
Add the possibilty to log SIG and SIG_INFO

daccle:
This would solve the issues for Park2Park and Castle2Castle and WWFF2WWFF contacts. Like described in #700 and #704 

This still work in progress


daccle:
I am pretty sure that I got all possible spots to add SIG and SIG_INFO. There is no report in the AWARD section though, but one would have to figure out a clever solution to get all the possible stuff based on SIG/SIG_INFO. I don't want to add something for WWFF, POTA, COTA and so on. There must be a more general solution.

daccle:
I have rebased on top of the current master and added the now missing fields to AdifHelper

magicbug:
Looks great, we might need to add some info to https://github.com/magicbug/Cloudlog/wiki/Logging about the SIG fields and what they are used for.

daccle:
Okay, I write some documentation in the wiki. Since I am non-native writer of English, some proof-reading is highly apreciated.

daccle:
@magicbug I've added a very brief information about the general tab in the wiki. I guess we should invest some time and love on the wiki.

magicbug:
Shall have a read this evening, but yes wiki always needs some love seen your Issue and will get involved and get some new images up 👍🏻

PR #706, opened 2020-11-17
Display submode if available

ofadam:
This looks for submode first on the "MODE:" display fields – matching how the logbook itself functions.


magicbug:
Reverted merge after noticing that lots of the changes in logbook.php completely broke the system due to incorrectly closed tags.

PR #707, opened 2020-11-17
Display submode if available (more)

ofadam:
This looks for submode first on the "MODE:" display fields – matching how the logbook itself functions.


ofadam:
Accidental duplicate! My apologies.

PR #710, opened 2020-11-17
Use local copies of JS and CSS assets

daccle:
Based on the work of @AndreasK79 I replaced the links to external sources for js and css in the qso edit views


magicbug:
thanks for spotting these as I've said probably dotted about a few places!

PR #731, opened 2020-12-06
[KML Export] Bugfix, as it was not working. 

AndreasK79:
Also made a GUI with options for what you want to export.

### DIFF
--- a/application/controllers/Kml.php
+++ b/application/controllers/Kml.php
@@ -9,7 +9,28 @@
 
 class Kml extends CI_Controller {
 
-	public function index()
+    public function index()
+    {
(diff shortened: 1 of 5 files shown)


magicbug:
Thanks for going back over this @AndreasK79 I think the code actually goes back to when I visit started the whole project

AndreasK79:
@magicbug happy to help. I tested the export by calling it directly, and it failed. I didn't change the way it fetches QSOs from the database. It fetches only QSOs with a grid, but since we have a DXCC list with a longitude and a latitude, we could include other QSOs as well. I'll see if I revisit this further down the road.

PR #734, opened 2020-12-06
Change theme loading + new themes

poll-busily:
## Overview
Change the way themes are bundled and loaded so it's more maintainable. Add a few nice themes from [Bootswatch](https://bootswatch.com), where the original themes came from IIRC.

### Theme Loading
Today the name of the theme CSS file is stored in the users table. This means that all themes have a common directory and can only use a single file for customisation. I've changed this so that the database stores the name of the theme (also its directory name under assets/css). We standardise the file naming so that the primary CSS file is "bootstrap.min.css" with an "overrides.css" file for any cloudlog-specific tweaks. These both work with the assets/css/general.css file that was already there.

### New Themes
I love dark mode so I pulled in some funky new themes from Bootswatch. Two dark and one light:  Cosmo, Darkly and Superhero. The existing "Dark" theme was, I think "Cyborg" from Bootswatch. Since there are now multiple dark themes that will now show up as "Cyborg (Dark)".

### RFC
Please check my database migration file as I haven't written one of those before. We need to set the default theme again for all users, using the new naming convention.


AndreasK79:
Excellent work Paul. I never thought of putting them in separate folders when I added them. You are correct, I got the themes from Bootswatch, as that needed minimal tweaking to get them to work. I don't remember the name of the blue one, but it must be Superhero. 

You need to add override for Darkly and Superhero. Check awards and dxcc. The table header is not readable.

AndreasK79:
I just noticed that you also need to change the version number in config/migration.php.

poll-busily:
Ah yes, that makes sense! Thanks for reviewing - there was a high chance I'd get something wrong with my first migration.

AndreasK79:
Happy to help. I think I found all the modifications done to the css file.

They are as follows:
`@media (min-width: 992px) {
  .modal-lg,
  .modal-xl {
    max-width: 900px; -- this was increased from 800 to 900 to increase width of bootstrapdialog.
  }
}

.table thead td {
  background-color: #2B3E50;
}

form .btn-outline-success {
  color: #fff;
  border-color: #77b300;
}

.text-muted {
  color: #E1EBEB !important;
}

td .alert-danger{
  color: #fff;
}`

magicbug:
Looks good, PHP wise etc I'm happy with it, I'm guessing with the Migration this will reset everyone back to default till they change it again, but worth it for greater good long term. I've made a few changes when saving changes to make it less clunky so I'll merge and any further theme changes can be added.

PR #739, opened 2020-12-09
[ADIF Parser] Modifications done to the ADIF parser.

AndreasK79:
This speeds up ADIF files containing several thousand QSOs.

I've tested this on several variants I have of ADIF files, including the type coming from LoTW.

### DIFF
--- a/application/libraries/Adif_parser.php
+++ b/application/libraries/Adif_parser.php
@@ -99,7 +99,6 @@ class ADIF_Parser
 				}
 			}
 
-            $this->datasplit = preg_split("/<eor>/i", substr($this->data, $this->i));
 			$this->i++;
 			
 		};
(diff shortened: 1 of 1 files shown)


magicbug:
Need to improve testing so need to confirm the following works

- [ ] General ADIFs import via the Admin -> Import
- [ ] eQSL functions all work
- [ ] LoTW functions all work

AndreasK79:
I agree, sorry about missing that one. I wasn't aware that the LoTW import used it, and that they don't follow the standard, as no EOF-tag exists. They even say that here: https://lotw.arrl.org/lotw-help/developer-query-dxcc-credits/?lang=en

I have tested several ADIF-files, from different software, through the import with no trouble. LoTW was the only one giving an error. That was fixed by checking that a call-tag exists.

I don't have eQSL, so I can't test how that works.

PR #749, opened 2020-12-11
[ADIF import] Skip QRZ Logbook upload when importing ADIF from import and LoTW

AndreasK79:


AndreasK79:
I don't see the point in uploading to QRZ Logbook when importing from ADIF. That will take extra time, and we have functions to upload this after the log is imported.

application/models/Logbook_model.php
```diff
@@ -1454,7 +1454,14 @@ function eqsl_not_yet_sent() {
       return $this->db->get();
     }
 
-    function import($record, $station_id = "0", $skipDuplicate, $markLotw, $dxccAdif, $markQrz) {
+    /*
+     * $skipDuplicate - used in ADIF import to skip duplicate checking when importing QSOs
+     * $markLoTW - used in ADIF import to mark QSOs as exported to LoTW when importing QSOs
+     * $dxccAdif - used in ADIF import to determine if DXCC From ADIF is used, or if Cloudlog should try to guess
+     * $markLoTW - used in ADIF import to mark QSOs as exported to QRZ Logbook when importing QSOs
```
k7tab, line 1461: I think you meant $markQrz instead of $markLoTW here.

AndreasK79: Ah yes. Thanks for noticing. Copy/paste and not changing what was copied...

PR #753, opened 2020-12-15
Timeplotter satellite mode correction

ryandzink:
The timeplotter currently does not work for satellite contacts when that is selected as the mode, but does when satellite contacts are bundled in when choosing Band=All.


AndreasK79:
@ryandzink nice catch. I missed this when I added this feature. Thanks for fixing.

PR #764, opened 2020-12-22
Tweaks to text (mostly capitalisation)

poll-busily:
Touches page titles, some table headers, menu items. There's some noise in the diffs because my editor was trimming whitespace - sorry about that.

### DIFF
--- a/application/views/contesting/index.php
+++ b/application/views/contesting/index.php
@@ -4,13 +4,13 @@
 
         <div class="col-sm-12 col-md-12">
             <div class="card">
-                <div class="card-header"><h5 class="card-title">Logging form</h5></div>
+                <div class="card-header"><h5 class="card-title">Logging Form</h5></div>
                 <div class="card-body">
                     <form id="qso_input" name="qsos">
 
                             <div class="form-group row">
 
-                                    <label class="col-auto control-label" for="radio">Exchange type</label>
+                                    <label class="col-auto control-label" for="radio">Exchange Type</label>
                                     <div class="col-auto">
                                         <div class="form-check form-check-inline">
                                             <input class="form-check-input" type="radio" name="exchangeradio" id="serial" value="serial" checked>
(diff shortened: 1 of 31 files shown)


magicbug:
Accepting this PR might take a while due to the sheer number of changes made see https://github.com/magicbug/Cloudlog/wiki/Development-Notes

Due to previous commits like this causing breaking changes each controller changed will need checked

poll-busily:
No worries. I'll tidy it up a bit and push another revision.

poll-busily:
Ahh, that's better. Apologies for the formatting changes before - SNR should be much better now. There are still controller changes but you'll see they're minor.

application/views/contesting/index.php
```diff
@@ -379,7 +379,7 @@
                                         </div>
                                 </div>
                             <button type="reset" class="btn btn-sm btn-warning" onclick="reset_log_fields()">Reset</button>
-                            <button type="button" class="btn btn-sm btn-primary" onclick="logQso();"><i class="fas fa-save"></i> Log QSO</button>
```
poll-busily, line 382: "Save QSO" was used elsewhere so I'm aiming to standardise on that.

application/views/notes/edit.php
```diff
@@ -47,7 +47,7 @@
 	</div>
 
 	<input type="hidden" name="id" value="<?php echo $id; ?>" />
-	<button type="submit" value="Submit" class="btn btn-primary">Save</button>
```
poll-busily, line 50: Many save buttons are in the form "Save X", where X is a QSO or whatever. Using the same format for notes.

application/views/contesting/index.php
dg9vh: I don't think that saving a QSO is the same as logging a QSO... here we are in Contest-Context creating new log-entries, with "save" you are in edtiting context...

poll-busily: Good point. Let me go back and review the current usage again.

AndreasK79: I feel that "Log QSO" is more correct, but you are also saving the QSO in the database. I named it "Log QSO" in the Contest part.

poll-busily: @magicbug, would you like:

- "Save QSO" all around. I'll keep this commit as-is.
- "Log QSO" all around. I'll change the buttons in the Live QSO and Post QSO pages to match the Contest Logging page and I'll undo this part of my commit.

PR #782, opened 2020-12-28
[Contesting] Restoring QSO table.

AndreasK79:
@magicbug can you test this out and see if I missed anything?

### DIFF
--- a/application/controllers/Contesting.php
+++ b/application/controllers/Contesting.php
@@ -51,4 +51,16 @@ class Contesting extends CI_Controller {
             }
 
     }
+
+    public function getSessionQsos() {
+        //load model
+        $this->load->model('Contesting_model');
+
+        $qso = $this->input->post('qso');
+
+        // get QSOs to fill the table
+        $data = $this->Contesting_model->getSessionQsos($qso);
+
+        return json_encode($data);
+    }
 }
--- a/application/models/Contesting_model.php
+++ b/application/models/Contesting_model.php
@@ -6,4 +6,30 @@ class Contesting_model extends CI_Model {
         parent::__construct();
 
     }
+
+    /*
+     * This function gets the QSOs to fill the "Contest Logbook" under the contesting form.
+     */
+    function getSessionQsos($qso) {
+        $CI =& get_instance();
+        $CI->load->model('Stations');
+        $station_id = $CI->Stations->find_active();
+
+        $qsoarray = explode(',', $qso);
+
+        $contestid = $qsoarray[2];
+        $date = DateTime::createFromFormat('d-m-Y H:i:s', $qsoarray[0]);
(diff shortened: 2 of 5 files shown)


magicbug:
The Contest info restores OK, but the QSO table isn't restoring

magicbug:
OK that was strange, QSO table didnt restore after just one QSO but when 002 went in it restored from 002 onwards OK so guessing there's a bug somwhere

magicbug:
While we're working on this the reset buttons need some work

- Reset contest session - needs to reset the exchange fields, remove the contest id etc
- Reset QSO - really should just clear out the callsign, rst R, exchange, name and comment

AndreasK79:
> OK that was strange, QSO table didnt restore after just one QSO but when 002 went in it restored from 002 onwards OK so guessing there's a bug somewhere
> 
> When I say restored 002 onwards I mean it didn't ever show 001 again

If I log a QSO, then press F5, the table is restored. The first QSO is logged with a contestid. If you change contestid after that, that QSO will never be restored. I'm not sure how one should go about this. We need some identifier to fetch the QSOs from the DB. The query is using the contestid, but if you have any other ideas, pitch in.

AndreasK79:
> While we're working on this the reset buttons need some work

> * Reset QSO - really should just clear out the callsign, rst R/S, exchange, name and comment

I changed the Reset QSO button. It doesn't reset the whole form anymore. I think that should be better now.

AndreasK79:
> * Reset contest session - needs to reset the exchange fields, remove the contest id etc

It's resetting all the localStorage now, but do you mean that it also should perform the same as Reset QSO + set Contest Name to Other in form?

magicbug:
Yes so Reset QSO should only be for resetting QSO fields when your logging the contest QSOs.. 

Resetting the contest session should reset the exchange serial to 001, clear all the qso fields, set Contest Name to Other, exchange type to serial. aka a full clean reset of contest environment.. When I press the button at the moment nothing happens

magicbug:
As for my first comment however of the first QSO saved disappearing from the QSO table, that still happens

If I log QSO 001 then QSO 002 and reload the Contesting Page, 002 is populated but 001 is missing

magicbug:
> As for my first comment however of the first QSO saved disappearing from the QSO table, that still happens
> 
> If I log QSO 001 then QSO 002 and reload the Contesting Page, 002 is QSO is populated but 001 is missing in the Past QSOs table

This seems to be a not every time event, think its something I'm doing rather than a bug :p

magicbug:
OK think I have the buttons basically working how I'd expect them

AndreasK79:
@magicbug great. Been outside in the snow here. Seems like you have taken care of refining it :)

PR #798, opened 2021-01-07
[WIP] Dockerize Cloudlog

dgdavid:
Hi there!

First things first: thank you a lot guys for your work here as well as for making it open source :heart:

Now, the context of this (and two more comings) PR:

After a little research, @leroydiazg and I concluded that Cloudlog could be a perfect fit for our father (known in the amateur radio world as [EA8CPU](https://www.qrz.com/db/EA8CPU)) track his _QSOs_. But, before setting an instance for him, we'd have to translate it to Spanish. However, we soon realized that almost all _application strings_ were_ hardcoded.

That's why we met past [Jan 3rd](https://github.com/dgdavid/Cloudlog/commits/add-spanish-translations) to

* Dockerize the application, to speed up the development process.

  Probably a whim from my side, but, apart from all advantages of containerization, it would be great to just type `docker-compose up` and have the application running without caring about the requirements. Wouldn't it?

  This is what this PR intends for. Let's talk about it later on.
	
* Add support for internationalization.

   As said, we found a lot of hardcoded text in the application, although this is [partially not longer true after Jan 5th](https://github.com/magicbug/Cloudlog/commit/5623b004c1722509363721d94da3a1ece347b7c3). So we started going through the application and extracting them into language files.

  See #799.
	 
* Add the Spanish translations.
  
  See #800.

So, this is the PR intended for _dockerize_ Cloudlog. 

As my Docker knowledge is very basic and my first contact with CodeIgniter X was three weeks ago or so (when we found Cloudlog),  I started playing with the [detailed example](https://avenir.ro/codeigniter-4-using-docker-apache-mysql/) made by @avenirer.

But the idea would be to simplify it a bit, use a MariaDB image instead of the MySQL one, and try to _automate_, if possible, the Cloudlog installation. What do you think, @magicbug?

PR #799, opened 2021-01-07
[WIP] Add support for translations

dgdavid:
Hi there!

As said in #798, a few weeks ago @leroydiazg and I found Cloudlog and started thinking to use it after contributing to the project by adding the Spanish translations.

However, we realized that actually the application _didn't_ have support for internationalization... and we got on adding it last [Sunday, Jan 3rd](https://github.com/dgdavid/Cloudlog/commits/add-spanish-translations). ~~Unfortunately, we didn't saw that @magicbug had an open [project](https://github.com/magicbug/Cloudlog/projects/7) to address it and we couldn't join forces~~.

Nevertheless, I'm opening the PR with the changes done and notes we took, just in case it could be useful.

* System/Framework translations

  CodeIgniter documentation [suggest using _approved translations_](https://codeigniter.com/userguide3/libraries/language.html) that can be found in [CodeIgniter 3 Translations repositories](https://github.com/bcit-ci/codeigniter3-translations)

* Auto-loading language files

  Maybe there is a remarkable disadvantage of auto-loading all needed language files, but we were not able to find any mention of it in the documentation. So, we added all lang files to be [auto-loaded during system initialization](https://codeigniter.com/userguide3/libraries/language.html?highlight=autoload#auto-loading-languages). It looks as _better_ (?) than loading them per controller.
  
  For example, think in a translation located in a, let's say, _station_lang_ file that it is also needed in a QSO view. Why loading both lang files in the controller method instead of having them already available?
  
  However, it still not being clear to me the _right way_ to do it in case of having the language switcher feature.

* `lang` helper

  The [Language Class](https://codeigniter.com/userguide3/libraries/language.html) documentation has a reference to the [Language Helper](https://codeigniter.com/userguide3/helpers/language_helper.html) which has been auto-loaded and used, since
  
  * It's shorter to write `lang('whatever')` than `$this->lang->line('whatever')`
  * It allows to produce the form labels too (which is weird, BTW) using the second param as `for`: `lang('whatever', 'input-id')`
  

* Avoid crowding the `general_words` lang file

  Although we didn't know all the concepts you manage in the amateur radio world, we tried to use a lang file per _concept_ (QSO, QSL, Notes, Station).
  
* Missing translations

  As far as I could see, a not found translation will be an empty string. Probably because there is not a _fallback_ language since we are setting the language of our choice as a [default language](https://github.com/magicbug/Cloudlog/blob/e85c9a6ec0a2421ed54d9b34f76be93df86bcc6e/application/config/config.sample.php#L158). But not completely sure about it.
 
* A _user language_ feature would be nice.

  Maybe this [great article](https://www.sitepoint.com/multi-language-support-in-codeigniter/) could help (found in [an answer in StackOverflow](https://stackoverflow.com/questions/25343608/how-can-i-switch-languages-in-codeigniter#44917410))

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -110,6 +110,7 @@ $autoload['language'] = array(
     'pagination',
     'upload',
     'qso',
+    'notes',
     'station'
     );
 
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -109,6 +109,8 @@ $autoload['language'] = array(
     'number',
     'pagination',
     'upload',
+    'qso',
+    'station'
     );
 
 
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -64,7 +64,7 @@ $autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib');
 |	$autoload['helper'] = array('url', 'file');
 */
 
-$autoload['helper'] = array('url', 'security');
+$autoload['helper'] = array('url', 'security', 'language');
 
 
 /*
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -96,7 +96,20 @@ $autoload['config'] = array('cloudlog', 'bands', 'lotw');
 |
 */
 
-$autoload['language'] = array('general_words');
+$autoload['language'] = array(
+    'general_words',
+    'calendar',
+    'date',
+    'db',
+    'email',
+    'form_validation',
+    'ftp',
+    'imglib',
+    'migration',
+    'number',
+    'pagination',
+    'upload',
+    );
 
 
 /*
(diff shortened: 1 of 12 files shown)


dgdavid:
> Unfortunately, we didn't saw that @magicbug had an open [project](https://github.com/magicbug/Cloudlog/projects/7) to address it and we couldn't join forces.

I just notice that the project was created four days ago. That's the reason we didn't find any reference to a WIP translation.

dgdavid:
>     * It allows to produce the form labels too (which is weird, BTW) using the second param as `for`: `lang('whatever', 'input-id')`

Just to clarify it, I prefer explicit HTML. So, I'd use the helper just for _plain_ translations.

dgdavid:
Due to the fact that this PR is in conflict with changes already done, probably is better to close it. Hope @leroydiazg don't mind.

PR #800, opened 2021-01-07
[WIP] Add Spanish translation

dgdavid:
Based on changes done on https://github.com/magicbug/Cloudlog/pull/799.

For system translations, it follows the CodeIgniter documentation [suggestion and uses approved translations](https://codeigniter.com/userguide3/libraries/language.html) that can be found in [CodeIgniter 3 Translations repositories](https://github.com/bcit-ci/codeigniter3-translations).

See #798 too for more context.

### DIFF
--- a/application/language/spanish/calendar_lang.php
+++ b/application/language/spanish/calendar_lang.php
@@ -1,0 +1,57 @@
+<?php
+/**
+ * System messages translation for CodeIgniter(tm)
+ *
+ * @author	CodeIgniter community
+ * @author	Iban Eguia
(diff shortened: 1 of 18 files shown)


dgdavid:
@leroydiazg Probably would be better to wait @magicbug [ends adding the support for translations](https://github.com/magicbug/Cloudlog/projects/7). Or maybe just add Spanish translations iteratively like @ClaudeMa is doing with [French translations](https://github.com/magicbug/Cloudlog/pull/796).

Whatever approach you decide, let's discuss it offline during the week. 

Thanks!

PR #863, opened 2021-01-31
[QSO Dialog] County has been added.

AndreasK79:
You can now specify USA County in the QSO entry form.
USA County will be shown when viewing a QSO.
Changing County is also possible when editing a QSO.
@magicbug Please test. I think I covered everything.


magicbug:
Looks great and functions as I'd expect it to

Issue #864, opened 2021-02-02
Problem adding DOK into DOK-Field


dg9vh:
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the forum see readme for details**

**Describe the bug**
If trying to add Sonder-DOK 65G13 (valid from 03.09.2019 to 02.09.2020) into the DOK-Field, the entry is deleted, so I can't add this DOK. It is not listed in the "check-list" that pops up.

**To Reproduce**
Steps to reproduce the behaviour:
1. Edit QSO
2. Click on Awards
3. Enter DOK

**Expected behaviour**
DOK should be added.

magicbug:
This is an interesting problem to have @phl0 do you grab deleted? 

Guess other options just allow deleted to be typed in

magicbug:
I'll tag in @AndreasK79 too

dg9vh:
Thanks, I think "overloading" the list with manual entries would be a solution... actually we have a nice "autocorrection"-function that automatically upcases all entries, if they are in the list... so you can type q03 which will go to Q03...

This should be persistent beneath all solutions :-)

phl0:
We did only include DOKs by now. If special DOKs aka. SDOKs should be included as well we only need a slight change in code. I have a (separate) list of those. Will come up with a patch.

phl0:
General question is should we "allow" (S)DOKs that are not on the official list of DARC? Don't see any reason for that other than if Cloudlogs DOK database is out of date though ...

AndreasK79:
Thanks for testing @dg9vh. This never crossed my mind, and somehow I thought that you could enter whatever you wanted. I see I was wrong, but selectize has a create function. I tried it out, and with that enabled, it is possible to use values not in the dropdown:
![image](https://user-images.githubusercontent.com/6977712/106718206-03d19780-6601-11eb-9853-a2918f9c5252.png)

I'll add this in later.
If you want to test, edit footer.php and add
`create: function(input) {
            return {
                value: input,
                name: input
            }
        },`

to this function
`$('#darc_dok').selectize({`

phl0:
That would be one option. But still the question: Does it make sense to add "arbitrary" (S)DOKs or should we rely solely on the list by DARC (+ plus maybe "NM" for "not member" sometimes needed during contests)?

AndreasK79:
@phl0 I didn't answer, because I don't know. Are they official, and valid for any award?

phl0:
[...]

PR #879, opened 2021-02-08
Fix gitignore

phl0:
Removes (outdated) dok.txt and adds it to .gitignore


AndreasK79:
@phl0 removing dok.txt will break the autocomplete feature. Then people need to run the cron job before using this feature.

dg9vh:
But people also need to do country list updates before using cloudlog...

AndreasK79:
And some people aren't doing that. In my opinion, it would be better to supply a correct file, that can later be updated via cron.

dg9vh:
Yes but consequently on all files like that. Providing a system that initially runs anyway... Also with country files etc.

dg9vh:
Run the same way on every stage...

AndreasK79:
Then we agree @dg9vh. Everything needed, should be supplied at start, with the possibility to update.

phl0:
That was my basic initial question. But does it help to provide a file that outdated by default? 
This already was the issue for @dg9vh in https://github.com/magicbug/Cloudlog/issues/864 ....

AndreasK79:
@phl0 no, but you can correct that with and updated file now. The reason it was outdated, was that SDOKS weren't present, and NM was thought to be a good idea to add as well.

phl0:
@AndreasK79 True. But this will be outdated next month as soon as the DOK info by DARC is updated. That is that I meant. :-)

AndreasK79:
@phl0 I see. Well then, add an updated file now, then add information in the documentation that a cron job needs to be executed for this to be updated. That's all we can do.

phl0:
That would be an option. 
Only issue then is: It is part of the github repo and cannot be added to .gitignore. Meaning it will display as changed as soon as someone updates the files locally. Or maybe even add custom strings. 
Well, I will stand back from adding that file to .gitignore and just update the dok.txt in the repo.
Agreed @magicbug @dg9vh @AndreasK79 ? :)

magicbug:
If I followed correctly its agreed we keep the dok.txt in the file system so this can be closed for now

PR #881, opened 2021-02-09
Updates "distances worked" satellite dropdown to sort by bird name

ryandzink:
Without this sort, it looks like the satellites in the combo box are sorted by date of first QSO on each satellite. Sorting by satellite name seemed easier to use in my head at least.


AndreasK79:
@ryandzink thanks. It slipped my mind when adding this feature.

PR #889, opened 2021-02-17
Replaced signal reports on initial dashboard with country

ofadam:
Per this discussion: https://github.com/magicbug/Cloudlog/discussions/855


ofadam:
Forgot to also change table header – this should now work as expected.

magicbug:
While I know a few want this I'd much prefer it was selectable as to be honest I'd prefer it stayed the way it is, but maybe as my internal dxcc lookup is good.

This would be much better as a global option setting

magicbug (changes requested):
**Changes needed for merge**

For this to be merged changing signal report to dashboard will need to be general by an option

dgdavid:
@ofadam maybe the @AndreasK79's PR https://github.com/magicbug/Cloudlog/pull/897 could be source of inspiration / guide to achieve the requirement mentioned above. 

Or, who knows, perhaps @AndreasK79 could help here if you raise your hand. Just a suggestion :)

AndreasK79:
Since this seems stale, I'll take a stab at it.

AndreasK79:
#1019 has been added. This can now be closed.

AndreasK79:
@magicbug close this will you?

PR #897, opened 2021-02-26
[QSO Dialog] Auto-fill location and locator when SOTA is chosen.

AndreasK79:
The information is fetched from the API. This can be turned on/off in the user settings.

Did a refactor at the same time and moved a lot of javascript from footer.php to qso.js.


dgdavid:
Hi @AndreasK79!

I think it would be nice that, now that translations exists, to stop adding hardcoded texts in new PR and use or add their translatable counterpart instead.

What do you think? Make sense for you?

AndreasK79:
Hi @dgdavid!

You are right of course. As these parts weren't translated before, I didn't think too much about translation, as it didn't touch the view for the QSO entry area, just the user account. I would have waited with this feature if I had touched the QSO view, as I know there are 2 PR's waiting for these (hopefully they'll be merged soon).

application/views/user/add.php
```diff
@@ -120,6 +120,15 @@
...
 
+		<div class="form-group">
+			<label for="sotalookup">SOTA auto lookup gridsquare and name for summit.</label>
+			<select class="custom-select" id="sotalookup" name="user_sota_lookup">
+				<option value="0">No</option>
+				<option value="1">Yes</option>
+			</select>
+			<div class="small form-text text-muted">If this is set, name and gridsquare is fetched from the API and filled in location and locator.</div></td>
+		</div>
+
```
dgdavid, line 131: > I would have waited with this feature if I had touched the QSO view, as I know there are 2 PR's waiting for these (hopefully they'll be merged soon).

I see. But I actually meant to start adding translations with new features. I.e., to do something like 

```suggestion
		<div class="form-group">
			<label for="sotalookup"><?php echo $this->lang->line('sota_lookup_label'); ?></label>
			<select class="custom-select" id="sotalookup" name="user_sota_lookup">
				<option value="0"><?php echo $this->lang->line('general_word_no'); ?></option>
				<option value="1"><?php echo $this->lang->line('general_word_yes'); ?></option>
			</select>
			<div class="small form-text text-muted"><?php echo $this->lang->line('sota_lookup_helptext'); ?></div></td>
		</div>

```

and move those new _text lines_ to the appropriate lang file.

That way, there is not needed to _work twice_ to achieve the translations goal.

Just my two cents :)

AndreasK79: Thanks for enlightening me. I have not studied the translations too much. I did not think about general words and the general_words_lang file. I added translation for the dropdown now.

Perhaps a feature freeze until all translations are done would be the best. Not that I'm the one to decide this.

magicbug:
This looks good but as it uses a third party API endpoint and its in the QSO entry realm it needs to have the ability to turn off as not everyone will want it and also I'm worried about the impacts of the other server adding delays to cloudlog

magicbug (changes requested):
This looks good but as it uses a third party API endpoint and its in the QSO entry realm it needs to have the ability to turn off as not everyone will want it and also I'm worried about the impacts of the other server adding delays to cloudlog

Would make sense this is perhaps a user level option rather than a global one, for me personally, I'd keep it off, as I like keeping things snappy as possible

AndreasK79:
If you read what I wrote, it says that it can be turned on/off. It's in the user settings.

magicbug:
> If you read what I wrote, it says that it can be turned on/off. It's in the user settings.

Somehow missed or rather it's not showing in the interface for me, I'll double check it later.

AndreasK79:
@magicbug strange that it does not show up. I totally agree with keeping it snappy. That's why an option was provided.

ryandzink:
@AndreasK79 I just updated to the latest SHA from master and this looks great! Auto population only happened after I turned on the setting in my user options. Thanks!

AndreasK79:
@ryandzink happy to help!

PR #913, opened 2021-03-01
Create number_lang.php

oh1jw:


magicbug:
Rejected this seems to adjust a main system lang file rather than a country one

PR #919, opened 2021-03-01
Update log_ajax.php

oh1jw:
The RST fields at table header came from wrong lang line. It caused the illogicality of the translation, at least into Finnish. The page now looks consistent with the "Previous contacts" section in the QSO view.


oh1jw:
Any second opinions about this? Anyone..?

AndreasK79:
@oh1jw Yep, just checked. You are right. @magicbug is the one who must merge, as I have no write rights here.

PR #878, opened 2021-02-08
Add support for QSO Detail translations

leroydiazg:
To continue helping with translations, I decided to start by myself adding support for translations in the 'QSO Details' lightbox (as this is one of my father's most used view from Cloudlog).

Notes:
* I tried using already existing keywords, but also created the ones missing / incomplete.
* The "Station profile" is still hardcoded, becuase I've seen there are many parts of the project where these keywords are used (i.e. administration-> add station, or main view) and maybe it would be better to have all related keywords in one single lang file - perhapts station_lang.php? -

TO-DO:

* Station profile
* Gridsquare and Gridsquare (multi)
* USA-related fields
* File size warning

PR #925, opened 2021-03-07
[QSO Edit] Notes would not work if you edit a qso while being in the …

AndreasK79:
…qso entry area.

Thanks @leroydiazg for pointing this out.


leroydiazg:
Thanks @AndreasK79 for noticing the bug I posted in #878 

As I explained there, this probably was my last contribution as I don't feel this project being quite _contributing friendly_ . I'll keep an eye out on this project but I will focus now in other things.

AndreasK79:
@leroydiazg sorry to hear that. Thanks for the contributions to this project from you and @dgdavid. Btw there is a missing ) in frequency.php on line 38 in your 11m branch,

dgdavid:
> @leroydiazg sorry to hear that. Thanks for the contributions to this project from you and @dgdavid.

Thanks a lot for your always kind words @AndreasK79 :) I agree with Leroy's feelings and I join him in both, leaving and wishing the project the best.

> Btw there is a missing ) in frequency.php on line 38 in your 11m branch,

Yep, you're right. Thanks! I have to fix it properly (it was a quick and unfortunately dirty hotfix to avoid a crash when missing a default frequency after adding 11m)

See you around!

PR #937, opened 2021-03-14
[Quick lookup] New feature to quickly check worked/confirmed on band/…

AndreasK79:
…mode. Invoked by ALT-L


magicbug:
@AndreasK79 Looks really good, seems fairly fast for my setup too with 30,000+ QSOs in the database

AndreasK79:
@magicbug once I realized I didn't need some of the looping, that helped a lot. That meant I could fetch all info for the table in just 2 queries instead of having 2 queries for each mode. Adding more to search for is also pretty straightforward now.

AndreasK79:
Guess you can merge it so that others can get to test it as well.

magicbug:
Agreed no news is good news in my opinion

PR #956, opened 2021-03-20
Award sig

AndreasK79:


magicbug:
This is great @AndreasK79 and I think what @daccle was after, its getting merged right away.

In fact this is also a better way to handle WAB and WACRAL so I'll suggest we remove there separate items and from now on merge into SIG and cleans up code somewhat

PR #1012, opened 2021-05-01
[WAS MAP] Minor tweaks. Always show button. Fetches band directly fro…

AndreasK79:
…m the dropdown instead. Removed unneeded call to summary.


magicbug:
Thats much better thanks @AndreasK79

PR #1015, opened 2021-05-02
[WAS Map] Added JS from this fork https://github.com/kcaran/us-map to…

AndreasK79:
… enable labels on all states. Made a minor tweak to show popup with QSOs from clicked state. Tooltip is available, but disabled right now.


magicbug:
@AndreasK79 Thanks for all the extra work on the map! looks fantastic merging it :)

PR #1026, opened 2021-05-08
Updated Finnish translation 

oh1jw:
- New Finnish language file (account_lang.php)
- New Finnish lines to general_words_lang.php


magicbug:
Fab thanks @oh1jw

PR #1028, opened 2021-05-10
Update Qrz.php

whyIsMyNameUsed:
Not all character works for the qrz-password without urlencode.


magicbug:
Well spotted and thanks

PR #1029, opened 2021-05-13
Update satellite_data.json to add JO-97 linear transponder

ve1cwj:
Added JO-97 U/V linear transponder


magicbug:
Thanks john, for some reason I thought this was already added!

PR #1032, opened 2021-05-17
Spanish fields translations

leroydiazg:


magicbug:
Super thanks very much

PR #1082, opened 2021-07-05
Add timestamp to 'cat' update, to enforce timestamp update in mysql when other values have not changed.

jtmnt:
In mysql, timestamp (with ON UPDATE CURRENT_TIMESTAMP) is not automatically updated when field values are same as in database. 

When posting radio info via /api/radio with no changes to frequency or mode, it does not change the timestamp value and after some time there will be a warning on QSO page: "Radio connection timed-out" while in fact it is updated frequently, there is just no change in frequency or mode.

My fix is to set timestamp to the value that was posted via api.  Maybe there is better solution?

### DIFF
--- a/application/models/Cat.php
+++ b/application/models/Cat.php
@@ -41,6 +41,7 @@
 						$data = array(
 						'frequency' => $result['frequency'],
 						'mode' => $result['mode'],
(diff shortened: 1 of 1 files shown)


lu1dmf:
.

PR #1111, opened 2021-07-27
Add the gridsquare from LotW QSL to the logbook

T012m3n7oR:
The gridsquare from the QSL will now update the log if the gridsquare
was not previously populated, or if the LotW gridsquare is more
specific than what previously existed.  Previously the gridsquare was
not being used from the LotW QSL record.

This will enable the map above the logbook to more accurately display
where the QSL was with instead of default values.

 - Add 'qsl_gridquare' to the logbook_model.php:lotw_update
   implementation.
 - Refactor code in lotw_update to remove duplicate code.
 - Add gridsquare into the LotW output table on what is imported.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1401,28 +1401,45 @@ class Logbook_model extends CI_Model {
     }
   }
 
-  function lotw_update($datetime, $callsign, $band, $qsl_date, $qsl_status, $state) {
+  function lotw_update($datetime, $callsign, $band, $qsl_date, $qsl_status, $state, $qsl_gridsquare) {
 
-    if($state != "") {
-      $data = array(
-           'COL_LOTW_QSLRDATE' => $qsl_date,
-           'COL_LOTW_QSL_RCVD' => $qsl_status,
-           'COL_LOTW_QSL_SENT' => 'Y',
-           'COL_STATE' => $state
-      );
-    } else {
-      $data = array(
-           'COL_LOTW_QSLRDATE' => $qsl_date,
-           'COL_LOTW_QSL_RCVD' => $qsl_status,
-           'COL_LOTW_QSL_SENT' => 'Y'
-      );
-    }
+	$data = array(
+      'COL_LOTW_QSLRDATE' => $qsl_date,
+      'COL_LOTW_QSL_RCVD' => $qsl_status,
+      'COL_LOTW_QSL_SENT' => 'Y'
+    );
+	if($state != "") {
+      $data['COL_STATE'] = $state;
+	}
 
     $this->db->where('date_format(COL_TIME_ON, \'%Y-%m-%d %H:%i\') = "'.$datetime.'"');
     $this->db->where('COL_CALL', $callsign);
     $this->db->where('COL_BAND', $band);
 
     $this->db->update($this->config->item('table_name'), $data);
+	unset($data);
+
+	if($qsl_gridsquare != "") {
+      $data = array(
+        'COL_GRIDSQUARE' => $qsl_gridsquare
+      );
+      $this->db->where('date_format(COL_TIME_ON, \'%Y-%m-%d %H:%i\') = "'.$datetime.'"');
+      $this->db->where('COL_CALL', $callsign);
+      $this->db->where('COL_BAND', $band);
+      if(strlen($qsl_gridsquare) > 4) {
+        $this->db->group_start();
+        $this->db->where('COL_GRIDSQUARE', "");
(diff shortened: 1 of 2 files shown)


T012m3n7oR:
@magicbug Is there something that you would like me to do for this before you merge it/review it?

magicbug:
Sorry delays purely as I've been working on adding "Station Logbooks" to Cloudlog so you can group Station Locations so haven't been keeping up!

magicbug (changes requested):
I cannot see anything in the code for it to handle VUCC gridsquares,
obviously when importing from LoTW if VUCC is already set you don't
want to import something into the usual gridsquare field.

imlonghao:
FYI, if someone sets their Gridsquare to 8 chars length, the map on the index page and logbook page will crash. (no point on map)

T012m3n7oR:
@imlonghao I have an 8character grid and mine isn't crashing.  It definitely isn't handling it where I would expect though, that that isn't from this update.  this is only going to add this.

@magicbug Do you have any examples where I can look a VUCC lines in cloudlog? I don't work VHF/UHF frequently, and don't have any of my own data to review for this.  I will look for fields after work though.  Thank you for the feedback.

AndreasK79:
> FYI, if someone sets their Gridsquare to 8 chars length, the map on the index page and logbook page will crash. (no point on map)

This is obviously a bug in the handling of the gridsquares. I had no trouble reproducing it.

AndreasK79:
> @magicbug Do you have any examples where I can look a VUCC lines in cloudlog? I don't work VHF/UHF frequently, and don't have any of my own data to review for this. I will look for fields after work though. Thank you for the feedback.

Not sure what kind of examples you need, but it is stored in COL_VUCC_GRIDS, and it is separated by a comma.
It can be 2 or 4 grids, depending on if you are on the border of 2, or in the corner of 4 grids.

I've handled this in the vucc model, and it is handled in the gridsquare map.

Let me know if you need some more info.

application/models/Logbook_model.php
```diff
@@ -1411,28 +1411,49 @@ function import_check($datetime, $callsign, $band) {
...
+
+	if($qsl_gridsquare != "") {
+      $data = array(
+        'COL_GRIDSQUARE' => $qsl_gridsquare
+      );
+      $this->db->where('date_format(COL_TIME_ON, \'%Y-%m-%d %H:%i\') = "'.$datetime.'"');
+      $this->db->where('COL_CALL', $callsign);
+      $this->db->where('COL_BAND', $band);
+	  $this->db->group_start();
+	  $this->db->where('COL_VUCC_GRIDS');
```
AndreasK79, line 1440: @T012m3n7oR should this be $this->db->where('COL_VUCC_GRIDS', NULL)? COL_VUCC_GRIDS is nullable, so it makes sense to check if null or blank.

T012m3n7oR: They are one in the same.  I can change it for readability if you prefer, but codeigniter will translate that line into where `COL_VUCC_GRIDS` IS NULL

AndreasK79: My bad. Then I agree. I tried to check the documentation, but I didn't manage to find that it interprets it this way.

And yes, I did see the following line, that's why you use group_start and end.

No need to change it.

T012m3n7oR: I used the getCompiledUpdate function to verify it.  
To verify it you can: 
set(); the parameters, 
echo the getCompiledUpdate(<table>); back, 
and comment out the actual execute.

Because I do not have a separate test environment getCompiledUpdate lets me test without risking my data.

AndreasK79: @T012m3n7oR thanks. I'm still learning Codeigniter stuff. I usually find it easier to write out the sql (at least more complicated quiers) as a string and test it out in DBeaver (which is connected to my MariaDB).

AndreasK79:
I've looked through this, and it seems ok. The only thing that comes to mind, is if LoTW handles VUCC_GRIDS. If LoTW sends VUCC_GRIDS in the ADIF-file, that should be handled as well. My file didn't contain that, but that is because I haven't worked anyone on a border or in a corner. Do you know anything about this @magicbug ?

T012m3n7oR:
@AndreasK79 According to lotw developer querying - https://lotw.arrl.org/lotw-help/developer-query-qsos-qsls/?lang=en

The vucc_grids is something that could be returned, would have to validate if it existed on the record.  I can add that in a separate PR, unless we really think it should be included in this one.

AndreasK79:
@T012m3n7oR I'll merge this into station_logbooks. That's the dev branch we're using now. Please add vucc_grids to another PR.

PR #1116, opened 2021-07-30
[QSO Entry] Maritime Mobile identification fix

### LINKED ISSUE #1090: Maritime mobile QSOs
QSOs with maritime mobile stations are not handled properly by Cloudlog. I recommend adding a "NONE" option for DXCC (DXCC ID of 0) and perhaps automatically selecting the NONE DXCC if the suffix entered after the call is "/MM"

AndreasK79:
This should fix #1090 by identifying /MM QSOs.

I am not sure if this fixes all use cases regarding /MM. I've fixed the two functions identifying a DXCC.
The REGEX is based on a DXCC Lookup DJ1YFK made: https://git.fkurz.net/dj1yfk/dxcc. I've converted this to PHP, but I'm not sure if we need this complexity in Cloudlog. I ripped out the part for identifying /MM QSOs instead.

Since we have no CQ Zone, I return this as 0. By doing this, I needed to add 0 to dropdown in the QSO Entry.
I also needed to fix the summary in the CQ Zone award to look for values > 0.

Long and lat is returned as 0 values. This would give an error QSO Data popup if these values were not set.

@magicbug please give this a test. I am unsure if every part using dxcc_lookup in loogbook_model has everything it needs to work correctly.

### DIFF
--- a/application/models/Cq.php
+++ b/application/models/Cq.php
@@ -239,7 +239,7 @@ class CQ extends CI_Model{
     function getSummaryByBand($band, $station_id) {
         $sql = "SELECT count(distinct thcv.col_cqz) as count FROM " . $this->config->item('table_name') . " thcv";
 
-        $sql .= " where station_id = " . $station_id;
+        $sql .= " where station_id = " . $station_id . ' and col_cqz > 0';
 
         if ($band == 'SAT') {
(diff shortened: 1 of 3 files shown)


magicbug:
Seems to work OK, given it a test on some random callsigns! shall merge can always tweak.

PR #1143, opened 2021-08-29
Input shortcuts for date/time in post QSO section

phl0:
Would like to copy a smart idea from CQRlog: This allows you to input date and time in the post QSO section without delimiters and automatically gets reformatted. So you can just write "20210829" and it would be re-formatted to "2021-08-29". Same goes for time: "1723" -> "17:23" and even allows for omitting the leading zero: "723" -> "07:23".

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -459,7 +459,7 @@ if ($('#frequency').val() == "")
 	});
 }
 
-/* format time input shortcut */
+/* time input shortcut */
 $('#start_time').change(function() {
 	var raw_time = $(this).val();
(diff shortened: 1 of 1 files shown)


dg9vh:
I think this is no problem to merge, because it is not an invention of CQRlog... software like SAP also uses this "tricks".

AndreasK79:
Nice idea. Reminds me a bit of Taclog. You had free format input (random order) when entering data in one input field.

phl0:
Tnx merge. Wiki updated with https://github.com/magicbug/Cloudlog/wiki/Logging/7b2845eb98887f9a5a6b34b20ae689c648801ec6

PR #1150, opened 2021-09-05
Dxatlas grid export

AndreasK79:
This exports all the gridsquares worked and confirmed from the log into textfiles that DXAtlas can read

### DIFF
--- a/application/controllers/Dxatlas.php
+++ b/application/controllers/Dxatlas.php
@@ -30,7 +30,7 @@ class Dxatlas extends CI_Controller {
 		$this->load->helper('file');
 
 		// Load Database connections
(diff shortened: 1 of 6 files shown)


magicbug:
@AndreasK79 so that I dont have to worry, can we reroute this to merge with station logbooks branch which will ultimately be version two release, just kept it vague :) do love this however

Do we need to handle this to export per Logbook or per Profile? interested in thoughts as ever

AndreasK79:
@magicbug yep, changed now. Seems no complaint from the system when changing base branch.

I built it for a per logbook basis, but I guess users would want the possibility for both. I say merge this now, and then revisit this after all the other things are changed.

PR #1152, opened 2021-09-08
[CQ Award] Updated code to support station logbooks

AndreasK79:


magicbug:
See told you it wasn't complex thanks for giving it a go!

PR #1155, opened 2021-09-09
Added SOTA (SOTA_REF) to possible columns

Werzi2001:
As SOTA activations and chasing is my main activity it would be very helpful to have the SOTA (Reference) as a possible column in the logbook. Therefore I added it to the possible values. I hope that I found all positions to adapt. If my missed something or did something wrong please just let me know.


AndreasK79:
From what I can see, that should cover it.

PR #1167, opened 2021-09-10
Added support for dark mode to further charts

Werzi2001:
I added support for dark mode to more charts (I hope that I got all that were missing) as discussed in #1163. For that I created a helper function `isDarkModeTheme()` that can easily be adapted in the future if a theme should "know" if it is dark or not. Based on that method colors in the charts are changed.

I had quite some "trouble" in the `views/statistics/index.php` file because it contained a mix of indention via tab and spaces (2 oder 4 spaces). I adapted that to be consistent. Therefore the diff looks much bigger than it should. Basically I just adapted all the chart options.

One thing that does not work:
The google charts (Analytics > Statistics) show a message "No data" if there is - well - no data (e.g. on the Satellites tab). I did not find how to change the color that. Therefore it is now black on dark grey which is not really readable. But I guess that's not that bad as it is just an edge case and probably not that important.

Edit:
The background color of the charts is not just set to black but it uses the original body background color. Therefore it also works if the background is dark blue or something other.


AndreasK79:
@Werzi2001 great work. I just tried it out. It looks nice.

@magicbug would be great if you could merge this now.

PR #1171, opened 2021-09-11
Login form bootstrap

Werzi2001:
Changed the login form to use bootstrap to create a better looking form.


AndreasK79:
Looks a lot nicer this way!

PR #1180, opened 2021-09-17
[Gridsquare Map] Bugfix. The confirmed fields would not be correct. 

AndreasK79:
@magicbug Or_where seems to work only when where are two conditions in the where statement.

I couldn't figure out why the gridsquare maps showed a few fields confirmed that I know aren't confirmed.
In master, this was correct. I traced it down to or_where. I changed back the confirmed functions.

Take a look, and if you agree, merge it in.


magicbug:
Well spotted I agree 👍🏻

PR #1188, opened 2021-09-20
[Hardware] Added userid to everything that has to do with hardware.

AndreasK79:
@magicbug have a quick look at this one too.

I also see one thing.

The menu says "Hardware Interfaces". The page says "Radio Interface". Rename one of them?


magicbug:
I'd go with "Hardware Interfaces"

PR #1189, opened 2021-09-20
[API] Added userid to the API key.

AndreasK79:
@magicbug I didn't add user_id to the access and authorize functions in api_model. I think that is correct. I'm guessing the key wouldn't work if it was checked against a user there.


magicbug:
Correct, its more to filter it in the UI within Cloudlog!

PR #1190, opened 2021-09-20
Custom maps station logbooks

AndreasK79:
@magicbug take a look. Think I've fixed everything that needs to be fixed regarding custom maps and station logbooks.


magicbug:
Not forgotten just not been at home! I'll check it first thing tomorrow morning against my log :)

PR #1194, opened 2021-09-22
Correct user date formating on QSO pages (Live/Post)

T012m3n7oR:
The date field will now be populated from the user's preferences, if
this is not set it will fallback to the file config setting.

-Move date format detection to earlier in the index.php file to allow a
single lookup of the format for the entire file.
-Convert hardcode date string to format selection in index.php to allow
the custom format to show on the Post QSO page.
-Convert php format string to jclock format string in footer.php to
allow the custom format to show on the Live QSO page.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -498,10 +498,35 @@ $(document).on('keypress',function(e) {
           $('.input_time').jclock(options);
         });
 
+        var fmt = "<?php
+		// Borrowed from ..\qso\index.php
+		// Get Date format
+        if($this->session->userdata('user_date_format')) {
+            // If Logged in and session exists
+            $custom_date_format = $this->session->userdata('user_date_format');
+        } else {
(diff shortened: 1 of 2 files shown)


AndreasK79:
@T012m3n7oR just tested it. Works great, thanks! Could you also implement this in the contest part? QSO -> Contesting Logging.

T012m3n7oR:
@AndreasK79 That request is a little harder than this one.  the contesting logging uses a .js file for that, not .php.  I would have to work out a completely different method to get the contesting page to support a php call for format.  

Might be able to call a function with a get/post on a different .php page to get it to load it, convert it and return it?  I'm going to have to tinker that and see what I can come up with.

AndreasK79:
@T012m3n7oR you are right. That was done to reduce the size of footer.php, and to split out the different sections. A lot of the js wasn't dependent on php, so it was moved out.

If you figure out a way, it would be nice. I'll merge this now. Thanks.

PR #1205, opened 2021-09-28
[FIX] Solved a SQL syntax problem while using the latest MariaDB version.

Byloth:
It seems that the previous version of this query was misunderstood by the SQL engine.  
The key `offset` was read as a SQL keyword instead of a name of a column.

This solve this problem.

### DIFF
--- a/application/models/User_model.php
+++ b/application/models/User_model.php
@@ -362,7 +362,7 @@ class User_Model extends CI_Model {
 	// FUNCTION: array timezones()
 	// Returns a list of timezones
 	function timezones() {
-		$r = $this->db->query('SELECT id, name FROM timezones ORDER BY offset');
+		$r = $this->db->query('SELECT id, name FROM timezones ORDER BY `offset`');
 		$ts = array();
 		foreach ($r->result_array() as $t) {
 			$ts[$t['id']] = $t['name'];


AndreasK79:
I do not have this problem with MariaDB here. Though I'm using version 10.4.13. When I try your fix, the timezone list is not sorted by offset under user account, so I'm not sure what is going on here.

Byloth:
Oooh... You're right! 😅

My bad... Sorry!   
I didn't notice that, actually, the timezones aren't sorted correctly. It seemed to me because some of those are already sorted but not all... Sorry! 😥

I mixed up PostgreSQL syntax with MySQL syntax...  
Actually, written like it was, `offset` was parsed simply as a string... That's why sorting wasn't working.

I just pushed the correct fix.  
Sorry again and thank you for your patience.

---

FYI...
I'm using [MariaDB v10.6.4](https://github.com/Byloth/cloudlog-docker/blob/c1d98eee0f0b03332712ebb4adf6ddb8737bce5f/docker-compose.yml#L4) and without this fix, it returns me an exception when I visit the route `/user/add`.

image

AndreasK79:
That works much better. I'll merge it in now. Thanks for the PR.

PR #1206, opened 2021-09-29
"Mark QSL Card Not Required" sets status to Ignore but sets red arrow

Werzi2001:
... which reflects status No. Changed color to grey.

If you use the feature "Mark QSL Card Not Required" the arrow is set to red color after the ajax call is finished. Anyway the status is set to Ignore and as soon as you reload the list the arrow is grey. Changed it to be consistent (Ignore = grey after ajax call).


AndreasK79:
Thanks for noticing and fixing.

PR #1216, opened 2021-10-07
[ADIF Export] Added contest id to export

AndreasK79:


magicbug:
Perfect merge away

AndreasK79:
I will, just trying to most of the stuff in VSCode here.

AndreasK79:
Yes, it works. just needed to add a few extensions to make things a bit easier. And I need to get more familiar with how to everything.

I agree with dev/master. Guess station_logbooks needs to be finished first.

PR #1229, opened 2021-10-14
dashboard: fix RSTS/RSTR table header in mobile mode

samhjn:
origin code break on mobile mode and seems to be verbose
just do a simple modification


AndreasK79:
Thanks for the fix, and I liked the idea with the function to avoid repeating much code.

PR #1277, opened 2021-11-06
Cloudlog php8 compatible

AndreasK79:
Since this is a dev branch, and more people are installing PHP8, I made the necessary changes to get it work in PHP8.
I've installed it from scratch, and fixed the everything I stumbled onto.

The main things changed are:
* Updated CodeIgniter to the latest version on GitHub: https://github.com/bcit-ci/CodeIgniter
* Removed constructors in all models, since this is no longer supported.
* Added checks in logbook_model to check if logbooks_locations_array exists, else return null. Cloudlog would not load on a new install without this.

@magicbug I'm not merging this, as I would like to get others to test this out first. I've tested many things in Cloudlog, and I've not had any trouble, but the more eyes on it, the better.


AndreasK79:
@magicbug this doesn't work on it's own anymore, since I've fixed merge conflicts from the change I pushed to station_logbooks. This didn't bring over the changes I made.

Do you have any objections to this getting merged into station_logbooks?

PR #1286, opened 2021-11-14
Version 2 dashboard todos

Werzi2001:
This should resolve your Dashboard todos from the version 2 task list. Basically simple adaptions to the queries to add stations IN instead of = filter.


magicbug:
Thanks @Werzi2001 for crossing these ones off!

PR #1296, opened 2021-11-15
map on dashboard now shows the same QSOs as the list below

Werzi2001:
the dashboard list shows the last 11 QSOs whereas the map shows the QSOs from the last week. I find that a bit confusing so I changed it to be the last 11 QSOs in both cases. Still it might not be the same number of QSOs as the map only shows QSOs with a gridsquare.

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -99,13 +99,7 @@ class Dashboard extends CI_Controller {
 		
 		$this->load->library('qra');
 
-		//echo date('Y-m-d')
-		$raw = strtotime('Monday last week');
-		
-		$mon = date('Y-m-d', $raw);
-		$sun = date('Y-m-d', strtotime('Monday next week'));
-
-		$qsos = $this->logbook_model->map_week_qsos($mon, $sun);
+		$qsos = $this->logbook_model->get_last_qsos('11');
 
 		echo "{\"markers\": [";
 		$count = 1;


magicbug:
I'm a bit torn on this one as its always been the last week, however, you are right that it causes confusion, perhaps it should just be for the "day" or 24 hour period.

Not sure what @AndreasK79 thinks I'm not stubborn that it cant change just want to make sure its the right change

Werzi2001:
> I'm a bit torn on this one as its always been the last week, however, you are right that it causes confusion, perhaps it should just be for the "day" or 24 hour period.
> 
> Not sure what @AndreasK79 thinks I'm not stubborn that it cant change just want to make sure its the right change

Perfectly understand that. I'm not very active actually doing QSOs as my SOTA activations are usually 2-3 weeks apart. That causes my map to be empty most of the time although the list below isn't which confused me. From my point of few both should show the same QSOs. Doesn't have to be "last 11" but could also be "last week" for both (from my perspective).

But anyway I don't have "strong feelings" about that as well. Of course I can also live with the status quo :)

AndreasK79:
To make most sense, this change is correct. What you see in the log, is what you get on the map. Logbook does that, so why not the dashboard? If you need more, you could use custom maps.

Other than that, it doesn't really matter to me. Whatever it ends up with, it should be documented in the Wiki.

magicbug:
I can agree with that.. merge accepted

Issue #1289, opened 2021-11-14
Add "Send QSL" option in the coq on the logbook page


DO1DSH:
Currently there is only a hidden option to mark your sent QSO card in the QSO details. Please add the option "Sent QSL (Buero)" and "Sent QSO (Direct)" to the cogwheel of each QSO on the logbook page. This would drastically simplify QSL management.  

See also: https://twitter.com/cloudloghq/status/1459874205927890946

PR #1300, opened 2021-11-15
added "mark qsl sent" options to qso cog menu

Werzi2001:
This solves the "issue" #1289 
Anyhow the cog menu gets a bit full now. maybe we should solve it differently?


magicbug:
Thanks, yes not sure the best solution for tidying it up really, its handy to have the short cuts however, perhaps take this to a separate Discussion to figure it out

PR #1307, opened 2021-11-17
version 2 hide satellites

Werzi2001:
hide satellites selection if there are not satellite QSOs in
* statistics (task from version 2)
* distances worked (added to be consistent)
* gridsquares (added to be consistent)

Not sure if it is good as I did it for gridsquares but as in that case the statistic uses different controller actions simply hiding the field was not enough. Therefore I used a redirect if there are not sat QSOs.


AndreasK79:
I've tested this with no issues. Redirect works for me. Good work.

PR #1312, opened 2021-11-19
added eQSL and LotW to the dashboard

Werzi2001:
As I'm mostly using eQSL and LotW I'd like to these to the dashboard as well. In order for it to be of somewhat equal length I increased the number of shown QSOs to 18 (also in map).

Not sure about the layout though. As the tables for eQSL and LotW only have 3 rows (and not 4) table rows of the same column do "meet" now. Anyhow I don't know what to do about that without changing the layout away from using striped tables.


AndreasK79:
I have no objection to this. For the layout, a hack would be to use <td width='50%'>.

Tested locally, and it produces something like this:
image

What do  you say @Werzi2001 and @magicbug ?

magicbug:
Looks OK to me!

AndreasK79:
@magicbug and @Werzi2001 I've pushed a commit adding width. Will merge now.

PR #1329, opened 2021-12-22
use station profile grid locator to calculate distance

Werzi2001:
The bearing/distance in the create QSO forms currently only use the locator of the user. I think it's better to use the locator of the selected station profile as that is the location that the QSO is created for.

### DIFF
--- a/application/controllers/Awards.php
+++ b/application/controllers/Awards.php
@@ -134,7 +134,7 @@ class Awards extends CI_Controller {
 
 		$dxcclist = $this->dxcc->fetchdxcc($postdata);
         $data['dxcc_array'] = $this->dxcc->get_dxcc_array($dxcclist, $bands, $postdata);
-        $data['dxcc_summary'] = $this->dxcc->get_dxcc_summary($data['worked_bands']);
+        $data['dxcc_summary'] = $this->dxcc->get_dxcc_summary($bands, $postdata);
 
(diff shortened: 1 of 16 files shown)


magicbug:
Gets 👍🏻 from me, think so too! 

@AndreasK79 thoughts?

AndreasK79:
@magicbug I totally agree, but it needs one addition.

@Werzi2001 I was just testing it out, and couldn't get it to work. Turns out the gridsquare is fetched from QRZ (in my setup), and it uses logbook.php:694:function bearing when that is the case. Could you update this one too?

Werzi2001:
@AndreasK79 You're right, I missed that. Should be fine now. Maybe you could check as well that I got all calls of "json" (just one?). I tried to search for them but "json" is not the best search term (a lot of matches) and I might have missed one.

AndreasK79:
@Werzi2001 works as it should now, thanks. I did have a search for json. Seems to me that it's only called from qso.js, but it is difficult to know for sure. I'll dig some more tomorrow.

AndreasK79:
@Werzi2001 I had another look. I can't find any more references, so I'll merge this one in. Thanks.

AndreasK79:
@magicbug it seems like I can't merge this one. It says conflict, but it doesn't say what the conflict is. Could you take a look?

PR #1343, opened 2022-01-02
Add SOTA icon/link

phl0:
As I just logged a SOTA station and needed the summit info I thought this might be a good idea... ;-)
This time patch against station_logbooks.

Screenshot from 2022-01-02 12-50-10


Werzi2001:
@phl0 I really like that :) Anyhow target="_new" should be "_blank". "_new" is not "open new window" but "open window named _new". If you do it multiple times (leaving the window open) the same window will be used again and no new one will be opened. Details: https://stackoverflow.com/questions/4964130/target-blank-vs-target-new

phl0:
@Werzi2001 tnx for spotting. Corrected with https://github.com/magicbug/Cloudlog/pull/1343/commits/e8d6b17406a4fcabe91f8809e6a67fcae9d6794a

PR #1348, opened 2022-01-03
Update satellite_data.json with CAS-9 OSCAR designation HO-113

ve1cwj:
Update satellite_data.json with CAS-9 Oscar designation HO-113 to work with LOTW


magicbug:
@ve1cwj Now if LoTW would accept it!

PR #1359, opened 2022-01-09
added logbook filter to simple and advanced search

Werzi2001:
This should resolve the version 2 task for the search

Edit:
I implemented in a way that only QSOs of the active logbook are shown. Maybe it was meant to be "all logbooks of the user" but the task doesn't state that clearly.


AndreasK79:
@Werzi2001 thanks. I'm not really sure about this one myself, but I would think that all the logbooks for the user should be searched, not just the ones for the active logbook.

@magicbug could you enlighten us here?

AndreasK79:
@magicbug bump. This can perhaps be checked. Could you give some input to how it should be?

Werzi2001:
@magicbug Could you give us your thoughts about that? If you want to I can adapt it to be all logbooks instead of the active one.

AndreasK79:
@Werzi2001 could you just adapt it to all logbooks for the user?

Werzi2001:
@AndreasK79 No problem, I adapted it accordingly. Please note that I also had to adapt get_qso($id) in Logbook_model because that also filtered the active logbook. Without that it would not be possible to open the details of a QSO (click on callsign) of a QSO that is from another logbook (but of the same user).

AndreasK79:
@Werzi2001 thanks. Just tested it out. As far as I can tell, it works as it should. I've tested it with several users, logbooks and locations. 

@magicbug seems ok to merge to me. Like to check it out first?

AndreasK79:
@magicbug any objections to this one getting merged?

magicbug:
Go for it

PR #1363, opened 2022-01-14
Order QSO in map details by time descending

phl0:
Would you mind sorting the QSO list for gridsquare map details by time and descending by default? :)


AndreasK79:
@phl0 I agree.

PR #1373, opened 2022-01-20
Change how comments in CL show on eQSL.

M1SJE:
Adding the ability to send comments in Cloudlog to show on eQSL.

They use QSLMSG, and if you send anything in the COMMENT section, like signal reports from WSJT-X, it will be ignored by eQSL. 

So this edit adds reading the comments section and will upload it as QSLMSG.

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -451,6 +451,15 @@ class eqsl extends CI_Controller {
                     $adif .= "%20";
 				}
 
+                 if ($qsl['COL_COMMENT'] != ''){
+                    $adif .= "%3C";
+                    $adif .= "QSLMSG";
+                    $adif .= "%3A";
+                    $adif .= strlen($qsl['COL_COMMENT']);
+                    $adif .= "%3E";
+                    $adif .= $qsl['COL_COMMENT'];
+                    $adif .= "%20";
+                                }
 
 				# Tie a bow on it!
 				$adif .= "%3C";


magicbug:
My only issue is that I dont use the "comment" in the QSO window for QSL comments it could just be I've added what they said their running.. hence I don't think this is appropriate.

It could be argued we need a new field QSL_Comment that's stored in the "QSL" tab.

WSJT-X reports should be shown correctly as the reported reports so there should be zero need to send this info twice.

magicbug:
@phl0 @AndreasK79 Not sure your views on this one or maybe I'm being pedantic

phl0:
Me is not a fan of eQSL at all but wouldn't mind having added this as a separate field in the QSL tab. I use the existing comment field only for my own purposes which don'nt need to appear on the QSL card. :-)

AndreasK79:
I agree with @magicbug. We need to add another field for this. "Magic" things happening like pushing comments to upload is not good.

magicbug:
Yeah, easily to slip up and write "lid" in the comment field.. however also could be useful for when I finally get labels integrated as it can be used for comments on those tooo

M1SJE:
Guess I never use the comments for that, but for putting the signal reports from WSJT-X in I do, and what I was doing before with Log4OM and when I moved onto Cloudlog was edit the DB directly copying everything over from comments to QSL message.

Each to their own I guess :)

phl0:
Something like this maybe? 


Screenshot from 2022-01-24 23-46-15

phl0:
Just copied some lines of code for display purposes .... @M1SJE  If you wanna take a look (and take it as a start):

https://github.com/phl0/Cloudlog/commit/a971434f40dc8d2142541f5dd464888fee0e33ef

magicbug:
@phl0 Yeah along those lines, it will need a field added to the log column so this is probably a V2 branch addon so I don't have to battle out of number migration files :)

phl0:
Oh yes. Forgot to mention that my quick and dirty patch was based on station_logbooks branch. Tbh I didn't notice that this PR wasn't. Hope we didn't upset @M1SJE 😇✌️

M1SJE:
Nah, you didn't, it's just I like to put the signal reports on the qsl message, and the easiest way of doing that is by doing that when uploading to eqsl.

The reasons stated for why it's not done like that by default make sense, although why anyone would be putting insulting comments into their log book about someone is beyond me, although I can see how getting a QSL card with "Utter wanker" could be seen as unfriendly :P.

phl0:
Roger that. I am happy to test :)

PR #1374, opened 2022-01-20
Cleanup SQL CHARSET and COLLATE

onovy:
* Set same charset and collate for all tables
* Recommends collate utf8mb4_0900_ai_ci which is MySQL default

This clean up different CHARSET and/or COLLATE in DB. For example my DB had different COLLATE for dxcc_prefixes and TABLE_HRD_CONTACTS_V01 which resulted in not working CSV export of requested QSLs. Error message was "Illegal mix of collations", because it's not possible to "LIKE" strings with different collations.

I changed all CHARSET to utf8mb4 and all COLLATE to utf8mb4_0900_ai_ci (which are MySQL recommended defaults).


magicbug:
@onovy This looks fantastic I really don't want to add a new migration file to "master" branch at the moment as I'm readying version 2.0 of cloudlog so would prefer only to merge this in with "station_logbooks" branch which is V2. 

Not sure if you're willing to rebase this for that branch or would prefer I did it, however would also recommend you checkout that branch too there's quite a bit going on!

onovy:
Rebased, https://github.com/magicbug/Cloudlog/pull/1382

I will look into new branch, thanks for pointing out.

PR #1382, opened 2022-01-25
Cleanup SQL CHARSET and COLLATE

onovy:
* Set same charset and collate for all tables
* Recommends collate utf8mb4_0900_ai_ci which is MySQL default


magicbug:
In  which case something odds happening as both my installs don't have said table!

PR #1387, opened 2022-01-31
Fix QSO with time_on < time_off and QSO with date_off field

onovy:


application/models/Logbook_model.php
```diff
@@ -1645,9 +1645,19 @@ function import($record, $station_id = "0", $skipDuplicate = false, $markLotw =
         $time_on = date('Y-m-d', strtotime($record['qso_date'])) ." ".date('H:i:s', strtotime($record['time_on']));
 
         if (isset($record['time_off'])) {
-            $time_off = date('Y-m-d', strtotime($record['qso_date'])) ." ".date('H:i:s', strtotime($record['time_off']));
+            if (isset($record['date_off'])) {
+                // date_off and time_off set
+                $time_off = date('Y-m-d', strtotime($record['date_off'])) . ' ' . date('H:i:s', strtotime($record['time_off']));
+            } elseif (strtotime($record['time_off']) < strtotime($record['time_on'])) {
```
onovy, line 1651: This is explained in ADIF spec (https://adif.org/312/ADIF_312.htm#QSO_Field_QSO_DATE_OFF)

in the absence of <QSO_DATE_OFF>, the QSO duration is less than 24 hours.  For example, the following is a QSO starting at 14 July 2020 23:55 and finishing at 15 July 2020 01:00:<QSO_DATE:8>20200714 <TIME_ON:4>2355 <TIME_OFF:4>0100
--

PR #1388, opened 2022-01-31
Fix ADIF export

onovy:
* Fix ADIF format to specification
* Export (almost) all ADIF fields from DB

### DIFF
--- a/application/libraries/AdifHelper.php
+++ b/application/libraries/AdifHelper.php
@@ -3,13 +3,144 @@
 class AdifHelper {
 
     public function getAdifLine($qso) {
-        $line = "";
-        $line .= $this->getAdifFieldLine("CALL", $qso->COL_CALL);
-        $line .= $this->getAdifFieldLine("BAND", $qso->COL_BAND);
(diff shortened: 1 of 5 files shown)


application/libraries/AdifHelper.php
```diff
@@ -3,13 +3,144 @@
...
+        $normalFields = array(
+            'ADDRESS',
+            'AGE',
+            'A_INDEX',
+            'ANT_AZ',
+            'ANT_EL',
+            'ANT_PATH',
+            'ARRL_SECT',
+            'AWARD_GRANTED',
+            'AWARD_SUMMITED', // Typo in DB!
```
onovy, line 15: going to fix this with next MR (to v2), because needs to rename field in DB

AndreasK79:
@onovy tested working ok with export and import.

Any reason you put each field on a separate line? And why the double line shift after eor?
I can see it's more human readable, so I don't mind, just curious.

@magicbug as far as I can tell, it's good for merge.

onovy:
@AndreasK79 as field/entry separtion. It's much more simple to check export now, (GNU) diff it and it's readable (if needed). So yep, I changed it by intention :)

PR #1392, opened 2022-02-04
file manager: add a file manager layer to handle qsl card storage & etc

samhjn:
Added a file manager driver to enable both local file storage & AWS S3 competible object storage.

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -53,6 +53,7 @@ $autoload['packages'] = array(APPPATH.'third_party');
 */
 
 $autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Frequency');
+$autoload['drivers'] = array('File_manager');
 
 
 /*
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 73;
+$config['migration_version'] = 74;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/File_manager.php
+++ b/application/controllers/File_manager.php
@@ -1,0 +1,25 @@
+<?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
+class File_manager extends CI_Controller
+{
+
+	/* Controls who can access the controller and its functions */
+	function __construct()
+	{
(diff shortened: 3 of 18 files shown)


samhjn:
Frontend settings pages may be done in the future.......

magicbug:
Looks good, however, I'm not currently merging anything which a database change into "master" at the moment, this is while I get things changed over for version 2.0 release with new features this is the "station_logbooks" branch so if you'd like to put it as a PR for that branch it will get merged in :)

Sorry I should have added a note to the Readme about this!

samhjn:
Okay, I will create a new PR after tested on that branch......

magicbug:
Added note to readme while i rembered!

samhjn:
I have to say..... 
The "station_logbooks" branch is more unstable than I expected 😅.......
The patch may take a while after I figure out how to make the qsl card function work.......

Here are some bugs I have encountered, hope it will help avoid future confusion:
a) migrate 083 are totally the same as 073, which results in a confliction
b) migrate 085 fails on utf8mb4_general_ci, with varchar(200) exceeds the key length limitation
c) unhandled list_logbook_relationships empty return on many logbook relative pages

AndreasK79:
@samhjn it's not released yet, and that's one of the reasons why. Bugs exists.
a) This is easy to fix. The reason was that a script was added in master that was not in station_logbooks. Those coming from master, would not get script 73 executed. I'll fix this now.
b) seems 191 bytes are the limit for utf8mb4. https://stackoverflow.com/questions/1814532/mysql-error-1071-specified-key-was-too-long-max-key-length-is-767-bytes @magicbug do we need 200 bytes, or is 191 sufficient?
c) can't say I've encountered that one. It would be nice if you could provide more details here.

magicbug:
I'd hope 191 was enough unless told otherwise!

C) make sure you have claimed your Station Locations and then setup a Station Logbook with some assigned Locations.

The branch is stable, its my "current" logging live branch so I soon notice things :)

samhjn:
> I'd hope 191 was enough unless told otherwise!
> 
> C) make sure you have claimed your Station Locations and then setup a Station Logbook with some assigned Locations.
> 
> The branch is stable, its my "current" logging live branch so I soon notice things :)

Well it is caused by no station logbook......
But it just throw out uncaught exceptions...... I thougt the false return value of list_logbook_relationships should be handled somewhere......

PR #1393, opened 2022-02-05
file manager: add a file manager layer to handle qsl card storage & etc

samhjn:
Cherry pick works of file manager on 1.0 branch to 2.0

Note: There is a bug in original code that some of qsl model interface privilege checker mixed qsl_id with qso_id. I modified them to let the interface work, but there obviously need another patch to fix it.


samhjn:
There is a problem that the file name generator may produce same random at a chance of 1/65536 if both back & front are uploaded. A bigger random length may help reduce the chance of collision.

application/models/Qsl_model.php
```diff
@@ -44,41 +46,41 @@ function saveQsl($qsoid, $filename) {
...
         // Clean ID
-        $clean_id = $this->security->xss_clean($id);
+        $clean_id = $this->security->xss_clean($qsl_id);
 
         // be sure that QSO belongs to user
         $CI =& get_instance();
         $CI->load->model('logbook_model');
         if (!$CI->logbook_model->check_qso_is_accessible($clean_id)) {
-            return;
+            //return;
```
AndreasK79, line 65: Why did you comment out this? This must be there to check that the user is allowed to delete the qsl. If no rights, return.

application/models/Qsl_model.php
```diff
@@ -44,41 +46,41 @@ function saveQsl($qsoid, $filename) {
...
```
AndreasK79, line 80: Why did you comment out this? This must be there to check that the user is allowed to view the qsl. If no rights, return.

AndreasK79:
@samhjn thanks for the PR. I've tested it, and as far as I can tell, it works as it should with no errors. I've only tested it with storing QSLs on the same server (my local machine). The migration script also takes care of the existing QSLs, so they are in place after migration (at least with my test).

application/models/Qsl_model.php
samhjn: Here should pass $qso_id instead of $qsl_id, the qso accessible check here may block some of request unexpectedly.

application/models/Qsl_model.php
samhjn: Here should pass $qso_id instead of $qsl_id, the qso accessible check here may block some of request unexpectedly.

application/models/Qsl_model.php
AndreasK79: I see, that's what you meant with needing another patch?

samhjn: Yes, a new query may needed here.

application/controllers/File_manager.php
```diff
@@ -0,0 +1,25 @@
...
+			redirect('dashboard');
+		}
+	}
+
+	public function index()
+	{
+		$data['page_title'] = "File Manager";
+
+		$this->load->view('interface_assets/header', $data);
+		$this->load->view('file_manager/index');
```
magicbug, line 22: This folder and view doesn't exist.

application/controllers/File_manager.php
```diff
@@ -0,0 +1,25 @@
+<?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
+class File_manager extends CI_Controller
```
magicbug, line 2: Not sure if this was tested but usually CI forces class names for controllers to be without special characters e.g "_"  in its current state this controller returns `Unable to locate the specified class: File_manager.php`

magicbug (changes requested):
Good start, seems to be missing some files and view areas seem untested and not referenced within the menu system. I'm guessing File manager would only be available to "Admin" if it's storing AWS data but can't see a reference to that.

magicbug:
> There is a problem that the file name generator may produce same random at a chance of 1/65536 if both back & front are uploaded. A bigger random length may help reduce the chance of collision.

I'd suggest we implement this to stop any chance!

application/controllers/File_manager.php
samhjn: Yes, I may implement them in the future......

application/controllers/File_manager.php
samhjn: In fact this file was not intended to be uploaded 😢 . In my plan, the project was devided into 2 parts. One parts for the model & driver, another for view & template. The pathset was for the model & driver.

samhjn: From Jan. 31 to Feb. 6 is a traditional vacation here for me. In this period of time I may finish the most key part of the project, afterwards, I may finish the rest of the functions in late Feb. or Mar.
Would you allow to merge them as two patch? I may prefer to delete the controller in this patch, and complete it in the next one, or I may push the end to end code together later.

samhjn:
> > There is a problem that the file name generator may produce same random at a chance of 1/65536 if both back & front are uploaded. A bigger random length may help reduce the chance of collision.
> 
> I'd suggest we implement this to stop any chance!

I came up with serveral solution for this:
1. extend the random str length to 8 or more, which can lower the collision chance to less than 1/4billion, which seems enough in many cases.
2. add qslfront/qslback to filename, which can avoid collision in normal use case, but it can not prevent bad case created by automatic scripts posting requests too frequent.
3. check s3 file existance before write, for s3 does not provide atomic check before write apis, it may also fail on some automatic scripts.
4. modify the filename with $file_id, it will insert mysql first to get a $file_id, a transaction may need here to handle file upload failure.

I'd prefer 1 or 4, which do you prefer?

magicbug:
Option 1 seems reasonable :)

application/controllers/File_manager.php
samhjn: The unfinished controller has been removed.

samhjn:
> > There is a problem that the file name generator may produce same random at a chance of 1/65536 if both back & front are uploaded. A bigger random length may help reduce the chance of collision.
> 
> I'd suggest we implement this to stop any chance!

now length=10 was set for the random hex

magicbug (approved):

magicbug (changes requested):
Okay on not having the views in place, however for most users they will be a much-needed item so hopefully, we can work together and finish those parts.

It could be that it could be stored within Global Settings under "Admin"

samhjn:
> Okay on not having the views in place, however for most users they will be a much-needed item so hopefully, we can work together and finish those parts.
> 
> It could be that it could be stored within Global Settings under "Admin"

Okay, I will work on it later... The weekend time is precious 🥲 , and I may have to take some time to repick my memory of jQuery.ajax and PHP request handling, I may need some time...

Thank you

samhjn:
Sorry for the delay of the frontend of this feature. Now I finished a version that is able to run end to end on former branch.
I cherry-picked the code to recent branch to carry out more test, but I had trouble to run the current branch. It seems the install process failed on the migration procedure. I recieved error message below:

> A Database Error Occurred
Error Number: 1146
Table 'cloudlog.options' doesn't exist
SELECT * FROM `options` WHERE `autoload` = 'yes'
Filename: D:/sdk/xampp/htdocs/Cloudlog/system/database/DB_driver.php
Line Number: 656

Could anyone reproduce this problem?

AndreasK79:
Did you run this code against master? I know that it does not work with PHP8. If not, then I'm not sure what is causing this. 

I'll see if I can give this PR a test run one of these days.

samhjn:
> 



> Did you run this code against master? I know that it does not work with PHP8. If not, then I'm not sure what is causing this.
> 
> I'll see if I can give this PR a test run one of these days.

No, I run it with station_logbooks. It fails during installation.

AndreasK79:
That is strange. I just tried this branch, and it fails with the same problem here.

AndreasK79:
I found the problem. It seems to me that you based the code on a broken commit of station_logbooks.
You need to edit 073_create_station_logbook_relationship_table.php and rename the class from Migration_create_station_logbook_relationship_table2 to Migration_create_station_logbook_relationship_table.

samhjn:
> I found the problem. It seems to me that you based the code on a broken commit of station_logbooks. You need to edit 073_create_station_logbook_relationship_table.php and rename the class from Migration_create_station_logbook_relationship_table2 to Migration_create_station_logbook_relationship_table.

You are right, it seems that I submitted some local fix of station_logbooks branch few months ago. Tests are carrying out now.

samhjn:
@AndreasK79 @magicbug 
I think the frontend now is able to use. If there is any other things to do before the pull request being accepted?

AndreasK79:
@samhjn sorry for not getting back sooner. I'll take a look this weekend.

AndreasK79 (approved):
Tested out ok on my local copy. Everything seems to be working as it should regarding QSL card management here.

I can't test out AWS storing of cards, as I don't have an account.

AndreasK79:
@magicbug did you check this out? Should we get this in v2 branch?

PR #1398, opened 2022-02-06
Activated grids

phl0:
This adds one more item in the Analytics section showing activated (SAT) grids

FK7DALYX0AY8Dlr


AndreasK79:
Thumbs up from me. Tested working ok.

phl0:
Tnx feedback.

magicbug:
Looks great, just one problem that we don't use COL_MY_GRIDSQUARE in Cloudlog or at least I don't populate it, exports ADIF gets its gridsquare from the station_profile tables, station_gridsquare column so for me maps completely empty.

magicbug (dismissed):
Looks great, just one problem that we don't use COL_MY_GRIDSQUARE in Cloudlog or at least I don't populate it, exports ADIF gets its grid square from the station_profile tables, station_gridsquare column so for me maps completely empty.

the COL_MY_Grid might have content from old imports but generally, its ignored even by LoTW uploads and eqsl etc.

The ADIF import feature will in theory at some point create a "station Location" automatically for a user but currently it doesn't (probably a bad thing its not supported yet)

phl0:
Uugghh. Now that's bad. How come mine works then? It's not selely imported ADIF data in my log. Need to investigate...

onovy:
technically, COL_MY_GRIDSQUARE is populated (https://github.com/magicbug/Cloudlog/blob/master/application/models/Logbook_model.php#L207), but we should not use it. ```station_profile``` is correct place (table).

phl0:
I see. But shouldn't @magicbug's tables have that column filled then?!

onovy:
create_qso is called only during manual entry of QSO (thru web interface), not during ADIF import

magicbug:
It needs to be deleted if anyone fancies doing a migration file for it plus most of the MY_ fields they need to come from Station Location table.

magicbug (changes requested):
> Think my first comment got a bit of a mixed understanding so hopefully going to explain it better.

This is fantastic and a tool lots of **rover** style operators would want, while Cloudlog does has the `COL_MY_GRIDSQUARE` & `MY_VUCC_GRIDS` within the `TABLE_HRD_CONTACTS_V01` table and they are populated during ADIF importing overall I'd prefer we didn't use them and prefer `COL_MY` columns got completely removed from the main QSO table.

This is because it's better handled separately within the `station_profile` table and as you can see from its schema it has a lot of the fields, grid squares are stored in the `station_gridsquare` column.

The column has been designed to have a single grid square or multiples being comma-separated, this table is used to generate all the MY columns on ADIF export but also building uploads for LoTW which needs some exact info.

Anyways going back to my comment, while I appreciate its a pain and means a rewrite of some of the queries but I'm willing to of course help with this.

Hopefully explained things better.

magicbug:
Sometimes one shouldn't write feedback when brains foggy for various reasons ;-)

That's a good question, I'm tempted to say all logbooks, but I can also see some wanting just a certain one.

phl0:
Ok thank you @magicbug. No worries. I'll try to switch from COL_MY_GRIDSQUARE to station locations first. The question about with station logbooks to take into account can be discussed later then.

phl0:
First step going to using station_gridsquare as sources for grids ...

magicbug:
Looking good @phl0 just need to support comma-separated grids and its perfect :D

magicbug:
Step in the right direction how flo!

With the popup window, when clicking on squares users might come into an issue where there are too many QSOs, for example, my home grid I get

```
<p>Severity: Error</p>
<p>Message:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 12288 bytes)</p>
<p>Filename: mysqli/mysqli_result.php</p>
<p>Line Number: 229</p>
```

However, if I just open up the band and go to 2m it loads fine as there are only a couple of QSOs.

I'm wondering if this ajax call needs to do a quick count on the number of QSOs it will display and if its over X amount only show the last x number of QSOs maybe @AndreasK79 has a view on this one.

phl0:
Ah confirmed. It also loads slow here on JO31 with 4k QSOs. I think limiting it to the last x QSOs or distinct callsigns is a good idea.

AndreasK79:
@magicbug and @phl0 yes, we need to limit it. I would say limit to 500, sorted by date, latest first. I don't think distinct callsigns would be enough.

phl0:
I will add a LIMIT to the SQL command.

dg9vh:
>
>
> Combination of select distinct, sort and limit by would be best.
-- 
Amateurfunk ist das, was uns verbindet! Bleiben wir in Kontakt!
DARC e.V., Ortsverband Q03, Neunkirchen

phl0:
So limit to 500 is committed with https://github.com/magicbug/Cloudlog/pull/1398/commits/af1451b5421141ea3c43f11a52034301f129d021. Distinct callsigns seems to be a heavy SQL task ... Trying ...

phl0:
Found a way to only show the last 500 unique callsigns: https://github.com/magicbug/Cloudlog/pull/1398/commits/8d0984857d5f4061dd08370059a5025673482817 Not sure if I like the code though.... 
Comments?

magicbug:
Not the prettiest but functional :) 

The dialogue boxes now show on my copy!

magicbug (approved):

phl0:
> Not the prettiest but functional :)

Feel free to improve ;-)

magicbug:
I'm happy to merge it as is as long as @AndreasK79 agrees with me for some sanity :)

AndreasK79:
Must be my version of MariaDB, since I get no popup. It fails:
FUNCTION cloudlog.ANY_VALUE does not exist

That means every call to qso_details_ajax fails as well, since it is used several places.

Need to see if it is possible to speed up the query in another way.

AndreasK79:
Yeah, same with REGEXP_LIKE :(

But really, do we need distinct? If we had fetched call, band and mode, that might have made some difference, but we are fetching far too many columns. The only difference distinct will make, is to exclude duplicate entries in the table, which is nearly non-existent. Either an ADIF-import has been done several times, or F5 has been pressed when logging QSOs.

AndreasK79:
Then I think this needs to be split into it's own function.
If you worked the same call on several bands, only 1 QSO would be listed, and that is the point of this function in this scenario. 

However, this function is in use several places, and I for one would like to see all QSOs, not only distinct calls.

phl0:
hhmmm but only for the activated grids section but not for other awards sections then?

magicbug:
> hhmmm but only for the activated grids section but not for other awards sections then?

Yes, handy to have dupes in the awards bit for finding callsigns to annoy for cards!

AndreasK79:
Yes, give activated grids it's own function to fetch QSOs (like it is in this PR, just rename it), and copy back the old one.

phl0:
Ok, I moved that to a separate function now. Only need to find out how to work around ANY_VALUE for MariaDB dbs...

phl0:
Hhmmm I wanted to list every last QSO per callsign and max. 500 items... Need to re-think about dlthe SQL query...

AndreasK79:
I have an idea for a query. Subquery that fetches id's for the QSOs. Use that to fetch information for the QSOs you want to display. I can see if I can write the query tomorrow.

AndreasK79:
@phl0 I think something like this should work. It's very simplified, but it shows the method.

> select * 
from table_hrd_contacts_v01 thcv
join (select max(COL_PRIMARY_KEY) maxkey, col_call  
	from table_hrd_contacts_v01
	group by col_call
	order by COL_TIME_ON desc
	limit 100
) x on thcv.COL_PRIMARY_KEY = x.maxkey

magicbug:
I'd try not to use `SELECT * `all the time as it's quite an expensive process between the client and the server but rather select only the fields that are needed to complete the task.

AndreasK79:
@magicbug no objection there. It was more to show how it can be done, but very stripped down, so parameters etc need to be added, and the columns to fetch needs to be specified.

phl0:
> @phl0 I think something like this should work. It's very simplified, but it shows the method.
> 
> > select *
> > from table_hrd_contacts_v01 thcv
> > join (select max(COL_PRIMARY_KEY) maxkey, col_call
> > from table_hrd_contacts_v01
> > group by col_call
> > order by COL_TIME_ON desc
> > limit 100
> > ) x on thcv.COL_PRIMARY_KEY = x.maxkey

Working on that. It seems this gives the same error as I had before concerning only full group-by. Will try to solve...

Screenshot from 2022-02-23 16-22-49

phl0:
Committed a refactored SQL query to include the last 500 QSOs. See https://github.com/magicbug/Cloudlog/pull/1398/commits/82cd15bc07cf5403b31ee552afd79c0caa8ad09d What do you think?

phl0:
Maybe make the limit configurable in config.php?

AndreasK79:
> Committed a refactored SQL query to include the last 500 QSOs. See [82cd15b](https://github.com/magicbug/Cloudlog/commit/82cd15bc07cf5403b31ee552afd79c0caa8ad09d). What do you think?

I think that is as good as it gets, since only_full_group_by is active.

Just need to fix the hardcoded table :)

phl0:
Opppz. Fixed. :-)

AndreasK79:
@phl0 it fails, but I don't think that's MariaDB's fault:
image

phl0:
Ah. Maybe because you selected other columns to show in the config?

AndreasK79:
Yes. That was the reason. Easy fix when I figured it out.

phl0:
Well that will most likely fail then for others with other selections.... 😭

phl0:
Oki good news then. Thank you.

phl0:
Then we are keen to read from @magicbug 's test with large DB :)

magicbug:
Shall test it against my log in the morning fingers crossed its OK :)

magicbug:
I can confirm this works fine with my large logbook @phl0, @AndreasK79 do we think this is OK for merge?

AndreasK79:
@magicbug as far as I can tell, merge should be ok no.

magicbug:
Lets give it a whirl :)

phl0:
Looks good to go from my side.

PR #1409, opened 2022-02-21
[QRB Link] Fixed correct home locator. Now uses the one logged.

AndreasK79:


Werzi2001:
@AndreasK79 Interesting, didn't even know this exists. As far as I can tell from the code that should be the grid locator in the logbook/dashboard lists. Anyhow in my system these are just plain text without the link around them. Do I have to enable something to get that?

AndreasK79:
@Werzi2001 I added this very recently into station_logbook. Spawn with ALT-Q.
Then I added the globe icon to spawn the dialog from logbook, dashboard and QSO dialog.

However, I didn't think to use the connected gridsquare from the location (an error on my part). I added so that it fetches the gridsquare from the station location connected to the QSO.

Could you elaborate on what you mean by plain text?
The globe link looks like this: 
image

Werzi2001:
@AndreasK79 I see. I thought that this was an "older" feature and you just added the fix now to use the correct grid square. My system is not 100% up-to-date so the feature just isn't in there already.

AndreasK79:
@Werzi2001 fix for a new feature :)
image
image

PR #1422, opened 2022-03-09
[Custom map] Responsive map

AndreasK79:
Small changes to be able to use custom map id.
This allows for setting of height in custom css.

Added CSS for height in custom map to be relative.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -645,7 +645,7 @@ function getLookupResult() {
         <?php } ?>
 
         var qso_loc = '<?php echo site_url('map/map_data_custom/');?><?php echo rawurlencode($date_from); ?>/<?php echo rawurlencode($date_to); ?>/<?php echo rawurlencode($this->input->post('band')); ?>';
-        var q_zoom = 2;
+        var q_zoom = 3;
 
       $(document).ready(function(){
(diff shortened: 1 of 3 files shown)


phl0:
What could be tested here? Overwriting height value in custom.css?

AndreasK79:
I guess that would be the only thing in this case.

phl0:
I just set 

```
.custommap {
   height: 400px;
}
```

in my template's overrides.css to no avail. What am I doing wrong?

AndreasK79:
It's and ID not Class. Try
```
#custommap {
   height: 400px;
}
```

phl0:
No change. I expected the map to be smaller as per default but cannot see a difference.

AndreasK79:
Thanks for testing. I removed !Important. Pull latest commit, it should work now.

phl0:
Screenshot from 2022-03-17 12-48-47

Seems the overriding still does not work...

AndreasK79:
```
#custommap {
	height: 400px !important;
}
```

Add !important like above.

phl0:
Ok, now confirmed working! Tnx

phl0:
Go for merge :)

AndreasK79:
Thanks for testing. @magicbug any objections?

PR #1431, opened 2022-03-16
Add SOTA csv export function

phl0:
This adds another export function that filters QSOs with SOTA information and allows for upload to SOTA database.


magicbug:
My only comment as I need a wiki page for it :)

phl0:
Here you go: https://github.com/magicbug/Cloudlog/wiki/SOTA-CSV-Export

PR #1443, opened 2022-03-25
Fix options lib to return values not named 'options_*'

phl0:
Switching themes for global appearance options does switch values in database but will not correctly display them on the config page and always show "Blue' as (default because first item in the list) theme.

Reason:
The options library would not return configuration items not having 'options_' as prefix in their name. Instead it would return config items from the config file. If the file does not contain that item (e.g. theme) nothing is returned. In this case the options lib is queried for "theme" but as that does not have the "options_" prefix it looks for "theme" in the config file and as that also fails the return result is simply empty.

I think @AndreasK79 was meant to have an if - else if - else structure in the lib when he changed that. Could you please confirm?

### DIFF
--- a/application/libraries/OptionsLib.php
+++ b/application/libraries/OptionsLib.php
@@ -56,15 +56,10 @@ class OptionsLib {
     
                 // return option_value as a string
                 return $options_result;
-            } else {
-                return $CI->config->item($option_name);
             }
-        } else {
-
-                return $CI->config->item($option_name);
         }
 
-        if(!$CI->config->item('option_'.$option_name)) {
+        else if(!$CI->config->item('option_'.$option_name)) {
             //Load the options model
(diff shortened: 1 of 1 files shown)


phl0:
Updated this PR to also load default template from database instead of file as the options were moved from file to database. The config file is read as last resort though.

magicbug (commented):
Completely breaks other parts of Cloudlog like maps etc

phl0:
Ooops. Let me check.

magicbug:
This might now be fixed in the station_logbooks branch just set theme option to auto load and it seems to load correctly

phl0:
Was fixed with eaf814b336ff3f4b368b4ef19e2ca2e1446334eb and previous commit. So cancelling here. Tnx fer fix.

PR #1444, opened 2022-03-27
post contest logging using manual parameter

Werzi2001:
Pull request for the discussion #1441 . I tried adding the new button as proposed but didn't succeed (or better: would be more work than thought) as it would require starting/stopping the realtime clock (jclock) instead of just adding/removing it on the server side (as it is done at the moment). For my requirement this solution would be enough. If you prefer the button (general solution) I'd dig a little deeper into the code.


AndreasK79:
Tested working ok. I see no reason to merge. @magicbug ?

PR #1445, opened 2022-03-28
Remove regex

phl0:
and make sure that only valid grid locators (and grid lines/corners) are entered into the database. 
Therefore implemented a check of the input values. It allows:
- no grid at all (i.e. empty value)
- 2-digit grids
- 4-digit grids
- 6-digit grids
- 8-digit grids
- grid line (i.e. two 4-digit grids separated by comma)
- grid corner (i.e. four 4-digit grids separated by commas)

### DIFF
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -44,6 +44,7 @@ class QSO extends CI_Controller {
 		$this->form_validation->set_rules('start_date', 'Date', 'required');
 		$this->form_validation->set_rules('start_time', 'Time', 'required');
 		$this->form_validation->set_rules('callsign', 'Callsign', 'required');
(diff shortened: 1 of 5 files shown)


magicbug:
Looks good, my only concern is its extra text on the frontend thus increased space, could we perhaps chuck this as a tooltip?

chrome_6NLNCOrsou

Would also be nice todo the gridsquare validation on the frontend using JS so its realtime too but beyond scope of the task at the moment :)

magicbug:
When you first visit the Gridsquare Activators area you automatically get an error see below which can be off-putting and tends to cause support requests. needs to be hidden until there is an actual error

chrome_KDdO28nGso
.

PR #1446, opened 2022-03-29
search function for Qrz.com now has a default for $use_fullname

OE2WNL:
the call to
[index.php/update/check_missing_grid](https://cloudlog.mydomain.com/index.php/update/check_missing_grid)
threw an error because of the missing 3rd parameter - this PR fixes this as it sets the 3rd parameter ($use_fullname) optional with true as the default

### DIFF
--- a/application/libraries/Qrz.php
+++ b/application/libraries/Qrz.php
@@ -54,7 +54,7 @@ class Qrz {
 	}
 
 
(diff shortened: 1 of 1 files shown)


magicbug:
I'm going to suggest as its optional the default would always be "false" unless otherwise sent.

PR #1457, opened 2022-04-14
Prefill DOK field from previous QSO

phl0:
This adds a lookup function for German callsigns so that the DOK field is automagically prefilled with the DOK from the previous QSO. Saves some time to not enter it again and again. If last QSO did not contain a DOK value it is left empty (i.e. DB lookup does not disregard emtpy fields), so that if someone leaves the club the field is not filled with the former DOK.

### DIFF
--- a/application/controllers/Lookup.php
+++ b/application/controllers/Lookup.php
@@ -112,4 +112,20 @@ class Lookup extends CI_Controller {
 
 	}
 
+	public function dok($call) {
+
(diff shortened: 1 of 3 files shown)


dg9vh:
Good idea!

phl0:
> Maybe we should add this also for QSL Manager?

I think this is taken from the callbook lookups anyway. But I would have to re-read the code here.

Issue #1066, opened 2021-06-16
US county does not populate from Callbook


tarantir:
When logging a US callsign the US county does not populate.

1. Ensure Cloudlog callbook is setup (I'm a QRZ Logbook XML subscriber)
2. Enter a us callsign (e.g. N8AUU) on the live QSO screen 
3. Tab to let callbook retrieve the callsign details

**Expected behaviour**
US County will populate on the QSO general tab.

I've attempted to debug this without luck.

**Desktop (please complete the following information):**
 - OS: Mint Linux 20.04 and Windows 10
 - Browser Chrome (latest)
 - Version Latest cloudlog from Repo

phl0:
Thought I'll make a quick fix here due to latest experiences with auto-filling German DOKs. This should be fixed by PR https://github.com/magicbug/Cloudlog/pull/1460

phl0:
Its now in station_logbooks.

PR #1460, opened 2022-04-16
Fill US county from callbook / logbook

phl0:
This adresses issue https://github.com/magicbug/Cloudlog/issues/1066 and uses mechanism from DOK fill feature for populating US county with content.

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -149,6 +149,7 @@ class Logbook extends CI_Controller {
 			$return['callsign_iota'] = $this->logbook_model->call_iota($callsign);
 			$return['qsl_manager'] = $this->logbook_model->call_qslvia($callsign);
 			$return['callsign_state'] = $this->logbook_model->call_state($callsign);
+			$return['callsign_us_county'] = $this->logbook_model->call_us_county($callsign);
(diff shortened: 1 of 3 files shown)


magicbug (commented):
My only comment and it probably affects a few of the functions is that will check ALL qsos in the table rather than just the ones associated with the user/logbook I mean its for a county / state so we could argue it helps provide more data but option to anyone else's thoughts on this

phl0:
But that may give more than one result then?

magicbug:
yup

phl0:
Ok,  how would you decide which one to take then?

AndreasK79:
Order by date and limit to 1 result. The last one is the most probable one.

phl0:
> Order by date and limit to 1 result. The last one is the most probable one.

Jupp that is what I imoplemented.

AndreasK79:
> > Order by date and limit to 1 result. The last one is the most probable one.
> 
> Jupp that is what I implemented.

Ah, I see that now :) Then I think this is fine as is.

AndreasK79:
@phl0 tested working ok after query fix. Merge time :)

phl0:
Oops. Must have been drunk while coding :) Tnx fix.

PR #1461, opened 2022-04-16
Adding MO-112 digipeater

ryandzink:


magicbug:
This is what happens when I'm not in FP :-) thanks for adding

ryandzink:
Yikes... I typo'd the name of the sat 🤦

PR #1466, opened 2022-04-27
add station profile name to qso detail

Werzi2001:
I think it makes sense to add the station profile name in addition to the station callsign. Personally I have a lot of stations with the same callsign but different names. Only the callsign is therefore not distinguishable for me.


phl0:
I like the idea :)

AndreasK79:
Agreed. Merging.

PR #1470, opened 2022-05-04
added possibility to sort locations by id

Werzi2001:
In the pull request #1243 I changed the server side sort of the logbooks and stations list to be in alphabetical order instead of creation (ID). During the last months I came to the conclusion that this might have been an "error". Elsewhere the list is not sorted alphabetical (e.g. when assigning a location to a logbook or in the dropdown to select the location when creating a QSO). Therefore it is inconsistent with the rest of the system.

Coming to that conclusion I wanted to create a pull request to revert the sort back to its original form. While doing that I noticed that the server side sort doesn't even matter anymore because in the meantime there was a client side sort added.

In order to have both possibilities I added that sorting by the 4th column (contains the ID among other stuff) is sorting by the ID of the location. By doing that it basically sorts by creation. Sorting by the first column (which is the default) is sorting by name.

What do you think?

### DIFF
--- a/application/models/Logbooks_model.php
+++ b/application/models/Logbooks_model.php
@@ -4,7 +4,6 @@ class Logbooks_model extends CI_Model {
 
     function show_all() {
         $this->db->where('user_id', $this->session->userdata('user_id'));
-        $this->db->order_by('logbook_name');
 		return $this->db->get('station_logbooks');
 	}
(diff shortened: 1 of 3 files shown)


magicbug:
I like this seems to be a good solution and shall merge it in :)

PR #1472, opened 2022-05-06
Sort QSL images by last added to be on top

phl0:
I think it would be nice to have the latest images at the top of the QSL images table rather than at the bottom.
Also thought using qsoid but that is not necessarily ordering by QSO date.

### DIFF
--- a/application/models/Qsl_model.php
+++ b/application/models/Qsl_model.php
@@ -9,6 +9,7 @@ class Qsl_model extends CI_Model {
         $this->db->from($this->config->item('table_name'));
         $this->db->join('qsl_images', 'qsl_images.qsoid = ' . $this->config->item('table_name') . '.col_primary_key');
         $this->db->where_in('station_id', $logbooks_locations_array);
+        $this->db->order_by("id", "desc");
 
         return $this->db->get();
     }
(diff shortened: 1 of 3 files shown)


phl0:
Enabled that. Only issue now is that the date column (at least for d.m.Y format) sorts alphabetically instead of per date. Any workaround for this?

Screenshot from 2022-05-09 10-12-17

AndreasK79:
I guess you need to consult datatable documentation about setting datatypes.

AndreasK79:
@phl0 try this commit out. Got it working with all formats..

phl0:
@AndreasK79 Works ok for me now after slight correction of the opening PHP tag ;-)

AndreasK79:
@phl0 oops, I was a bit quick there. Sorry about that. My config allows the short tag, so I didn't notice.
If you are happy with this, I'll gladly merge it in.

phl0:
np np.It works with the long tag now. As the xode uses the longer one throughout obviously I thought it would also make sense here. Go for merge :-)

phl0:
np np.It works with the long tag now. As the xode uses the longer one throughout obviously I thought it would also make sense here. Go for merge :-)

PR #1475, opened 2022-05-09
Lotw import IOTA info

phl0:
This pulls IOTA information from LotW confirmations. The IOTA info is generally contained in LotW reports if the report is requested with 'qso_qsldetail='yes''. In the code there was a line:

```
TODO: Option to specifiy whether we download location data from LoTW or not
//$lotw_url .= "&qso_qsldetail=\"yes\";
```

but a few lines before:

```
$lotw_url .= "&qso_query=1&qso_qsl='yes'&qso_qsldetail='yes'&qso_mydetail='yes'";
```

So qso_qsldetail was enabled every time independent of an option. Not sure but I liked the idea to make this option selectable to the user. So I added a LotW config option to the user profiles section which enables or disables this behaviour.
Current default setting would be disabled. But it might make sense to have that enabled by default (as it was hard coded before)?


phl0:
Will leave out the option stuff as details are downloaded anyway. Will make a new and clean PR.

PR #1478, opened 2022-05-13
Replace REGEXP_LIKE in DOK stats

phl0:
As @dg9vh reported there is an issue with REGEXP_LIKE not being supported in older mySQL versions I replaced the code with backwards compatible stuff. Fix was tested with positive result by @dg9vh.


dg9vh:
Test CFM.

PR #1486, opened 2022-05-26
Make the manual query parameter optional

Cadair:
This tweaks the code introduced in #1444 to "fail safe" if the query parameter isn't in the URL.

### DIFF
--- a/application/views/contesting/index.php
+++ b/application/views/contesting/index.php
@@ -1,3 +1,6 @@
+<?php
+$manual_time = $_GET['manual'] ?? 0;
+?>
 <div class="container qso_panel contesting">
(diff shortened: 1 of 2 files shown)


AndreasK79:
Tested ok. Thanks.

AndreasK79:
It was because of this change in footer.php:
`<?php if ( ($this->uri->segment(1) == "qso" && $_GET['manual'] ?? 0 == 0) || ($this->uri->segment(1) == "contesting" && $this->uri->segment(2) != "add")) { ?>`

This change wasn't really necessary, as it works without.

@magicbug Why couldn't you just have spent a minute fixing instead of reverting?

Issue #1485, opened 2022-05-25
Make the locator field always upper case like callsign


Cadair:
I would like to make the locator field always uppercase like the callsign field, is this something which would be a good PR?

If anyone knows where the bit of upcase magic for the callsign field is that would be handy :wink:

phl0:
Well I just tested on master and station_logbooks and it makes the chars be upper case always... ?!

magicbug:
It would be the model for adding a QSO that could have strtoupper added to the gridsquare field if its not there already

phl0:
Seems like there is some JS in place already that makes the input chars upper case on input.

magicbug:
I wonder if the issue is actually in the edit box Flo?

phl0:
Oh right. I was talking about the Live QSO input window. Shall check the others incl. the QSO edit ones.

magicbug:
I did too then started questioning my thinking :)

mpentler:
Some websites do display the second pair of letters in lowercase to disguish them somewhat, it is worth saying. Do we really want to do it? Is it convention for those last two digits to be displayed in that way?

magicbug:
I'm not aware of any fixed convention, personally, I always keep them all uppercase but that's a purely personal choice.

phl0:
> I wonder if the issue is actually in the edit box Flo?

Did a few more tests (all on station_logbooks): The QSO log windows (live and post) do convert to upper case on input instantly. Guess its some JS here. The edit QSO windows doesn't do anything like that and would also take lower case. owever the PHP code as some toUpper in here as call and grid are stored upper case only even if edited in lower case.

phl0:
Found it in the CSS code and fixed various minor issues. See commit https://github.com/magicbug/Cloudlog/commit/673dc0f6fbb5bafd97bccd4ee16304082a35c0c1. You might want to test that and report before I make it a PR.
On the log QSO section it uppercases fields callsign, locator, iota-ref, sota-ref and darc_dok. On the edit QSO section it also uppercases vucc_grids besides the aforementioned. Happy to hear from your results/tests.

Cadair:
To clarify I noticed this on the contesting input fields.

phl0:
> To clarify I noticed this on the contesting input fields.

[...]

PR #1487, opened 2022-05-29
Make various field inputs upper case on input

phl0:
This addresses issue https://github.com/magicbug/Cloudlog/issues/1485 along with some other fields that are supposed to be upper case only (e.g. DARC DOK).

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -179,7 +179,7 @@ class Logbook_model extends CI_Model {
             'COL_SOTA_REF' => trim($this->input->post('sota_ref')),
             'COL_SIG' => trim($this->input->post('sig')),
             'COL_SIG_INFO' => trim($this->input->post('sig_info')),
-            'COL_DARC_DOK' => trim($darc_dok),
+            'COL_DARC_DOK' => strtoupper(trim($darc_dok)),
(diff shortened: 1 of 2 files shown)


Werzi2001:
@phl0 I like this. Was just wondering if it would be needed/better to do it via JavaScript or server side? Doing it by CSS only affects the output and the stored values are still lower case. Might that cause problems elsewhere (e.g. ADIF export/import)?

phl0:
> @phl0 I like this. Was just wondering if it would be needed/better to do it via JavaScript or server side? Doing it by CSS only affects the output and the stored values are still lower case. Might that cause problems elsewhere (e.g. ADIF export/import)?

Let me check. The QSO input stuff has strtoupper in the PHP code anyway. So on logging the values are converted to upper case before storing to the data base. I will check and (if missing) also add strtoupper to the code that handles editing of QSOs.

phl0:
Just checked. The input for callsign and grid is converted to upper case even if you input lower case in the edit QSO window. So my fix should be sufficient and addresses only the display in the browser so it shows what is stored.

phl0:
Following the discussion in https://github.com/magicbug/Cloudlog/issues/1485 I added a few more strtouppers to make e.g. DARC DOK also uppercase on server side. Also fixed @Cadair's request for the contest logging section.

PR #1492, opened 2022-06-12
DXCC vs. Country

Werzi2001:
While playing around with `cloudLogOffline` I noticed that the country field is handled differently during ADIF import compared to QSO edit/create.

When doing an ADIF import the country is used from the ADIF data and from DXCC as a fallback. For QSO create/edit it is the other way around. That means when creating a german QSO via `cloudLogOffline` the country is set to `Germany` (which is what most german OMs have in their QRZ profile). If afterwards the QSO is edited in `Cloudlog` the country is changed to `Federal Republic Of Germany` which is the country name of the DXCC. I find that inconsistent and and unexpected change. Therefore I changed the ADIF import to give higher priority to DXCC country and use ADIF country as a fallback.

Furthermore I added `ucwords(strtolower(...))` to QSO edit also to be consistent to QSO create and ADIF import.

What do you think?

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -558,7 +558,7 @@ class Logbook_model extends CI_Model {
 
     $entity = $this->get_entity($this->input->post('dxcc_id'));
     $stationId = $this->input->post('station_profile');
-    $country = $entity['name'];
+    $country = ucwords(strtolower($entity['name']));
 
     // be sure that station belongs to user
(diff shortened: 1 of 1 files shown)


Werzi2001:
@magicbug Just noticed that this was not merged yet. Do you see a problem with this?

AndreasK79:
@Werzi2001 thumbs up from me. @magicbug ?

PR #1493, opened 2022-06-12
adapted visible conditions for lotw arrows to be consistent to eqsl

Werzi2001:
Again something that I noticed while playing around with `cloudLogOffline`

I noticed that ADIF imported QSOs do not show the (red) arros in the LotW column (column is just empty) while the arrows in the eQSL column are shown. While researching I noticed that the eQSL database columns are set to `NULL` while the LotW columns are set to `''` in the ADIF import. So first I changed that to be consistent. Afterwards I noticed that the LotW arrows had in some views (but not all) an additional condition (`$row->COL_LOTW_QSL_SENT != ''`) to be shown while the eQSL arrows did not. So I adapted that as well.

That "solved" the behaviour for me but I'm not sure if that behaviour was intentional. I searched in the code but couldn't find a hint to that. So what do you think?

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -2415,16 +2415,16 @@ class Logbook_model extends CI_Model {
         if (isset($record['lotw_qsl_rcvd'])){
             $input_lotw_qsl_rcvd = mb_strimwidth($record['lotw_qsl_rcvd'], 0, 1);
         } else {
-            $input_lotw_qsl_rcvd = "";
+            $input_lotw_qsl_rcvd = NULL;
         }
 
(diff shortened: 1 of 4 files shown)


phl0:
Think this also applies to QSOs imported/logged via API? Screenshot from WSJT-X uploading via python script.

Screenshot from 2022-06-12 14-32-55

Werzi2001:
@phl0 Yes exactly. This affects all tools/integrations that use the QSO API.

phl0:
Nice so I will have a test and report :)

phl0:
Test completed. Works as expected here.

PR #1494, opened 2022-06-12
Add map for activated grids by activators

phl0:
This adjusts the activators listing a little and adds a link to a map showing all grids activated per activator.

Screenshot from 2022-06-12 13-52-01

### DIFF
--- a/application/controllers/Activatorsmap.php
+++ b/application/controllers/Activatorsmap.php
@@ -1,0 +1,30 @@
+<?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
+/*
+
+	Data lookup functions used within Cloudlog
+
+*/
+class Activatorsmap extends CI_Controller {
+
+	function __construct() {
+		parent::__construct();
+
+		$this->load->model('user_model');
+		if(!$this->user_model->authorize(2)) { $this->session->set_flashdata('notice', 'You\'re not allowed to do that!'); redirect('dashboard'); }
+	}
+
+	public function index() {
+		$data['page_title'] = "Activators Map";
+
+		$this->load->model('stations');
+        $data['station_locator'] = $this->stations->find_gridsquare();
+
+		$this->load->view('activatorsmap/index', $data);
+	}
+
+	public function calculate() {
+		$data['grids'] = "test";
+		header('Content-Type: application/json');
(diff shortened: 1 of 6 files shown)


AndreasK79:
Just tested. Worked ok here. What about also marking the field(s)? That way, one can actually get a hint that something is marked in the map. When I now open the map, all I see is a world map with nothing marked. This of course, is that the map is zoomed. Another solution might be to center the map somewhere and zoom in when opening the map.

What do you say?

phl0:
It is there but only for higher zoom levels. Maybe also Display while zoomed to World?

AndreasK79:
I do not have any fields marked, only gridsquares, and those are only displayed when zoomed in. If they are supposed to be shown, I need to check the code and see if I can figure out why I don't have then.

AndreasK79:
What I mean is something like this:
image

phl0:
> I do not have any fields marked, only gridsquares, and those are only displayed when zoomed in. If they are supposed to be shown, I need to check the code and see if I can figure out why I don't have then.

oh sorry misunderstood. Fields are not marked. Only 4 digit grids. I think centering would be also a good option. But would require some more code to find the Center point. I might go with the fields option first...

AndreasK79:
No problem. Great work with the PR!

Another suggestion. Add this in general.css:

#mapactivators {
	height: calc(100vh - 380px) !important;
	max-height: 900px !important;
}

phl0:
> Another suggestion. Add this in general.css:
> 
> #mapactivators { height: calc(100vh - 380px) !important; max-height: 900px !important; }

Added. Still need some time for the marking of fields.

phl0:
Think I got it. @AndreasK79 could you please try latest commit and report?

Screenshot from 2022-07-02 22-40-36

AndreasK79:
@phl0 thumbs up from me. Seems to me that you nailed this one. If you are happy, I'll merge.

phl0:
> @phl0 thumbs up from me. Seems to me that you nailed this one. If you are happy, I'll merge.

Confirmed. Didn't work on the first try on my side. But may have been a cache issue. Seems fine now. OK for merge from my side.

AndreasK79:
@phl0 merge time :)

Issue #1500, opened 2022-06-19
Give summary of the number of grid squares worked


g4klx:
It would be nice if at the top of the grid square map, or somewhere else, that there could be a line summarising the number of squares claimed to have been worked, alongside the number confirmed.

phl0:
How about this?

![Screenshot from 2022-06-20 08-17-17](https://user-images.githubusercontent.com/7112907/174537329-6f719ecf-c8fc-47bc-9926-e824c14100b5.png)

![Screenshot from 2022-06-20 08-17-28](https://user-images.githubusercontent.com/7112907/174537363-d3e12117-8e3c-4c1b-862d-f6f4524d3a42.png)

https://github.com/phl0/Cloudlog/tree/addGridCountToMap

g4klx:
Spot on, exactly as I envisioned.
    On Monday, 20 June 2022, 07:20:00 BST, Florian (DF2ET) ***@***.***> wrote:  
 
 


How about this?





https://github.com/phl0/Cloudlog/tree/addGridCountToMap

magicbug:
Merged into the beta branch so it will filter across my hosted installs in the next hour or those who self host will need to git pull ;-)

phl0:
Updated my installation. Working fine here. Have fun with it @g4klx :-)
Think we can close here?

magicbug:
Think so, I'll close can always reopen at a later date

PR #1501, opened 2022-06-20
Add grid count to map

phl0:
This adds some counters to the label below the maps as requested in https://github.com/magicbug/Cloudlog/issues/1500.


magicbug:
Looks good to me @phl0 merged

Issue #1506, opened 2022-06-29
[station_logbooks] Radio timeout cannot be set


m0urs:
In Global Options if I try to set the value for "Radio Timeout" it does display correctly "Radio Timeout Warning changed to xxx seconds" but then the timeout value is "1800" again. It does not matter which number of seconds I try, it does not change.

phl0:
Confirmed. This is due to the option having been moved to store in the database in https://github.com/magicbug/Cloudlog/commit/4fb2d039130b1b32ec5e744b89f3c00d1e3d826d. But this commit did not remove the hard coded radio timeout in the application/config/cloudlog.php file where this overrides any other settings stored in the database. The fix is easy. See 149e3e169cbabe45719132a129301fb0584e6fc7.
But the update function currently removes the autoload setting for the option in the database. Need to debug and fix that before I am PR'ing this upstream.

phl0:
So I found a fix for the second issue mentioned above. The fix(es) are now in PR https://github.com/magicbug/Cloudlog/pull/1507 waiting for merge.

phl0:
Roger I will have a look. Maybe a wrong or no value is read and it is still waiting 30 minutes ...

m0urs:
It seems that https://github.com/magicbug/Cloudlog/pull/1507/commits/149e3e169cbabe45719132a129301fb0584e6fc7 breaks the CAT communication completely :-( It is not only that I do no longer get a timeout warning, I do not get any updates like frequency from the rig at all. If I undo the change, it is as before. again.

phl0:
Commit https://github.com/magicbug/Cloudlog/pull/1507/commits/570ac1e505a3455343bbae13c0021a7facd797f7 fixes this. @m0urs could you please verify?

PR #1507, opened 2022-06-29
Fix Radio timeout

phl0:
As reported in https://github.com/magicbug/Cloudlog/issues/1506 the radio timeout is not used correctly. As the config item is present in the hard coded `application/config/cloudlog.php` config the value from database is not read (though stored there correctly). This fixes it by removing the hard-coded config item.
There is also a bugfix that the autoload value is set to 'no' if updated from the web interface. So a third argument 'yes' is given to the options->update call.


phl0:
Please do not merge yet. The current state breaks the radio API completely. I am going to check that.

PR #1514, opened 2022-07-02
Fix for Logbook Model - now loads without a logbook defined. Fixed #1481

### LINKED ISSUE #1481: [station_logbooks] Logbook Model fails to load without a logbook defined
When opening the Logbook Tab without a logbook defined a non user friendly error occurs.

image
[Logbook_Model.php Line 1004](https://github.com/magicbug/Cloudlog/blob/6fe44f3b579200b37df3763c87bc53de1c294a0a/application/models/Logbook_model.php#L1004)

This then is receiving data from the list_logbook_relationships() function in [Logbook**s**_Model.php Line 177](https://github.com/magicbug/Cloudlog/blob/6fe44f3b579200b37df3763c87bc53de1c294a0a/application/models/Logbooks_model.php#L177)

This function returns a bool if there are no records found, and then errors when opening logbook with a non-obvious error.  I believe this function should either return records or an empty array - if it returns an empty array the front end should add a user friendly error/warning that there are No station logbooks defined.

_Originally posted by @T012m3n7oR in https://github.com/magicbug/Cloudlog/discussions/1480_

AndreasK79:
Before these changes, the page would not load when logbook was not defined.
@magicbug agree with these minor changes?

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -51,7 +51,7 @@ class Logbook extends CI_Controller {
 		$data['results'] = $this->logbook_model->get_qsos($config['per_page'],$this->uri->segment(3));
 
 		if(!$data['results']) { 
-			$this->session->set_flashdata('notice', $this->lang->line('error_no_logbook_found'));
+			$this->session->set_flashdata('notice', $this->lang->line('error_no_logbook_found') . ' <a href="' . site_url('logbooks') . '" title="Station Logbooks">Station Logbooks</a>');
 		}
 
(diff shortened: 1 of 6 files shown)


phl0:
Just tested here and look ok for me.

magicbug:
Looks OK, ideally however I think if logbook isn't defined it would be better to display a message to the user to go set it up

AndreasK79:
An installation wizard would be better, but how about this?
image

magicbug:
Can we add a link to the Station Logbook section to the message, after that 👍🏻 good to merge

Issue #1511, opened 2022-06-30
Unable to use hamqth callbook in latest station_logbooks branch


ofadam:
Despite having credentials set properly, I'm unable to use the hamqth callbook in the latest station_logbooks branch. QRZ.com still works as expected, though.

No error is displayed, but using the Live or Post QSO form results in nothing being looked up (not even older QSOs already logged in Cloudlog.)

AndreasK79:
The problem is related to the QRZ lookup. Edit your account and turn off this:
![image](https://user-images.githubusercontent.com/6977712/177000834-9c6bca4f-821b-4a51-a84b-0287e342a09c.png)

This need to be fixed so that it is ignored if HamQTH is chosen.

phl0:
Oopps. That is probably my fault. I will take a look.

phl0:
Found a way to fix this by implementing the picture lookup functions also for hamqth profiles. Tests and reports welcome.

PR #1516, opened 2022-07-03
Add profile pictures for HamQTH users

phl0:
This fixes https://github.com/magicbug/Cloudlog/issues/1511 by implementing the function to show images from hamqth.com profiles (also renames the option to a more general name as it is not qrz.com solely any more).


phl0:
Could one of you @magicbug or @AndreasK79 please verify my changes? The op himself seems to be not responding...

PR #1534, opened 2022-07-17
Showing Submode

dg9vh:
Showing submode if existing in map-leaflet-bubble


dg9vh:
@phl0  you could revide it and merge it :-)

phl0:
I will do a review but I do not have permissions to merge :)

phl0:
I tested and worked ok here. Sorry I reported to Kim on Matrix and not here.

PR #1545, opened 2022-07-23
Fix issue where warning is displayed that CAT data from radio is out of date

netjordan:
Hello,

I have observed an issue with the Radio/CAT Integration, where because the PHP System timezone was used the updated_minutes_ago was incorrect preventing this radio being used on the QSO screen.

Thanks,
Jordan

### DIFF
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -32,7 +32,7 @@
 			// user is not logged in
 			redirect('user/login');
 		}
-		
+
 		$this->load->model('cat');
 		$query = $this->cat->status();
 		if ($query->num_rows() > 0)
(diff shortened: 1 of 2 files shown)


phl0:
Seems like since this is merged my CAT data is counting backwards and instantly stating CAT data is out of time for 2 hours (my timezone is UTC+2).

Screenshot from 2022-08-03 17-30-30

magicbug:
When did you last update Flo? the timestamp in the database should be UTC? might be with checking the CAT table.

magicbug:
Solved, just needed to force the PHP ingest to set the timestamp and to ignore what its sent by clients

phl0:
Issue fixed with c33c8e7153f914ce269af8d0f224f046a0798912.

PR #1552, opened 2022-08-05
Add link to SatNOGS SAT info

phl0:
Add a shortcut icon with link to SatNOGS satellite data.

Screenshot from 2022-08-05 08-55-37


magicbug:
I'm concerned that the /logbook display is getting busy, this might be better inside the QSO info or even make the sat name clickable

phl0:
Roger. Will change that.

phl0:
Like that?

magicbug:
yup

PR #1553, opened 2022-08-09
Add distance to tweet

dg9vh:
if QRA-locator is set, distance is calculated and added with value to tweet

### DIFF
--- a/application/views/view_log/qso.php
+++ b/application/views/view_log/qso.php
@@ -112,19 +112,20 @@
                                 $CI->load->library('qra');
 
                                 // Cacluate Distance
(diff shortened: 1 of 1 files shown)


phl0:
Nice. I like this idea. Will test this branch.

phl0:
Tested successfully.

phl0:
Tested successfully:

Screenshot_20220809-223739_Twitter

dg9vh:
Just waiting to merge :-)

phl0:
Screenshot from 2022-08-24 17-37-01

This might need another line of code so that this only applies to QSOs where a grid is actually logged :)

phl0:
Maybe this helps:

https://hastebin.com/ifijodetet.pl

PR #1555, opened 2022-08-10
Fix rogue echos on json QSL status code

hugosilvaparagramanet:
Two forgotten `echos` were outputting code to the browser during the "get qsos and qrz data" phase of the add QSO feature, thus breaking the json output.


phl0:
Oh, it could have been me who intruced this bug :)

phl0:
Ah no it was @dg9vh ;-)

https://github.com/magicbug/Cloudlog/commit/6813237c591c1f50b7b92bd73e81add5d4afceef

Issue #1539, opened 2022-07-18
[Station Locations] Make City field optional


magicbug:
### Discussed in https://github.com/magicbug/Cloudlog/discussions/1538

<div type='discussions-op-text'>

<sup>Originally posted by **KB2YSI** July 18, 2022</sup>
Is there a reason that the CITY field in a station location is required?

I frequently operate from many locations, satellite rover, and hardly know what county I am in much less than the town/city.  Having to enter that for each location is going to be either completely incorrect or a PITA to obtain the correct name.</div>

phl0:
PR https://github.com/magicbug/Cloudlog/pull/1572 is merged. Please test and report.

PR #1572, opened 2022-08-17
Make specification of city name optional

phl0:
This addresses https://github.com/magicbug/Cloudlog/issues/1539 and removes the need to specify a city name upon creation or editing a station location.

### DIFF
--- a/application/views/station_profile/create.php
+++ b/application/views/station_profile/create.php
@@ -56,7 +56,7 @@
 
 		  <div class="form-group">
 		    <label for="stationCityInput">Station City</label>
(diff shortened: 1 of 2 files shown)


magicbug:
The only thing I need to check is the LoTW sync stuff I cant remember if operators city was a required!

phl0:
> The only thing I need to check is the LoTW sync stuff I cant remember if operators city was a required!

Just pushed this change to my prod environment, deleted my city and uploaded to LotW. No issues seen. QSOs are displayed on on LotW and confirmed even matching. So no issue here I suppose.

PR #1576, opened 2022-08-19
Fix eQSL upload for linked account

nolith:
When uploading QSOs for a linked account the username must match the
station callsign. i.e. when operating portable it should be callsign/p.

While working on this I realized that if a station location has a QTH nickname but no QSO assigned, it will show up as an empty line in the log to be updated.

Screen Shot 2022-08-24 at 17 33 59
Here I have 2 locations with the same QTH nickname but no QSO.

Changing the join type from `left` to `right` fixed that issue.

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -149,7 +149,7 @@ class eqsl extends CI_Controller {
 			$active_station_id = $this->stations->find_active();
         	$station_profile = $this->stations->profile($active_station_id);
 			$active_station_info = $station_profile->row();
-			// Query the logbook to determine when the last LoTW confirmation was
+			// Query the logbook to determine when the last eQSL confirmation was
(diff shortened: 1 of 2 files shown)


AndreasK79:
@magicbug Changes looks fine for me, but since I don't use eQSL, I can't really test.

magicbug:
@AndreasK79 I'm in a similar position hence a bit scared to merge it! 

@phl0 you using eQSL these days?

magicbug:
> @magicbug Changes looks fine for me, but since I don't use eQSL, I can't really test.

Option #2 is YOLO and merge and wait for someone to complain..

phl0:
> @phl0 you using eQSL these days?

Well in the sense that I don't create separate station locations for each /p activity due to its contraints regarding overlapping times. I just upload my QSOs to eQSL. So my vote is for option #2 :)

magicbug:
Lets go with YOLO then :D

phl0:
Had to look that one up :-P Ok with that!

PR #1582, opened 2022-08-27
[WIP] Unify radio API

kb-light:
This is a first draft to unify the radio API, so it does not depend on the names.

Currently work in progress and needs some further testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 .DS_Store
 sync.sh
 *.p12
+*.swp
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 100;
+$config['migration_version'] = 101;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 99;
+$config['migration_version'] = 100;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -48,17 +48,23 @@
 			{
 				echo "<tr>";
 				echo "<td>".$row->radio."</td>";
-				if($row->frequency != "0" && $row->frequency != NULL) {
+
+				if (empty($row->frequency) || $row->frequency == "0") {
+					echo "<td>- / -</td>";
+				} elseif (empty($row->downlink_freq) || $row->downlink_freq == "0") {
 					echo "<td>".$this->frequency->hz_to_mhz($row->frequency)."</td>";
 				} else {
-					echo "<td>".$this->frequency->hz_to_mhz($row->downlink_freq)." / ".$this->frequency->hz_to_mhz($row->uplink_freq)."</td>";
+					echo "<td>".$this->frequency->hz_to_mhz($row->downlink_freq)." / ".$this->frequency->hz_to_mhz($row->frequency)."</td>";
 				}
 
-				if($row->mode != "non" && $row->mode != NULL) {
+				if (empty($row->mode) || $row->mode == "non") {
+					echo "<td>N/A</td>";
+				} elseif (empty($row->downlink_mode) || $row->downlink_mode == "non") {
 					echo "<td>".$row->mode."</td>";
 				} else {
-					echo "<td>".$row->downlink_mode." / ".$row->uplink_mode."</td>";
+					echo "<td>".$row->downlink_mode." / ".$row->mode."</td>";
 				}
+
 				$phpdate = strtotime($row->timestamp);
 				echo "<td>".date('H:i:s d-m-y', $phpdate)."</td>" ;
 				echo "<td><a href=\"".site_url('radio/delete')."/".$row->id."\" class=\"btn btn-danger\"> <i class=\"fas fa-trash-alt\"></i> Delete</a></td>" ;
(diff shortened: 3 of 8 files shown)


magicbug:
Fantastic pleased someones taken on this, so the major part will be the javascript on the frontend however should be straightforward

phl0:
Sound a good idea. Always crashed my head when two CloudlogCatQt instances overwrote each others data. Thought about adding a user-editable text field to make those distinguishable. Are you planning to add such?

magicbug:
Better option might be the app sending a unique name

phl0:
> Better option might be the app using the API sending a unique name

Well that was what I am thinking about. Like let the user enter an ID. E.g. "IC-7300" or the like which you can later find in the Radio section while logging the QSO.

kb-light:
Changing the name should be a feature of the app (e.g. CloudlogCatQt) and not within Cloudlog.

@magicbug did you have anything special in mind concerning the javascript frontend? As for now only the dependency on the radio/app name is removed, so the user should not notice any difference.

Some thoughts for further changes:
- Dropping `uplink_freq` and switch to `frequency` and `frequency_rx` to be consistent with the frontend
- Allowing both frequencies being set, even when not in satellite mode, e.g. one can log rx and tx frequencies when working via a relay

phl0:
> Changing the name should be a feature of the app (e.g. CloudlogCatQt) and not within Cloudlog.

That what I meant. I thought about an addition to the app name. But user-input text only is also ok for me. The name of the app itself is not important here :)

>     * Dropping `uplink_freq` and switch to `frequency` and `frequency_rx` to be consistent with the frontend

ok with that.

>     * Allowing both frequencies being set, even when not in satellite mode, e.g. one can log rx and tx frequencies when working via a relay

This is currently working already I think.

magicbug:
> Changing the name should be a feature of the app (e.g. CloudlogCatQt) and not within Cloudlog.
> 
> @magicbug did you have anything special in mind concerning the javascript frontend? As for now only the dependency on the radio/app name is removed, so the user should not notice any difference. EDIT: Found SatPC32 in`application/views/dashboard/index.php` and `application/views/interface_assets/footer.php`.
> 
> Some thoughts about further changes:
> 
> * Dropping `uplink_freq` and switch to `frequency` and `frequency_rx` to be consistent with the frontend
> * Allowing both frequencies being set, even when not in satellite mode, e.g. one can log rx and tx frequencies when working via a relay

Yes this is fine, SatPC32 references can just be changed to act upon propagation mode being SAT

kb-light:
Here the current structure of the radio API. At the moment the timestamp value is not used, I am not sure whether that is intended or just by accident.
```json
{
    "key": "cl0123456789abc", # API-Key
    "frequency": "2400170000", # Frequency in Hz
    "frequency_rx": "10489670000", # RX Frequency in Hz (optional)
    "mode": "SSB", # Mode
    "mode_rx": "SSB", # RX mode (optional, not logged)
    "prop_mode": "SAT", # Propergation mode (optional)
    "sat_name": "QO-100", # Satellite name (optional)
    "power": "5", # Transmit power in Watts (optional)
    "timestamp": "2022-08-28 12:00:00", # Timestamp of last update (optional, not logged)
}
```
Migration
- `uplink_freq` -> `frequency`
- `uplink_mode` -> `mode`
- `downlink_freq` -> `frequency_rx`
- `downlink_mode`-> `mode_rx`

ToDo
- Update DB accordingly

magicbug:
Is this ready for review now, I'm looking at wrapping station_logbook branch into v2.0 so now is the time to say.

magicbug:
@kb-light I think timestamp in the database table was always designed to be set by Cloudlog rather than third party app, more because of half don't send the time in UTC correctly.

kb-light:
I need to update the cat table for consistent naming scheme, after that this should be ready to merge. The current version is already working and can be used for testing.

phl0:
I will change my PR against CloudLogCatQt accordingly and test.

phl0:
I added a small PR against @kb-light's repo because for backwards compatibility with CloudLogCatQt for example we need to check for mode and freuqency being != "NULL". See https://github.com/kb-light/Cloudlog/pull/1.

phl0:
So I refactored my patch against @myzinsky'S CloudLogCatQt to add the manually editable identifier to the radio name. The branch is https://github.com/phl0/CloudLogCatQt/tree/refactorCat. Works with @kb-light's code and my additions from https://github.com/kb-light/Cloudlog/pull/1:

Screenshot from 2022-09-29 13-24-00

Screenshot from 2022-09-29 13-24-11

phl0:
SAT mode has an issue of not being set due to a forgotten change in variable name in `application/controllers/Radio.php` line 121. Fixed with https://github.com/kb-light/Cloudlog/pull/1/commits/7280e6b659935fa974c6385476c61b15a2105b4d. Now SAT mode field is filled correctly again.

phl0:
Found a new issue for the case that CloudLogCatQt is used for non-SAT mode where frequency_rx and mode_rx are missing. Results in a PHP error. It works anyway but I think we should cover that case to prevent filling of logs with error messages.

Screenshot from 2022-09-29 16-48-12

phl0:
Submitted a fix for the non-SAT cases under https://github.com/kb-light/Cloudlog/pull/1/commits/78771a43d3862490367fe8a6b6be612fc14e8281.

kb-light:
@phl0 should frequency_rx also checked against "NULL" or is it just omitted when not used?

kb-light:
First draft for `application/migrations/100_update_cat_table.php`, not tested yet. Not sure which checks are needed for up and also downgrade.

phl0:
Found some issues with your latest changes in `application/migrations/100_update_cat_table.php`. Committed fixes with https://github.com/kb-light/Cloudlog/pull/4 to your radioapi branch. There were some syntax errors in the migration script. And I added some checks so that the script does not fail in cases where columns do (not) exist.

phl0:
> @phl0 should frequency_rx also checked against "NULL" or is it just omitted when not used?

I think it is just ommitted in that case.

phl0:
Some thing I just noticed: `frequency` is a VARCHAR(10) whereas `frequency_rx` is (and `downlink_freq` was) BIGINT. Should we migrate `frequency` to BIGINT as well?

kb-light:
Should already be BIGINT, see `install/assets/install.sql`.

phl0:
> Should already be BIGINT, see `install/assets/install.sql`.

And made to VARCHAR in `090_make_mode_frequency_null.php` .... But why?!

phl0:
I think I have made a bad mistake in https://github.com/magicbug/Cloudlog/pull/1453/commits/f6feea53bd766f6cf39fbd481795f8dc1cd4a63d.......... 

I would got ahead and fix it with another PR against radioapi. Are you ok with that?

phl0:
After talking to @magicbug I added https://github.com/kb-light/Cloudlog/pull/4/commits/d63fdc757bffecf337dc63f22c56ec63e5cf1fa0 to the PR against the radioapi branch by @kb-light. This restores BIGINT as data type for `frequency` column in `cat` table.

PR #1584, opened 2022-08-30
[Install] add missing user_id to table cat

kb-light:
Seems the user_id column in cat table is missing. Just checked by code, have not done a new installation. For migrations this is done by [`application/migrations/078_add_userid_to_hardware.php`](https://github.com/magicbug/Cloudlog/blob/station_logbooks/application/migrations/078_add_userid_to_hardware.php).


kb-light:
Maybe I am wrong and this is archived by `INSERT INTO `migrations` VALUES ('15');` within `install/assets/install.sql`

phl0:
Jepp I think this is not needed. Upon installation version is set to 15 while application/config/migration.php says `$config['migration_version'] = 97;`. So upon loading the page for the first time all migrations between 16 and 97 in this case are executed. With `079_add_userid_to_api.php` the corresponding column is added to the `cat` table.
Just confirmed with a fresh install.

Issue #1586, opened 2022-09-01
[API][QSO] QSO not included in ADIF export when LoTW state is NULL


kb-light:
**Describe the bug**
When using the Cloudlog API to upload QSOs from WSJT-X the LoTW received state is Null. These QSOs are not included in ADIF export, when using "Export QSOs not uploadad to LoTW".

Branch: station_logbooks

**To Reproduce**
Steps to reproduce the behaviour:
1. Setup WSJT-X improved PLUS and link to Cloudlog
2. Make a QSO
3. Verify that it got added to Cloudlog (LoTW state shows not uploaded)
4. Go to ADIF Import / Export -> ADIF Export
5. Select your Callsign and "Export QSOs not uploadad to LoTW"
6. The QSO is not included in the exported ADIF

You can also use Advanced Search and compare the results of following queries:
- LOTW_QSL_SENT equal Y
- LOTW_QSL_SENT equal N
- LOTW_QSL_SENT is null
- LOTW_QSL_SENT not equal Y (not including LOTW_QSL_SENT is null)

**Expected behaviour**
All QSOs that are not flagged as LoTW QSL sent are included within the ADIF export. As an alternative the Default for LoTW upload state could be `'N'` and not `NULL`.

**Desktop:**
 - OS: Windows 10
 - Browser: Firefox
 - Cloudlog: station_logbooks
 - WSJT-X: WSJT-X improved PLUS v2.6.0~rc3 (@phl0)

phl0:
Could be related to https://github.com/magicbug/Cloudlog/pull/1523. I will check.

phl0:
Ok, just confirmed that `COL_LOTW_QSL_SENT? in NULL upon uploading a QSO via API (from WSJT-X improved):

```mysql> select COL_QSO_DATE, COL_TIME_ON, COL_CALL, COL_LOTW_QSL_SENT from TABLE_HRD_CONTACTS_V01 where 1 order by COL_PRIMARY_KEY  desc limit 1;
+--------------+---------------------+----------+-------------------+
| COL_QSO_DATE | COL_TIME_ON         | COL_CALL | COL_LOTW_QSL_SENT |
+--------------+---------------------+----------+-------------------+
| 2022-09-01   | 2022-09-01 12:27:00 | OE3DXA   | NULL              |
+--------------+---------------------+----------+-------------------+
1 row in set (0.00 sec)
```

kb-light:
[...]

PR #1588, opened 2022-09-02
Also export QSOs with LotW QSL SENT date of NULL

phl0:
This fixes issue https://github.com/magicbug/Cloudlog/issues/1586 by also exporting QSOs with QSL SENT date being NULL (besides "N"). The rest of my idea concerning setting the SQL columns to be `NOT NULL DEFAULT 'N'` to be discussed seaparately.

### DIFF
--- a/application/models/Adif_data.php
+++ b/application/models/Adif_data.php
@@ -93,7 +93,7 @@ class adif_data extends CI_Model {
             $this->db->where("date(".$this->config->item('table_name').".COL_TIME_ON) <= '".$to."'");
         }
         if ($exportLotw) {
-            $this->db-group_start();
+            $this->db->group_start();
             $this->db->where($this->config->item('table_name').".COL_LOTW_QSL_SENT != 'Y'");
             $this->db->or_where($this->config->item('table_name').".COL_LOTW_QSL_SENT", NULL);
             $this->db->group_end();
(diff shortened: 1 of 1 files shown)


application/models/Adif_data.php
```diff
@@ -93,7 +93,10 @@ function export_custom($from, $to, $station_id, $exportLotw = false) {
             $this->db->where("date(".$this->config->item('table_name').".COL_TIME_ON) <= '".$to."'");
         }
         if ($exportLotw) {
+            $this->db-group_start();
```
kb-light, line 96: Typo, `>` is missing, should be: `$this->db->group_start();`

application/models/Adif_data.php
```diff
@@ -111,7 +114,10 @@ function export_lotw() {
         $this->db->select(''.$this->config->item('table_name').'.*, station_profile.*');
         $this->db->from($this->config->item('table_name'));
         $this->db->where($this->config->item('table_name').'.station_id', $active_station_id);
+        $this->db-group_start();
```
kb-light, line 117: Same here

application/models/Adif_data.php
phl0: Foxed

application/models/Adif_data.php
phl0: Fixed

phl0:
Fixed the darn C+P mistakes with https://github.com/magicbug/Cloudlog/pull/1588/commits/5f922fe5b9b4a6b02aae5ca5b7113e569afcc95c. Tnx for spotting.

PR #1594, opened 2022-09-03
[Accumulated DXCC] Rewrote query for accumulated dxcc. Major speedup for monthly stats

AndreasK79:
Only the dxcc part has been touched so far. I need to know that this works for other people as well. If it does, I'll rewrote the others as well.

On my machine, I managed to decrease the time for monthly accumulated dxcc from over 30s, down to around 1s.

Anyone willing to test and give some feedback?

### DIFF
--- a/application/models/Accumulate_model.php
+++ b/application/models/Accumulate_model.php
@@ -25,7 +25,6 @@ class Accumulate_model extends CI_Model
     }
 
     function get_accumulated_dxcc($band, $mode, $period, $location_list) {
-
         if ($period == "year") {
             $sql = "select year(thcv.col_time_on) year";
         }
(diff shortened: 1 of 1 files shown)


phl0:
I tried to call the page using the old code. But my SQL server was running hot and I didn't even have a reply by 120 seconds. Using your code is now faster than a 1 second here. Thumbs up!

AndreasK79:
@phl0 excellent. That's what I was hoping to hear. 

I always knew this was slow, but I wasn't sure how to fix it without any special windowing functions, or write a function in php to get the result. After playing around with sql today, I managed to find a replacement that seems to work a lot quicker!

AndreasK79:
@phl0 Updated the rest of the queries now. Mind having another go?

phl0:
QSL. Working fine on this end. Go for merge :)

AndreasK79:
@phl0 thanks for testing.

phl0:
uw. Rolled out to my prod system. All working fine!

AndreasK79:
Much smoother now. Just like it should be!

PR #1595, opened 2022-09-05
User Selectable bands

AndreasK79:
This is what I have done so far.

1. Migration script for tables and values.
2. Menu item for editing bands (admin -> bands).
3. Settings can be saved.
4. Awards that use bands->get_worked_bands, use the active bands.
5. Some analytics use the active bands.
6. Inserts entries in bandxuser when a new user is created.
7. Message when band is saved.
8. Can now create new band.
9. Can now edit band.
10. Band dropdown in gridsquare map now uses active bands
11. Added bandgroup to bands table. To be used in band dropdown in qso entry/contest qso entry.
12. QSO entry now fetches band/bandrx from DB.
13. Contest logging dropdown now fetches band from DB.
14.  Frequency library updated to use db.
15. QSO edit also uses bands from db (shows disabled as well for editing purposes).

I say this is done. I think I've updated all places necessary.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 97;
+$config['migration_version'] = 98;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/Awards.php
+++ b/application/controllers/Awards.php
@@ -90,7 +90,7 @@ class Awards extends CI_Controller {
         $this->load->model('modes');
         $this->load->model('bands');
 
-        $data['worked_bands'] = $this->bands->get_worked_bands(); // Used in the view for band select
+        $data['worked_bands'] = $this->bands->get_worked_bands('dxcc'); // Used in the view for band select
         $data['modes'] = $this->modes->active(); // Used in the view for mode select
 
         if ($this->input->post('band') != NULL) {   // Band is not set when page first loads.
@@ -156,7 +156,7 @@ class Awards extends CI_Controller {
     public function vucc()	{
         $this->load->model('vucc');
         $this->load->model('bands');
-        $data['worked_bands'] = $this->bands->get_worked_bands();
+        $data['worked_bands'] = $this->bands->get_worked_bands('vucc');
 
         $data['vucc_array'] = $this->vucc->get_vucc_array($data);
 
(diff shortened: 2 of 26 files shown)


phl0:
Looking fine so far imho. Only issue so far:
If I am about to create a band I get his:

Screenshot from 2022-09-06 13-21-12

Seems like there is a mixture of creating a band vs. creating a mode? Is that maybe WIP?

AndreasK79:
Thanks for testing. Yes, WIP. Check my comment above over what is missing :)

phl0:
Should have re-read that. So all fine here :)

AndreasK79:
The whole thing is a WIP. I'm adding and fixing things when I find some time.

phl0:
All changes looking fine here. Just found a bug in the US Counties section. But not related to these changes. Will PR a fix separately.

Screenshot from 2022-09-07 13-04-19

phl0:
Fixed the US counties error with https://github.com/magicbug/Cloudlog/pull/1596.

AndreasK79:
@phl0 thanks for testing and fixing the oversight in other parts of the code :)

phl0:
Tested again. Still looking good and nothing b0rken :)

AndreasK79:
Thanks. I try to test when I make changes, and sometimes I miss something, so it's always good to have others take a look and test.

Once I figure out the frequency library, I think most of it is done.

phl0:
There is an issue with the migration:

Screenshot from 2022-09-10 13-13-42

'' is not valid for frequencies. I guess we should set them to 0 instead. See

AndreasK79:
Thanks. My installation didn't mind blanks there, it just put 0's in db. That's why it's very nice to get other to test out the code before merge.

AndreasK79:
@magicbug any feedback here? Any changes that needs to be done?

magicbug:
I like it, I do wonder if it needs to be on the Username dropdown rather than the admin, thinking as its possible to have a admin user and a standard user now. (Need to document that!)

I know its extra coding but would be sweet if we didn't have to click the save button, at first I didn't spot said button but understand its extra coding. as is think its ok to merge

AndreasK79:
@magicbug thanks for the feedback. Can we merge this now, and then I'll look into those changes?

magicbug:
Merge as is, the rest can be added :)

AndreasK79:
Great. I was thinking about another way to save, as I found that changing many bands, then saving each band was not a good way to do it. I had in mind a save button below, but I think a listener for changes might be the best. Need to investigate :)

magicbug:
If we could do it just by saving when user checks and uncheck would be fab! we can discuss this in a separate issue if you want.

PR #1599, opened 2022-09-10
Add WWFF Award infos

phl0:
Add a simple WWFF awards table base on recently added WWFF fields.


AndreasK79:
Tested working as it should.

PR #1603, opened 2022-09-12
Make <eor> upper case

phl0:
As all other ADIF tages are exported upper case it would make sense to also have <EOR> instead of <eor>.


magicbug:
@phl0 @AndreasK79 Just to say this was a breaking change to LoTW and has been reverted. Can be resubmitted once further testing carried out

application/views/lotw_views/adif_views/adif_export.php
```diff
@@ -10,7 +10,7 @@
 <CERT_UID:1>1
 <CERTIFICATE:<?php echo strlen(trim($cert2)) + 1; ?>><?php echo trim($cert2); ?>
 
-<eor>
+<EOR
```
phl0, line 13: This is causing the LotW issues.

phl0:
Found the usse. Was an awkward mistake by me having lost a closing tag on one EOR in the LotW section. See https://github.com/magicbug/Cloudlog/pull/1603/files#r973873612. I fixed it and made a new PR https://github.com/magicbug/Cloudlog/pull/1607. This has been tested and uploads to LotW fine. Other tests welcome.

phl0:
So this is for the people that need to fix their logs and have would to do so the SQL way. Find the latest QSO uploaded successfully from the LotW online view. Then lookup the ID of the QSO inside Cloudlog (In my case it was F6BIR/P):

Screenshot from 2022-09-19 07-56-06

With this you can update your cloudlog table using this SQL command:

`UPDATE `TABLE_HRD_CONTACTS_V01` SET COL_LOTW_QSLSDATE = NULL, COL_LOTW_QSL_SENT = 'N', COL_LOTW_STATUS = NULL WHERE COL_PRIMARY_KEY > 34647 `

AndreasK79:
> @phl0 @AndreasK79 Just to say this was a breaking change to LoTW and has been reverted. Can be resubmitted once further testing carried out

Sorry about that. I didn't spot that problem. Testing is always the key!

PR #1605, opened 2022-09-18
Statistics reworked

AndreasK79:
Changed from using Google Chart to Chart.js
Upgraded Chart.js to latest version.


phl0:
I like the new view. But it seems some graphs are cut off in my case:

Screenshot from 2022-09-18 17-32-08

Screenshot from 2022-09-18 17-32-16

AndreasK79:
Thanks for testing. My log didn't contain enough data to show that problem. I'll take a look.

AndreasK79:
@phl0 refresh and give it another go. The outlabel plugin I use, is kinda tricky to get right.
Might split into several tabs, and not have chart and table side by side.

phl0:
Well ... better :)

Screenshot from 2022-09-18 18-03-42

Screenshot from 2022-09-18 18-03-35

AndreasK79:
You need to work more on bands and satellites with few QSOs :)

If the plugin could put the labels on the other side, instead of just stacking them on top, it would not need as much space. Guess I need to research the code more.

phl0:
Na no worries. My log is a little SAT-centric. I agree :)

AndreasK79:
@phl0 This is very tricky to fix, since those with few QSOs will be stacked above each other.

Just for the fun of it. Could you edit statistics.js and add rotation: -45, after line 315 and 441 right after "options: {"?

phl0:
Good on the general one. But worse on the SAT page :)

Screenshot from 2022-09-18 20-50-45

Screenshot from 2022-09-18 20-50-36

AndreasK79:
Thanks. I've been looking into the code for the label-plugin, but I'm not sure how to fix this. Perhaps we should hide the labels for those below a given percentage...

phl0:
Just wanted to suggest that. Maybe just ignore everything below 5%?

magicbug:
Or just show the labels on roll over of the area of the pie chart.

AndreasK79:
@magicbug they do that now, but I like this design better. I'll see if I can find some way to not display the ones with low percentage. Worst case, I disable the label plugin.

AndreasK79:
@phl0 could you test this one? I set it to 1% just to see how it works out now.

phl0:
Looking fine for me:

Screenshot from 2022-09-19 09-37-42

Screenshot from 2022-09-19 09-37-49

AndreasK79:
@phl0 thanks. I think that must be the solution then.

@magicbug could you test it out and see if it's ready for merge?

phl0:
Looking good from my side :)

magicbug:
Looks ok my end

PR #1607, opened 2022-09-19
Make <eor> tags upper case (now with fixed LotW issue)

phl0:
New patch with fixed issue..


AndreasK79:
@magicbug As the error was identified (and fixed here), merge?

phl0:
Tnx merge. Just tested again with positive result. So this fix really works :)

PR #1608, opened 2022-09-19
Add "-(/" as separators to ucwords to display DXCCs (more) correctly

phl0:
Lgging my QSO with E7/4O4A I discovered that the country name is shown as "Bosnia-herzegovina". I guess "Bosnia-Herzegovina" would be a little nicer. So added "- (/" as separators to ucwords functions. I guess \t\r\n\f\v are not needed (from the default setting). The result is:

Screenshot from 2022-09-19 08-57-10

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -49,7 +49,7 @@ class Logbook_model extends CI_Model {
 
     if($this->input->post('country') == "") {
       $dxcc = $this->check_dxcc_table(strtoupper(trim($this->input->post('callsign'))), $datetime);
-      $country = ucwords(strtolower($dxcc[1]));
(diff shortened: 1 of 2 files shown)


phl0:
@AndreasK79 Could you maybe test this one for me?

AndreasK79:
@phl0 I'll test it out later today.

phl0:
According to the PHP docu the default separators are `" \t\r\n\f\v"`. But I think we only need the space from that. Tabs and newlines should not appear in the DXCC list. I have a look at the DXCC table and found - / and ( where it would also make sense to have an upper case character after those.

AndreasK79:
Worked as it should in the tests, so no reason not to merge :)

PR #1609, opened 2022-09-20
[Bands] Now saves when clicking checkbox

AndreasK79:


AndreasK79:
@magicbug this is what we discussed. It will create several call for saving to the database, but I guess you aren't editing the bands that often. I guess it's more of a set and forget.

phl0:
Tested ok on my end. Bands configuration now in the user's menu.

AndreasK79:
@phl0 thanks for testing. I also need to add WWFF to the configuration, thanks to you :)

magicbug:
Lovely! much more user friendly i think

AndreasK79:
Then it's merge time :)

PR #1611, opened 2022-09-20
Add user-editable identifier to CAT API interface

phl0:
This adds a separate column to the CAT interfaces table so that multiple instances of a CAT interface can be distinguished. CloudLogCatQt has a running PR under https://github.com/myzinsky/CloudLogCatQt/pull/8 which adds a user editable field to allow the user to enter an identifier. This is shown in the hardware interfaces as well as the radio selection during logging of QSOs. This is currently work in progress. Reports by testers would be fine :)

Screenshot from 2022-09-20 22-59-06

Screenshot from 2022-09-20 22-59-28

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 99;
+$config['migration_version'] = 100;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -39,6 +39,7 @@
 		{
 			echo "<tr class=\"titles\">";
 				echo "<td>Radio</td>";
+				echo "<td>Identifier</td>";
 				echo "<td>Frequency</td>";
 				echo "<td>Mode</td>";
 				echo "<td>Timestamp</td>" ;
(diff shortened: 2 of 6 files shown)


phl0:
Ready for merge from my point of view. No (negative) test results received though.

kb-light:
No need for an extra field after #1582 is merged, you can then customize the radio name.

phl0:
Hmm. What about two instances of SATPC32? I think they would still overwrite each other as the application name is hard coded in there?

kb-light:
That is right, but there has to be a change in the client software anyways or they would also overwrite each other after your change. So I don't see the benefit of the extra field instead of just changing the radio name.

Even when there should be the possibility to set radio name and also an additional identifier via the API, this should just be merged into the radio name on cloudlog site. (No additional column needed, easier to maintain)

phl0:
Okay then. For [CloudLogCatQt](https://github.com/myzinsky/CloudLogCatQt) this is no problem. I can change my PR to that repo accordingly. But how about SatPC32? @magicbug u know?

phl0:
Ok, then I will cancel my two PRs and we will use the `radio` field in the `cat` table for this. So applications should allow the user to configure the value `radio` of the API.

phl0:
Closing here in favor of https://github.com/magicbug/Cloudlog/pull/1582.

Issue #1618, opened 2022-09-24
US county is not re-enabled after field reset


phl0:
**Describe the bug**
The US county field is disabled upon entry of a non-US callsign. But it remains disabled after resetting the fields. So Logging a second QSO after that entry of US county or prefill of US county is not possible. Reloading the log entry page re-enables the field.

**To Reproduce**
Steps to reproduce the behaviour:
1. Enter a non-US callsign into the callsign field
2. Check the "General" tab
3. US County field is disabled (i.e. greyed out)
4. Press ESC or "Reset" button
5. US County field is not re-enabled

**Expected behaviour**
US County field should be re-enabled during field reset

**Screenshots**

![Screenshot from 2022-09-24 21-20-26](https://user-images.githubusercontent.com/7112907/192114997-a13624d5-66cd-4264-be82-d1403949db81.png)

phl0:
A patch for this behaviour can be found in https://github.com/magicbug/Cloudlog/pull/1619.

AndreasK79:
It is disabled because no state is filled. It does only check against that, not dxcc. And if you enable it, then press reset, it is still enabled. I remember that selectize.destroy would create trouble with some other js. I don't know what it was, but it was commented out in qso_edit in footer.php and also around line 2609.

AndreasK79:
It is also commented out in qso.js:58.

AndreasK79:
The only problem I see here, is that it is enabled after reset, so that it should not be enabled after reset. The reason why I did this, is that the state is needed to search correctly after county. There are so many counties, and many of them have the same name, just different state, so state if used to filter.

phl0:
I think it also has to do with https://github.com/magicbug/Cloudlog/issues/1615#issue-1382892278. The lookup function gives `callsign_usa_county` back. And there are two cases which obviously lead to different tesults. It can be an empty string "" which I think works correct. But there is also the case where it gives NULL and this leads to different behaviour (where it should not imho).

Need to look deeper into this why and where this differs. So my current patch my be inappropriate (though I think separate variable names to the selections would not harm).

PR #1619, opened 2022-09-24
[WIP] Reset County selectized input correctly

phl0:
This addresses Issue https://github.com/magicbug/Cloudlog/issues/1618. It seems that the selectized input also needs a re-enable and a destroy to be reset properly. Besides that I also added independend variable names for the selectzizes to be reset (as done in https://github.com/magicbug/Cloudlog/pull/1616). 

With this patch the US county field can be filled after a previous QSO with a non-US station was reset.


phl0:
Not ready for merge yet as per https://github.com/magicbug/Cloudlog/issues/1618. Need to dig into this again.

Issue #1549, opened 2022-07-27
Analytics could usefully use a page of statistics


g4klx:
It would be great to have a page in the analytics section that gave the raw statistics of activity across all of the bands. For example the list of bands in the left most column, followed by columns for the number of countries, grid squares, zones, etc. that have been confirmed (and maybe unconfirmed also). Only showing those bands where the numbers are greater than zero.

That way it would be easy to see progress towards awards, and general status of the confirmations. It wouldn't need any special graphics so would presumably be pretty efficient too.

kb-light:
I would be interested in number of unique stations (maybe also by band/mode).

kb-light:
I am fine with the new statistics from #1621

PR #1621, opened 2022-09-25
[Statistics] Raw QSO data and unique callsigns

AndreasK79:
Also remade the page a little bit:
image
Split them into sections, which only loads when clicked on.

Added 2 tables now. One with raw QSO data, and one with unique callsigns worked.

### DIFF
--- a/application/controllers/Statistics.php
+++ b/application/controllers/Statistics.php
@@ -146,4 +146,31 @@ class Statistics extends CI_Controller {
 		header('Content-Type: application/json');
 		echo json_encode($satstats);
 	}
+
+	public function get_unique_callsigns() {
+		$this->load->model('stats');
+
+		$result = $this->stats->unique_callsigns();
+		$total_qsos['qsoarray'] = $result['qsoView'];
+		$total_qsos['bandunique'] = $result['bandunique'];
+		$total_qsos['modeunique'] = $result['modeunique'];
+		$total_qsos['total'] = $result['total'];
+		$total_qsos['bands'] = $this->stats->get_bands();
+
+		$this->load->view('statistics/uniquetable', $total_qsos);
+	}
+
+	public function get_total_qsos() {
+		$this->load->model('stats');
+
+		$totalqsos = array();
+
+		$result = $this->stats->total_qsos();
+		$total_qsos['qsoarray'] = $result['qsoView'];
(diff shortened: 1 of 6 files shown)


phl0:
Tested with good result. And looking fine.
Might solve https://github.com/magicbug/Cloudlog/issues/1549.

AndreasK79:
It will solve partially that case. I have only included raw QSO numbers. I have not added any award stuff. Unique callsigns are shown by band/mode/total, so that should be fine.

Thanks for testing @phl0 .

kb-light:
image
17m has different values for total number of QSOs. I think I have identified the issue. It seems that QSOs where SUBMODE is null, only count for total values, but not for the more specific ones.

dg9vh:
Hi,

it seems to me, that this new development is not bug-free in any case as you can see in the screenshot:

image

I don't know, why there are more columns than needed, but maybe @AndreasK79 could take a look at it? If you need some stuff from me (like adif-exports or similar) let me know.

73 de Kim
DG9VH

dg9vh:
Hi, I solved it by converting my bands to lower-case... It seems that old ADIF-imports or something left some upper-case bands like 40M or something in it that confuses the SQL.

AndreasK79:
> image 17m has different values for total number of QSOs. I think I have identified the issue. It seems that QSOs where SUBMODE is null, only count for total values, but not for the more specific ones.

Ok, it might be that the code only checks on null, not blank. It might work if coalesce(col_submode, '') is done in the queries, and check if value is blank. I'll see if I can figure something out. Thanks for reporting.

AndreasK79:
> Hi, I solved it by converting my bands to lower-case... It seems that old ADIF-imports or something left some upper-case bands like 40M or something in it that confuses the SQL.

Thanks for reporting. I thought it was col_submode that was blank, but guess different casing on bands are distinct in the queries. Lowercase or uppercase all bands might work. Guess I need to revisit the code :)

AndreasK79:
@kb-light I've pushed a fix for this issue. Could you pull the latest code and give it a go?

AndreasK79:
@dg9vh I pushed a fix that should work for the issue you reported. If you updated your band in the log, it shouldn't make any difference for you now.

kb-light:
@AndreasK79 looks good to me now.

PR #1622, opened 2022-09-28
Add shortcut for copying API keys

phl0:
This adds a little icon which can be used to copy api keys instead of selecting the text and copying manually. Demo:

https://www.youtube.com/watch?v=9SpDfPOkIqw

### DIFF
--- a/application/language/english/general_words_lang.php
+++ b/application/language/english/general_words_lang.php
@@ -121,3 +121,5 @@ $lang['gen_from_date'] = 'From date';
 $lang['gen_this_qso_was_confirmed_on'] = 'This QSO was confirmed on';
 
 $lang['error_no_logbook_found'] = 'No logbooks were found. You need to define a logbook under Station Logbooks! Do it here:';
+
+$lang['copy_to_clipboard'] = 'Copy to clipboard';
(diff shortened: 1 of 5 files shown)


AndreasK79:
This only works if your page is served via HTTPS. I just could not get it to work locally, and found that it is because I don't have HTTPS on my dev machine. 

Some info here: https://stackoverflow.com/questions/71873824/copy-text-to-clipboard-cannot-read-properties-of-undefined-reading-writetext

phl0:
whoooops. My dev env is also HTTPS. So didn't notice. Have to check if a non-HTTPS solution exists....

AndreasK79:
A hack is described in the link above.

phl0:
Currently working on https://github.com/magicbug/Cloudlog/pull/1582 to get that into V2. Will look here afterwards.

phl0:
tbh the page looks like crap if served via HTTP anyway:

Screenshot from 2022-10-01 14-14-16

So not sure if we should try to fix this at all? @magicbug?

phl0:
Screenshot from 2022-10-01 14-17-45

Screenshot from 2022-10-01 14-17-53

phl0:
Statistics don't even work at all ....

Screenshot from 2022-10-01 14-17-53

PR #1624, opened 2022-09-30
[Cat] Some fixes for backward compatibility with CloudLogCatQt

kb-light:


phl0:
Good point. Tested ok here. 👌

PR #1658, opened 2022-10-12
[CQ Award] New map added

AndreasK79:


phl0:
Looks very nice on my end. Just noticed that there is a zone that I haven't worked at all ... :-)

(maybe make the borders a little thinner on the lower zoom level?)

Screenshot from 2022-10-12 10-52-04

AndreasK79:
@phl0 Then you need to get on the radio and work the last one :)

I guess borders can be adjusted according to zoom level, but that's something I need to look into.

phl0:
> I guess borders can be adjusted according to zoom level, but that's something I need to look into.

Nothing to prevent the merge. I will also have a look. Seems it is like proportinal to zoom level.

magicbug:
This is rather lovely!

phl0:
@AndreasK79 I made a patch for the issue with the tile URL. Actually fixes the issue plus allows to use the globally configured tile URL as source for the map :) See https://github.com/AndreasK79/Cloudlog/pull/11

AndreasK79:
@phl0 thanks, Yes, that was added quite recently. I had forgotten.

AndreasK79:
Works for me.

phl0:
> Works for me.

Ok there you go: https://github.com/AndreasK79/Cloudlog/pull/12

AndreasK79:
Looks sharper and better now. Think it should be ready for merge. Did you have a test run yet @magicbug ?

magicbug:
> Looks sharper and better now. Think it should be ready for merge. Did you have a test run yet @magicbug ?

Yep working good here

PR #1663, opened 2022-10-13
Convert Contest serial exchange to integer

DL2CC:
some logbook programs allow for non number characters in the exchange serial number field, cloudlog would crash on import. By converting those fields to integer we probably lose some (worthless) information but still have it import correctly


magicbug:
Dont mind losing useless info! thanks

PR #1699, opened 2022-10-18
Change WWFF link to CQGMA

m0urs:
Currently, if you click on a WWFF reference, you will be redirected to "https://wwff.co/". Unfortunately this link often has less information available about that reference than "https://www.cqgma.org/".  Even on the DX cluster page of "https://wwff.co/" you will be redirected to "https://www.cqgma.org/" if you click on a reference on that page. If you still want to go to the original entry on "https://wwff.co/" you can still do that directly from the page on "https://www.cqgma.org/".

I therefore suggest to change that link within Cloudlog as well.

### DIFF
--- a/application/views/view_log/qso.php
+++ b/application/views/view_log/qso.php
@@ -212,7 +212,7 @@
                     <?php if($row->COL_WWFF_REF != null) { ?>
                     <tr>
                         <td><?php echo $this->lang->line('gen_hamradio_wwff_reference'); ?></td>
-                        <td><a href="https://wwff.co/directory/?showRef=<?php echo $row->COL_WWFF_REF; ?>" target="_blank"><?php echo $row->COL_WWFF_REF; ?></a></td>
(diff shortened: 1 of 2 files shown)


phl0:
I like this. But should be PRred against `dev` instead of `master` @magicbug?

magicbug:
I can change it against the dev branch

While the other site might have more information, I'm keen to link Cloudlog to the actual organisers of the awards as overtime they are a safer bet in case of people falling out.

m0urs:
> While the other site might have more information, I'm keen to link Cloudlog to the actual organisers of the awards as overtime they are a safer bet in case of people falling out.

As I said: Even the WWFF site itself links to CQGMA on their WWFFwatch DX-cluster ...

phl0:
I talked to some guys around WWFF and CQGMA. The latter periodically pulls data from wwff.co and enriches it. This is an automatic process on cqgma servers. So from my point of view it is valid to use cqgma as source here. Would vote vor merging this (also see https://github.com/magicbug/Cloudlog/pull/1704)

PR #1705, opened 2022-10-19
Auto QTH function configurable on frontend

phl0:
This addresses https://github.com/magicbug/Cloudlog/discussions/1698 and make it user-configurable on frontend.

### DIFF
--- a/application/config/cloudlog.php
+++ b/application/config/cloudlog.php
@@ -110,19 +110,3 @@ $config['qso_date_format'] = "d/m/y";
 */
 
 $config['map_6digit_grids'] = FALSE;
-
-
-/*
-|--------------------------------------------------------------------------
-| Automatically populate the QTH
-|--------------------------------------------------------------------------
-|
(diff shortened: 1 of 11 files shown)


m0urs:
In general both functions are working for me. However, automatic creation of the missing field "user_qth_lookup" did not work. Both "application/migrations/103_add_user_wwff_lookup.php" and "application/migrations/104_user_auto_qth_option.php" are creating the same field ("user_wwff_lookup"). Looks like a copy&paste error?

m0urs:
And maybe it would be possible to also change the link, which appears after entering a WWFF reference, from WWFF to https://www.cqgma.org/ ?

phl0:
Mistake fixed. The WWFF link stuff is handled in https://github.com/magicbug/Cloudlog/pull/1699.

m0urs:
Ok, I only changed one link in #1699 yet, so I will change the other occurence as well in my PR

PR #1706, opened 2022-10-19
Add button that leads to search for other QSOs with the same DX station

phl0:
Thought this could be useful (saves me from copying the callsign to the search box). With a click on the button you will be directed to a search for QSOs with the same DX station.

Screenshot from 2022-10-19 23-26-20

### DIFF
--- a/application/views/view_log/qso.php
+++ b/application/views/view_log/qso.php
@@ -297,7 +297,8 @@
 
                     <?php if(($this->config->item('use_auth') && ($this->session->userdata('user_type') >= 2)) || $this->config->item('use_auth') === FALSE) { ?>
                         <br>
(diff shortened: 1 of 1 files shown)


phl0:
Tnx for testing. I can confirm the issue with slashes in the callsign though it is not caused by this patch as far as I can see. Will see if I can find a solution.

phl0:
> Though while testing, I did find that search acts strange when the call has a / in it. It searches for everything that contains what is in front of the slash in the call. I don't know if this is intentional, or a bug.

Strange. I cannot reproduce this... Does it only happen to calls with prefix *and* suffix?!

AndreasK79:
It seems to happen with prefix. It was with the call LA/SP7VC where I noticed this behavior.

AndreasK79:
I'll make a separate issue on this.

phl0:
Oh right. QSL on the issue with prefixed callsigns. As such not an issue of this PR. So got for merge here imho.

AndreasK79:
As mentioned above, it has nothing to do with this PR. I should have made a separate issue when I found it.
Go for merge :)

PR #1712, opened 2022-10-22
[Search] Fix for hidden menu in search result

AndreasK79:


phl0:
confirmed. Works fine on my end. TBH: Always annoyed me as well but never found the time to fix :)

AndreasK79:
Thanks for testing. Stack Overflow had the answer. I tried several solution, but this was the one I got to work.

phl0:
Never underestimate stackoverflow 😅👍

PR #1713, opened 2022-10-22
[CQ Map] Fix when reloading map after selection change

AndreasK79:


phl0:
Works fine now and does not produce a mess on showing other bands.

AndreasK79:
Just had to find the working solution.

PR #1726, opened 2022-10-28
[Search] Fix for partial search when callsign contains slash

AndreasK79:


phl0:
There is an issue with the code if I search for a simple callsign:

Screenshot from 2022-10-28 11-19-46

`fixedid` is undefined if id2 is set...

AndreasK79:
Oops. Thanks for testing. I was only focused on callsigns with /. Pushed a fix now.

phl0:
Ok, re-tested and works fine now.

PR #1730, opened 2022-10-30
[Contesting] Triggers table search when entering call

AndreasK79:


phl0:
works. Search field of table is filled in parallel :)

Screenshot from 2022-10-30 08-45-32

AndreasK79:
Thanks

PR #1734, opened 2022-10-31
[Search] Added tab to check for duplicate QSOs

AndreasK79:


AndreasK79:
I've added another tab in search:
image
When clicking callsign, you go to the search and get all QSOs listed with the callsign.

AndreasK79:
I know. I just noticed this myself. The list is correct, but only the first call works. 

I will look into it.

AndreasK79:
Thanks. Not sure how to do the back-link though. The way this is built, means that they are separate pages. I will think about it.
Right now, I'm thinking about this one:
image

AndreasK79:
> on my system it does not work, it delivers the dublicate list but clicking on a callsign delivers the list of the first row callsign.
This has been fixed in dev now.

dg9vh:
Yes, working now :-) Mni tnx!

PR #1739, opened 2022-11-01
[Search] Can search for wrong CQ Zone logged

AndreasK79:


phl0:
Tested ok with the fix I PRred in https://github.com/AndreasK79/Cloudlog/pull/13.
If that is merged/fixed it can be merged from my point of view.

AndreasK79:
Thanks for testing it out, and submitting a fix. Strange that I didn't get an error myself. I ran the script several times. Ah well. 

@magicbug test before merge?

PR #1744, opened 2022-11-07
Add grid to result table

phl0:
A little enhancement for @AndreasK79's wrong CQ zone search.


AndreasK79:
Ok for me, but perhaps do it like other places. If grid is blank, use vucc-grids?

phl0:
> Ok for me, but perhaps do it like other places. If grid is blank, use vucc-grids?

Roger. Fixed.

AndreasK79:
Just tested. Cell is blank here. I'll see if I can find out why.

AndreasK79:
Problem fixed. Probably copy/paste error, then forgetting to change variable name :)

phl0:
> Problem fixed. Probably copy/paste error, then forgetting to change variable name :)

Darn my bad. Tnx for fixing :)

Issue #1743, opened 2022-11-06
SOTA CSV Export does not interpret QSOs correctly


florianprischl:
**Describe the bug**
The SOTA CSV Export does not produce a correct list of QSOs for the given parameters. It appears that the generated CSV file switches the "other station" and "own station" SOTA reference. As a result, CSV files imported to https://www.sotadata.org.uk do not result in correct logs.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Station Menu (top right) --> 'SOTA CSV Export' --> Choose relevant parameters to generate logs for SOTA
2. Click on 'Export', download the CSV file.
3. Go to https://www.sotadata.org.uk
4. Go to 'Menu' (top left of the page) --> 'Submit Logs' --> 'Upload CSV/TSV log file' --> Chose your previously downloaded CSV file
5. See error on the import confirmation page: Only QSOs that include a SOTA reference for the other station will be shown.

Additional symptom: The CSV file actually only includes the wrong QSOs. The issue is obviously not in the formatting of the CSV file, but what cloudlog includes in the CSV file in the first place.

**Expected behaviour**
The CSV file should include all QSOs following the parameters chosen on the CSV Export page.

**References**
Check out the following files which helped me diagnose the problem:

1. This is an erroneous CSV file. It only includes 1 QSO even though it should include 4: https://oe1flp.at/files/OE1FLP-SOTA-06112022-1935.csv
2. This is the ADIF file for the same station and the same day, created through cloudlog's ADIF Export feature: https://oe1flp.at/files/OE1FLP-06112022-1938.adi

The ADIF file creates a correct log upon import on https://www.sotadata.org.uk/, the CSV file does not.

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro 21H2
 - Browser Google Chrome
 - Version 107.0.5304.88

**Additional context**
Running self-hosted cloudlog 2.0, fully updated.

phl0:
Hmm think I built this and uploaded QSOs correctly several times. That was chaser logs only. I will have a look. @magicbug please assign to me.

phl0:
I did a test export from my Cloudlog instance. The result is:

![Screenshot from 2022-11-07 15-52-12](https://user-images.githubusercontent.com/7112907/200340953-2f81a5f1-d098-44d3-a8bd-1f8913d6f67d.png)

Which from my point of view is inline with the example file on https://www.sotadata.org.uk/en/upload/activator/csv/info:

[...]

PR #1747, opened 2022-11-08
Also export QSO where MY_SOTA_REF is set

phl0:
This fixes https://github.com/magicbug/Cloudlog/issues/1743 so that also QSOs with COL_MY_SOTA_REF are exported (e.g. my SOTA activcations and S2S QSOs).


phl0:
Reverted last two commits as those are part of https://github.com/magicbug/Cloudlog/pull/1748.

PR #1759, opened 2022-11-09
Added partial russian translation.

magicbug:
Added static callsign lookup buttons for QRZ.ru


AndreasK79:
@magicbug seems to me you lost this one. I can't find the language files in dev or master.

PR #1762, opened 2022-11-10
Russian translation

cats-shadow:
Updated langiuage files for russian language
Rename recv/RECV to rcvd/RCVD to comply with ADIF fields and avoid]


magicbug:
Changed branch to dev but looks good!

PR #1772, opened 2022-11-11
Accept P12 cert mime type, fix error page for failed LOTW cert uploads

agrif:
My Cloudlog install failed to accept my LOTW certificate because it had what it viewed as an invalid mime type. The page it displayed to relate this error was also broken. This PR fixes both issues.

I can't be certain, but I'm pretty sure this is the same problem found in #1192.

Tracking this bug down was a wild goose chase. In summary:

 * CodeIgniter's upload library checks a file's mime type against allowed types for the given extension. Cloudlog was previously configured to only accept 'application/octet-stream'.
 * CodeIgniter attempts to learn the mime type in three ways:
   * If the `file` tool is installed, it is run on the file. `file` thinks these certs are 'application/octet-stream'.
   * If the PHP 'fileinfo' module is installed, it runs the `mime_content_type` function, which returns 'application/octet-stream'.
   * If neither of these are installed, it uses the 'Content-Type' header provided by the browser. Both Firefox and Chrome recognize LOTW certs as 'application/x-pkcs12'.

So, if you run Cloudlog with neither `file` nor the PHP 'fileinfo' module, LOTW cert uploads failed. When rendering the error page, the controller doesn't set the 'dxcc_list' variable, and the resulting broken error page looks like a DXCC info problem, even though this is completely unrelated to the actual issue.

I've added the correct mime type to the acceptable list, and along the way, fixed the error page.

### DIFF
--- a/application/config/mimes.php
+++ b/application/config/mimes.php
@@ -117,7 +117,7 @@ return array(
 	'json'  =>	array('application/json', 'text/json'),
 	'pem'   =>	array('application/x-x509-user-cert', 'application/x-pem-file', 'application/octet-stream'),
 	'p10'   =>	array('application/x-pkcs10', 'application/pkcs10'),
-	'p12'   =>	'application/octet-stream',
+	'p12'   =>	array('application/octet-stream', 'application/x-pkcs12'),
 	'p7a'   =>	'application/x-pkcs7-signature',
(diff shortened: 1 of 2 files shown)


phl0:
Good work. 👍

Issue #1783, opened 2022-11-15
two active stations


DL3EL:
After the upgrade to Cloudlog 2.0 yesterday I have two active stations. I do not know, how I managed to do that. It is not possible to get it inactive for deletion. Any ideas?

See screenshot.

73, Thomas, DL3EL

**Desktop (please complete the following information):**
 - OS: [Raspian Buster]
![Clipboard01](https://user-images.githubusercontent.com/35669080/201893117-3b609c38-09f6-4ac1-b487-e6739560bc15.jpg)

phl0:
Thats not an expected result :D What happens if you activate another station?

magicbug:
Definitely a bug, fastest method to fix would be to edit the MySQL table station_profile and find anything with an "Station_Active" then set them to null, once you do this you'll need to set a location to active in the GUI to get things working correctly again.

If your not comfortable can make a tool to fix this later :)

DL3EL:
> Definitely a bug, fastest method to fix would be to edit the MySQL table station_profile and find anything with an "Station_Active" then set them to null, once you do this you'll need to set a location to active in the GUI to get things working correctly again.
> 
> If your not comfortable can make a tool to fix this later :)

will try, can you point my to the location if the table?

phl0:
The table name is `station_profile`. Try something like:

```
UPDATE `station_profile` SET `station_active` = NULL WHERE `station_id` = 7;
```

DL3EL:
thanks, that has worked. Problem solved..

phl0:
I made a fix to prevent such in the future. See https://github.com/magicbug/Cloudlog/pull/1786.

PR #1786, opened 2022-11-15
Disable all other station profiles on activating one

phl0:
This fixes the issue mentioned in https://github.com/magicbug/Cloudlog/issues/1783 by deactivating all station locations of that specific user before activating another one.


AndreasK79:
Tested working ok here.

PR #1791, opened 2022-11-16
Basic check for non-future QSO date

phl0:
This addresses https://github.com/magicbug/Cloudlog/discussions/1775.
@magicbug @AndreasK79 could you please spend a second pair of eyes if this makes sense? My tests were positive.

### DIFF
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -42,7 +42,7 @@ class QSO extends CI_Controller {
 
 		$this->load->library('form_validation');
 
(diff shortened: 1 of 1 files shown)


AndreasK79:
@phl0 tested working ok with me.

AndreasK79:
If the date is less than todays date, and you change to 2023, you can log the qso.

phl0:
> If the date is less than todays date, and you change to 2023, you can log the qso.

Well I expected that it would be not too easy - especially with different date formats and so on. I will have another look. Tnx for experimenting.

PR #1801, opened 2022-11-21
Make the QSL colors override any other stuff

phl0:
Initially reported by @dg9vh. The skins blue and superhero do not cover the QSL icon colors after merging https://github.com/magicbug/Cloudlog/pull/1792:

Screenshot from 2022-11-21 09-46-46

So those are not marked as !important so that external bootstrap css stuff cannot overwrite. 
@dg9vh can you please test?

### DIFF
--- a/assets/css/blue/overrides.css
+++ b/assets/css/blue/overrides.css
@@ -13,46 +13,6 @@ thead>tr>td {
     border-width: 2px;
 }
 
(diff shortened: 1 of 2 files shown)


phl0:
This PR breaks the inital code changes. Have to find a sifferent solution. Closing here.

PR #1803, opened 2022-11-24
Fix CSS overrides for cosmo and superhero

phl0:
The two skins cosmo and superhero have a separate CSS for links. This overrides the behaviour for eqsl-green classes so that colors are not overriden (i.e. just being white).


phl0:
Wrong branch ... :(

PR #1805, opened 2022-11-24
Fix CSS overrides for cosmo and superhero

phl0:
The two skins cosmo and superhero have a separate CSS for links. This overrides the behaviour for eqsl-green classes so that colors are not overriden (i.e. just being white).

### DIFF
--- a/assets/css/blue/overrides.css
+++ b/assets/css/blue/overrides.css
@@ -13,46 +13,6 @@ thead>tr>td {
     border-width: 2px;
 }
 
-.eqsl-green {
-    color: #00A000;
-    font-size: 1.1em;
-}
-
-.eqsl-red {
-    color: #F00;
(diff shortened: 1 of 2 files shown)


phl0:
I found no better way. So this is basically what https://github.com/magicbug/Cloudlog/pull/1801 had already.

phl0:
@DL3EL I guess this will get merged to fix the two skins. This means you will have to add "!important" to your eqsl and/or other colors in CSS. See the changes in this PR. Found not other way to fix this. Sri

DL3EL:
Thanks, no problem with that. I will update as soon as it is publicly committed.
Thomas

phl0:
Roger that. So should be good for merge.

magicbug:
Into the dev branch it goes!

PR #1827, opened 2022-12-07
Add FreeDV and M17 submodes of digital voice

phl0:
This adds M17 and FreeDV as submodes of DIGITALVOICE.


phl0:
Ready for merge now.

Issue #1837, opened 2022-12-09
Backup logbook.adi header incorrect


pa8s:
**Describe the bug**
Backup logbook.adi header has an incorrect PROGRAMVERSION tag.

**To Reproduce**
Steps to reproduce the behaviour:
1. Click on 'Admin'
2. Scroll down to 'Backup'
3. Click on 'Backup ADIF data'
4. Click on generated link.
5. Open in text editor.
6. See error at line 4:` <PROGRAMVERSION:3>Version 1.7`

**Expected behaviour**
`<PROGRAMVERSION:11>Version 1.7`

AndreasK79:
Fixed in latest dev.

pa8s:
Quick fix. Thank you. :)

PR #1840, opened 2022-12-11
Fix ADIF version tags

phl0:
@AndreasK79 had fixed one occurrence but there are more. Moreover I would like to propose not to include "version" in the version tag itself. Seems somewhat doubled :) Hope @AndreasK79 does not mind. If so I will apply his patch to the other places in the code.

Fixes https://github.com/magicbug/Cloudlog/issues/1837 and similar code parts.

### DIFF
--- a/application/views/adif/data/clublog.php
+++ b/application/views/adif/data/clublog.php
@@ -1,7 +1,7 @@
 Cloudlog ADIF export
 <ADIF_VER:5>3.1.4
 <PROGRAMID:<?php echo strlen($this->config->item('app_name')); ?>><?php echo $this->config->item('app_name')."\r\n"; ?>
-<PROGRAMVERSION:<?php echo strlen('Version ' . $this->config->item('app_version')); ?>><?php echo $this->config->item('app_version')."\r\n"; ?>
+<PROGRAMVERSION:<?php echo strlen($this->config->item('app_version')); ?>><?php echo $this->config->item('app_version')."\r\n"; ?>
 <EOH>
 
 <?php
--- a/application/views/adif/data/clublog.php
+++ b/application/views/adif/data/clublog.php
@@ -1,7 +1,7 @@
 Cloudlog ADIF export
-<ADIF_VER:5>3.1.2
+<ADIF_VER:5>3.1.4
 <PROGRAMID:<?php echo strlen($this->config->item('app_name')); ?>><?php echo $this->config->item('app_name')."\r\n"; ?>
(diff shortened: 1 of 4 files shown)


application/views/adif/data/clublog.php
```diff
@@ -1,7 +1,7 @@
 Cloudlog ADIF export
-<ADIF_VER:5>3.1.2
+<ADIF_VER:5>3.1.4
 <PROGRAMID:<?php echo strlen($this->config->item('app_name')); ?>><?php echo $this->config->item('app_name')."\r\n"; ?>
-<PROGRAMVERSION:<?php echo strlen('Version ' . $this->config->item('app_version')); ?>>Version <?php echo $this->config->item('app_version')."\r\n"; ?>
+<PROGRAMVERSION:<?php echo strlen('Version ' . $this->config->item('app_version')); ?>><?php echo $this->config->item('app_version')."\r\n"; ?>
```
AndreasK79, line 4: Need to remove version in strlen.

application/views/adif/data/exportall.php
```diff
@@ -3,9 +3,9 @@
 	header('Content-Disposition: attachment; filename="'.$this->session->userdata('user_callsign').'-'.date('dmY-Hi').'.adi"')
 ?>
 Cloudlog ADIF export
-<ADIF_VER:5>3.1.2
+<ADIF_VER:5>3.1.4
 <PROGRAMID:<?php echo strlen($this->config->item('app_name')); ?>><?php echo $this->config->item('app_name')."\r\n"; ?>
-<PROGRAMVERSION:<?php echo strlen('Version ' . $this->config->item('app_version')); ?>>Version <?php echo $this->config->item('app_version')."\r\n"; ?>
+<PROGRAMVERSION:<?php echo strlen('Version ' . $this->config->item('app_version')); ?>><?php echo $this->config->item('app_version')."\r\n"; ?>
```
AndreasK79, line 8: Need to remove version in strlen.

AndreasK79:
@phl0 I don't mind at all. Just need to remove version in strlen.

application/views/adif/data/clublog.php
phl0: Tnx. Wasn't aware that this was present in the other files. Fixed now.

application/views/adif/data/exportall.php
phl0: Fixed.

phl0:
Fixed the other files. Ready for merge.

AndreasK79:
Tested adif-generating ok. Merge time.

PR #1864, opened 2022-12-23
Add "Electronic" as QSL selection

m0urs:
In the new advanced view, I am missing a possibility to mark QSOs for "Electronic QSLs". I use this e.g. if someone sends me a mail with a QSL card as attachment. I normally will then return my QSL card by the same way.

This pull requests adds this functionality.


AndreasK79:
@m0urs thanks for the PR. Yes, additions are welcome. This was a first release, so I guess people will come up with several requests here.

Tested working ok.

Issue #1863, opened 2022-12-22
[PHP 8.1] HTTP500 during ADIF export


magicbug:
### Discussed in https://github.com/magicbug/Cloudlog/discussions/1862

<div type='discussions-op-text'>

<sup>Originally posted by **HakanKoseoglu** December 22, 2022</sup>
Possible regression. I am not sure if I should raise an issue. Can anyone else replicate this?

Code at: commit 8b9d2598ba248e9cff60ba8b26e1f25f43d4833a (HEAD -> master, tag: 2.3, origin/master, origin/HEAD)

Edit: ran it on developer mode, additional log:

_An uncaught Exception was encountered
Type: Error

Message: Call to a member function format() on bool

Filename: /var/www/html/application/models/Adif_data.php

Line Number: 87_

Open ADIF Import/Export, select Export.

Selected logbook, did not click on any other options. Fails with HTTP500 as below, which isn't very helpful. 

172.20.0.1 - - [22/Dec/2022:22:29:24 +0000] "POST /index.php/adif/export_custom HTTP/1.1" 500 295 "https://logbook.koseoglu.org/index.php/adif" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36"

If I enter start/end dates, this executes w/o a problem.

![image](https://user-images.githubusercontent.com/10955987/209236783-ac891341-880e-49eb-990f-12116fa11fa0.png)
</div>

magicbug:
Hmm just testing this on my live and dev

- Error if I don't select a Location
- ADIF generates fine if dates or filled in or blank.

What PHP version?

pa8s:
I can reproduce your error on my system. 
code at commit 5d2ca034
PHP 8.1.13

PR #1866, opened 2022-12-23
Fix date validation on ADIF export

mattmelling:
#1862, #1863

If date fields are not set when ADIF export form is submitted, the dates are submitted as empty strings.

E.g. under Firefox on my system:

```
-----------------------------258695364640644243403098313945
Content-Disposition: form-data; name="station_profile"

2
-----------------------------258695364640644243403098313945
Content-Disposition: form-data; name="from"


-----------------------------258695364640644243403098313945
Content-Disposition: form-data; name="to"


-----------------------------258695364640644243403098313945--
```

`Adif_data->export_custom` was previously checking if `$from, $to != 0` which in this case is true, but then the blank string is passed through to `DateTime::createFormat` and interpreted as a truthy value, which causes the error referenced above.

This now works for me when exporting all QSOs to ADIF under PHP8.

### DIFF
--- a/application/models/Adif_data.php
+++ b/application/models/Adif_data.php
@@ -82,12 +82,12 @@ class adif_data extends CI_Model {
         $this->db->where($this->config->item('table_name').'.station_id', $station_id);
 
         // If date is set, we format the date and add it to the where-statement
-        if ($from != 0) {
+        if ($from) {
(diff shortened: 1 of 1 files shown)


AndreasK79:
Tested working ok here. Thanks.

PR #1868, opened 2022-12-23
LotW warning for cert expiry

phl0:
This addresses https://github.com/magicbug/Cloudlog/discussions/1860. Basically shows a notification for certificates that expire within 30 days and a warning for certificates that expired already.


AndreasK79:
Can't really test all functionality, as I do not have LoTW set up in Cloudlog.  However, no errors when nothing is set up either, so that should be a good thing :) Code looks ok.

@magicbug this one is up to you.

phl0:
Thank you for your feedback @AndreasK79

PR #1874, opened 2022-12-27
[DXCC identification] This identifies when DXCC comes last

AndreasK79:
This should fix the issue when prefix comes last, like N1MM/VP9.

DJ5CW wrote a PERL script for identifying a DXCC. I converted it to PHP. This is part of that conversion, which is now adapted to Cloudlog.

We can never 100% identify a DXCC, but this should help improve the identification.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -2833,14 +2833,24 @@ class Logbook_model extends CI_Model {
       $call = "KG4";
     } elseif (preg_match('/(^KG4)[A-Z09]{1}/', $call)) {
       $call = "K";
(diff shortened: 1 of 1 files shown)


phl0 (approved):

phl0:
This patch looks good to me. Tested with a hand full of callsigns with DXCC prefix and suffix. All worked fine.

magicbug:
I've pulled this PR out for now as it completely breaks /P and also seems to assign wrong DXCCs to prefixes

AndreasK79:
Isn't the point of a dev branch to fix the things, not just revert? I didn't have time to look into it before you reverted it.
Besides, wouldn't it have been better to test the PR before merge?

PR #1877, opened 2022-12-28
adapt mode in advanced logbook

Werzi2001:
In my environment I noticed that I have both mode "FM" and "FM|" (and "CW" and "CW|") in my advanced logbook mode filter. Of these only the ones with | work as the ones without | lead to a server side error in the search. Therefore I adapted the modes to always contain | as the search implementation seems to require the | anyways. This seems to work fine (at least with my data).


magicbug:
This is already addressed in the dev branch.

phl0 (changes requested):
Think we have something in dev already.

Werzi2001:
Sorry I didn't check the dev branch. You're right. It works there.

PR #1882, opened 2022-12-29
Fix oqrs flaws

phl0:
This fixes 3 errors with OQRS:
1. Set qsoid to 0 in case someone makes a not-in-log request. Otherwise the request will not be saved as qsoid has no default value in SQL table
2. Corrects ID of message input text field so that it is stored to DB (currently silently ignored)
3. Loads eQSL and LotW lang for OQRS to fix display errors (and missing lang tags in logs)

@AndreasK79 pls verify and approve.

### DIFF
--- a/application/controllers/Oqrs.php
+++ b/application/controllers/Oqrs.php
@@ -9,6 +9,8 @@ class Oqrs extends CI_Controller {
 
 	function __construct() {
 		parent::__construct();
+		$this->lang->load('lotw');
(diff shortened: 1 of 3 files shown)


phl0:
Fixed conflicts with recent changes in dev branch.

magicbug:
Just checked this here Andy and seems to fix the issues nicely

AndreasK79:
Excellent. Haven't had time to check yet.

PR #1884, opened 2022-12-30
[DXCC identification] Fix for prefix coming last

AndreasK79:
Identified the trouble with the last PR.

I think I've fixed the cases that were wrong now.


phl0:
I have tested the rewritten/corrected implementation yesterday. All my test cases went OK.

AndreasK79:
Yes, @phl0 and I tested this one a lot yesterday.
Regex needed to be adjusted so that portable was identified correctly.
A variable was also needed to be adjusted to that SQ1 was identified correctly as well.

PR #1886, opened 2022-12-30
[OQRS] Changed to visitor header

AndreasK79:
Changed to visitor header. Since we don't have a slug here, I added checks to hide if empty.
I didn't see the point of having "DEVELOPMENT" tag either, so I removed it.


magicbug:
Yep def not needed on there!

PR #1887, opened 2023-01-01
[Station logbooks] Tweaked the view a little bit and corrected the wa…

AndreasK79:
…rning text on delete.


phl0:
I like this. Tests went ok.

phl0 (approved):

PR #1889, opened 2023-01-01
[User view] Tweaked the edit layout a bit

AndreasK79:


phl0:
Tested and I like the new (old) view. :-)

PR #1892, opened 2023-01-02
Beautify table and add QRZ/HAMQTH links

phl0:
This makes the result table a little more beautiful and adds qrz.com and hamqth.com links for lookups.


AndreasK79:
Tested. Looks much better!

Issue #1900, opened 2023-01-03
[PHP 8.1] Message: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated


pa8s:
**Describe the bug**
When switching between tabs on search window, PHP error message is visible.

**To Reproduce**
Steps to reproduce the behaviour:
1. In main window, click 'Search'.
2. Click on any tab other tab than 'Search'.
3. After switching to other search tab, click on 'Search' tab.
4. See error in screenshot.

**Expected behaviour**
No error.

**Screenshots**
![BugSearch](https://user-images.githubusercontent.com/77061865/210370546-3c47c442-44b9-468f-862d-7f10bfd3da56.png)

**Logfile**
```
ERROR - 2023-01-03 13:55:11 --> Severity: 8192 --> htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated /mnt/www/application/views/search/main.php 30
```

**Server**
OS: FreeBSD 13.1
Server: nginx/1.22.1
PHP: 8.1.13
Codeigniter: 3.2.0-dev
Git branch: dev
Git commit: 254ee192
Git Tag: 2.1-206-g254ee192

PR #1901, opened 2023-01-03
Prevent passing null to htmlspecialchars

phl0:
Fixes: https://github.com/magicbug/Cloudlog/issues/1900


pa8s:
Verified. Fixed. :)

phl0:
> Verified. Fixed. :)

tnx.

PR #1911, opened 2023-01-09
Fix hide login

phl0:
This partially reverts my previous commit as it would hide the login button on all public pages. But instead it should only be hidden in the oqrs section. So better use an if here.


AndreasK79:
Tested ok.

Issue #1776, opened 2022-11-14
Feature request: Additional way to view new eQSL confirmations


ofadam:
When a user currently checks for new eQSL matches, they get a screen that shows a list of QSOs that either have a new or existing match. I would love to have a way to click somewhere on that same screen and look through all the new images, especially since finding older matches (sometimes from a few years ago) is quite difficult currently.

<img width="958" alt="eqsl_viewimages" src="https://user-images.githubusercontent.com/5504940/211400218-68be50dc-65c2-450a-a646-c5b8636d4e16.png">

Here's the code that currently allows viewing of eQSL images while in the logbook view:

<a class="eqsl-green" href="<?php echo site_url("eqsl/image/".$row->COL_PRIMARY_KEY); ?>" data-fancybox="images" data-width="528" data-height="336">&#9660;</a>

phl0:
Solution proposed with: https://github.com/magicbug/Cloudlog/pull/1912

PR #1912, opened 2023-01-09
Show link to eQSL image on import status page

phl0:
Addresses https://github.com/magicbug/Cloudlog/issues/1776


ofadam:
Just tested this and it's working very nicely.

phl0:
> Just tested this and it's working very nicely.

Thank you for report. Then this one is ready to be imported into dev upstream.

Issue #1910, opened 2023-01-09
Feature request: Pull county data for LoTW confirmations


ofadam:
When syncing with Logbook of the World, certain fields are automatically added to the Cloudlog database, like state and gridsquare. But would be great to add "CNTY" as an updated field for those of us in the United States so that the Awards/US Counties page can become more accurate.

Reference docs for LoTW: https://lotw.arrl.org/lotw-help/developer-query-qsos-qsls/

phl0:
Made a patch with https://github.com/magicbug/Cloudlog/pull/1913. Would you be able to test this and report?

PR #1913, opened 2023-01-09
Import county from LotW matches

phl0:
Addresses https://github.com/magicbug/Cloudlog/issues/1910

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -545,7 +545,13 @@ class Lotw extends CI_Controller {
 						$iota = "";
 					}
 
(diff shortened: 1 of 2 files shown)


ofadam:
Just tested and it works like a charm!

phl0:
Tnx fer test. Ready for merge :)

magicbug:
Are we handling said field already existing? or just going to overwrite with lotw data 

Just checking

phl0:
> Are we handling said field already existing? or just going to overwrite with lotw data
> 
> Just checking

We are going to overwrite with LotW data which makes sense in my eyes. I guess the LotW confirmation has the correct data in case of difference?

magicbug:
Good with me, just double checking

ofadam:
A little context on my request... there are times when the callbook has a county listed, but the operator is mobile (especially with Parks on the Air.) When they submit their LoTW logs, they indicate the correct county – so this makes sure that all is accurate. Thanks to both of you!

PR #1914, opened 2023-01-10
Updated and fixed translation for simplified Chinese.

Lingluoluo:


phl0 (changes requested):
This PR should be done against the dev branch not master.

PR #1939, opened 2023-01-18
Eqsl for awards

phl0:
This adds eQSL (disabled per default) as filter type to CQ/WAS/DXCC awards


AndreasK79:
Tested ok. Merging.

PR #1942, opened 2023-01-19
Fix syntax error

phl0:
Fixes https://github.com/magicbug/Cloudlog/discussions/1941


phl0:
Was introduced in https://github.com/magicbug/Cloudlog/pull/1884.

PR #1954, opened 2023-01-26
Order by primary key as second criterion

phl0:
If I enter post QSOs after SAT contacts sometimes there is a QSO with exactly the same time (with accuracy of a minute). The logbook shows contacts sorted by `COL_TIME_ON`. But when this value is identical for more than one QSOs these QSOs get sorted by `COL_PRIMARY_KEY` in ascending order it seems. I guess it would make more sense to have it in descending order so that if you log one QSO after another they get displayed in the same order as logged.
Example: I logged M5JFS before DL2GRC but the logbook shows:

Screenshot from 2023-01-26 01-01-02

With this patch it would be sorted correctly:

Screenshot from 2023-01-26 01-00-49


AndreasK79:
Makes sense. Thumbs up from me.

PR #1955, opened 2023-01-26
[QSO Entry] Added dxcc summary to spawn when dxcc is identified

AndreasK79:
After entering a callsign, the dxcc will be determined. When that is done, a dxcc summary will be added at the bottom of the qso entry.

Example:
image

### DIFF
--- a/application/views/qso/index.php
+++ b/application/views/qso/index.php
@@ -1,6 +1,6 @@
 <div class="container qso_panel">
 
-<div class="row">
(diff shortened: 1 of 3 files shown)


phl0:
Looks good and works as designed. I like it.

magicbug:
Can we look at making this collapse there's actually a feature in the UI framework https://getbootstrap.com/docs/4.1/components/collapse/

PR #1969, opened 2023-01-30
Add/fetch/lookup continent (from DXCC table)

phl0:
This addresses https://github.com/magicbug/Cloudlog/discussions/1935.
It queries the DXCC table also for CONT info and fills the corresponding fields in the database.
@pa8s: Looking forward to your test results here before un-drafting this PR. :)


phl0:
Closing this in favor of a reworked version in https://github.com/magicbug/Cloudlog/pull/1979

PR #1979, opened 2023-02-04
Add fields for continent and update function

phl0:
As per https://github.com/magicbug/Cloudlog/discussions/1935 this adds fields for continent.
Also added a separate button to update all QSOs missing continent info (tnx @AndreasK79 for SQL command).


AndreasK79:
Tested ok. Go for merge :)

PR #1986, opened 2023-02-06
[Login] Tweaked login

AndreasK79:
Tweaked the login form:
image
A logo with transparent background and cloud was added.
image
image
image


phl0 (approved):
Tnx. works a treat now.

PR #1990, opened 2023-02-08
Make `update_cloudlog.sh` executable

intrigus:


magicbug:
While I understand why some might prefer this executable out the box, for security I think its far better this is setup manually.

PR #1992, opened 2023-02-09
[Forgotten password] Tweaked layout to be the same as login

AndreasK79:
Updated forgotten password form as well:
image


AndreasK79:
I'll play some more with it :)

phl0:
Left of "forgot password" seems a good option to me.

PR #2028, opened 2023-03-07
New variables added in the cloudlog.php config file to control map po…

cats-shadow:
Added new variables to the cloudlog.php config file to control map position at the views with logbook table: visitor, dashboard and logbook overview.

map_at_right -- to move map from top position to trhe right column of the layout (dashboard and visitor views). hide_map_at_overview -- to hide map from logbook overview  view only. 
hide_map -- to hide map from all view above.
изображение

At the logbook overview view map had been moved to the logbook table container. Looks better on the wide themes, IMHO.
изображение

 Added additional css w-qsl class (?) to display QSL card images fitted to the placeholder (width = 55%): **w-qsl{width:55%!important}** in the bootstrap.min.css on the theme folder.

Modified footer.php and qso.php to use this class instead the w-100 to show QSL images.
изображение

For the time column in the QSO widget the same behaviour has been added as for the time column in the visitor view.

Suggested changes had been added.


application/config/cloudlog.php
```diff
@@ -32,7 +32,7 @@
 | Default is: M
 |
 */
-$config['measurement_base'] = 'M';
+$config['measurement_base'] = 'K';
```
phl0, line 35: I think we should leave the defaults here.

application/config/cloudlog.php
```diff
@@ -59,7 +59,7 @@
 |
 */
 
-$config['public_search'] = FALSE;
+$config['public_search'] = TRUE;
```
phl0, line 62: I think we should leave the defaults here.

application/config/cloudlog.php
```diff
@@ -110,3 +110,42 @@
...
+| Show map in right column
+|--------------------------------------------------------------------------
+|
+| Set this to TRUE to show the map in the right column at the Dashboard and Visitor views.
+|
+| Default is: FALSE
+|
+*/
+
+$config['map_at_right'] = TRUE;
```
phl0, line 125: > Default is: FALSE

So should it be here then.

application/views/dashboard/index.php
```diff
@@ -82,11 +82,14 @@ function echoQrbCalcLink($mygrid, $grid, $vucc) {
...
 </div>
 
-<!-- Map -->
-<div id="map" style="width: 100%; height: 350px"></div>
 
 <div style="padding-top: 0px; margin-top: 5px;" class="container dashboard">
 
+<?php if((!$this->config->item('map_at_right')) && (!$this->config->item('hide_map'))) { ?>
+<!-- Map -->
+<div id="map" style="width: 100%; height: 350px"></div>
```
phl0, line 90: We should add a little margin at the bottom to separate from other content.
```
  style="width: 100%; height: 350px; margin-bottom: 15px;"
```

application/views/dashboard/index.php
```diff
@@ -151,6 +154,11 @@ function echoQrbCalcLink($mygrid, $grid, $vucc) {
   </div>
 
   <div class="col-sm-4">
+	<?php if(($this->config->item('map_at_right')) && (!$this->config->item('hide_map'))) { ?>
+	<!-- Map -->
+	<div id="map" style="width: 100%; height: 350px"></div>
```
phl0, line 159: Same as above.

application/views/view_log/qso.php
```diff
@@ -514,7 +514,7 @@
                             echo ' active';
                         }
                         echo '">';
-                        echo '<img class="d-block w-100" src="' . base_url() . '/assets/qslcard/' . $image->filename .'" alt="QSL picture #'. $i++.'">';
+                        echo '<img class="d-block w-qsl" src="' . base_url() . '/assets/qslcard/' . $image->filename .'" alt="QSL picture #'. $i++.'">';
```
phl0, line 517: This leads to huge pix and many scrollbars:

Screenshot from 2023-03-07 12-15-14

application/views/visitor/index.php
```diff
@@ -45,11 +45,14 @@ function echoQrbCalcLink($mygrid, $grid, $vucc) {
...
 </div>
 
-<!-- Map -->
-<div id="map" style="width: 100%; height: 350px"></div>
 
 <div style="padding-top: 0px; margin-top: 5px;" class="container dashboard">
 
+<?php if((!$this->config->item('map_at_right')) && (!$this->config->item('hide_map'))) { ?>
+<!-- Map -->
+<div id="map" style="width: 100%; height: 350px"></div>
```
phl0, line 53: We should add margin-bottom here as well (see above).

application/views/visitor/index.php
```diff
@@ -114,6 +117,11 @@ function echoQrbCalcLink($mygrid, $grid, $vucc) {
   </div>
 
   <div class="col-sm-4">
+	<?php if(($this->config->item('map_at_right')) && (!$this->config->item('hide_map'))) { ?>
+	<!-- Map -->
+	<div id="map" style="width: 100%; height: 350px"></div>
```
phl0, line 122: We should add margin-bottom here as well (see above).

phl0 (changes requested):
There are a few changes to be made before that gets merged into dev IMHO.

magicbug:
This really needs to be implemented using the config settings in the database rather than static files.

phl0:
> This really needs to be implemented using the config settings in the database rather than static files.

Oh yeah. This totally makes sense to me.

application/views/view_log/qso.php
cats-shadow: I thin

> This leads to huge pix and many scrollbars:
> 
This is because the **w-qsl** is absent in the bootstrap.min.css of the theme which you use.

изображение

If image of the QSL is scaled to 100% i got:
изображение

phl0: Well I was testing with the default theme... I guess it should at least work with that theme. Which theme are you using?

application/views/dashboard/index.php
cats-shadow: > We should add a little margin at the bottom to separate from other content.
> 
> ```
>   style="width: 100%; height: 350px; margin-bottom: 15px;"
> ```

Thank you!

application/config/cloudlog.php
cats-shadow: > I think we should leave the defaults here.

Done

application/views/view_log/qso.php
cats-shadow: > Well I was testing with the default theme... I guess it should at least work with that theme. Which theme are you using?
I'm using Cosmo_Wide theme.

Added w-qsl class to CSS for all themes (commit 79bae6f9482696cbe4f3144393389a33efbf0e10). Found a hires QSL card scan image in my log (https://r1blh.cats-home.net/log//assets/qslcard/1538_1673473963.jpg). It looks like all is OK now.

Default Wide theme:
изображение
изображение

Default theme:
изображение
изображение

cats-shadow: @phl0 could you check this, please?

phl0 (changes requested):
Looks ok for me now. Only issue left is:
https://github.com/magicbug/Cloudlog/pull/2028#issuecomment-1458016305

AndreasK79:
And instead of adding the css to each theme, just add it to assets/css/general.css instead.

cats-shadow:
> And instead of adding the css to each theme, just add it to assets/css/general.css instead.

Thank you for an idea!

cats-shadow:
> Looks ok for me now. Only issue left is: [#2028 (comment)](https://github.com/magicbug/Cloudlog/pull/2028#issuecomment-1458016305)



> Looks ok for me now. Only issue left is: [#2028 (comment)](https://github.com/magicbug/Cloudlog/pull/2028#issuecomment-1458016305)

It's still difficult for me. :)

cats-shadow:
> Looks ok for me now. Only issue left is: [#2028 (comment)](https://github.com/magicbug/Cloudlog/pull/2028#issuecomment-1458016305)

It's still difficult for me. :)

phl0:
> It's still difficult for me. :)

I have looked up some changes I made a while back which also added user-configurable items to the frontend. You may check: https://github.com/magicbug/Cloudlog/pull/1450 or https://github.com/magicbug/Cloudlog/pull/1705.

cats-shadow:
> > It's still difficult for me. :)
> 
> I have looked up some changes I made a while back which also added user-configurable items to the frontend. You may check: #1450 or #1705.

Thank you for the examples. I will try to move variables to the database.

AndreasK79:
@cats-shadow if you need more help, take a look at an example here for adding options: https://github.com/magicbug/Cloudlog/pull/2093

phl0:
I guess this one can be closed then ...

PR #2035, opened 2023-03-16
Basic QO-100 Dx Club API integration

hugosilvaparagramanet:
Adds a very basic integration to QO-100 Dx Club API, to allow users to automatically upload QSOs to the club's database.

Currently only works if "realtime upload" is set to "Yes" since there's no manual upload page.

The code is pretty much a copy&paste from the QRZ.com upload functionality.

### DIFF
--- a/application/migrations/115_add_webadif_api_export.php
+++ b/application/migrations/115_add_webadif_api_export.php
@@ -25,18 +25,25 @@ class Migration_add_webadif_api_export extends CI_Migration {
 			$this->dbforge->add_column('station_profile', $fields);
 		}
 
-		if (!$this->db->field_exists('webadif_upload_date', $this->config->item('table_name'))) {
-			$fields = array(
-				"webadif_upload_date datetime DEFAULT NULL"
-			);
-			$this->dbforge->add_column($this->config->item('table_name'), $fields);
-		}
 
-		if (!$this->db->field_exists('webadif_upload_status', $this->config->item('table_name'))) {
-			$fields = array(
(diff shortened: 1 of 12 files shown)


application/migrations/115_add_webadif_api_export.php
phl0, line 1: I guess we should add some if's here to prevent failing on updates. See https://github.com/magicbug/Cloudlog/commit/5ac115ceee15782effb9147e709442696783f030

phl0 (changes requested):
Works ok for me. We should add some if's to the DB migrations scripts though.

application/migrations/115_add_webadif_api_export.php
phl0: Yeah. This works as expected.

phl0 (commented):
Not sure about the two columns in the base table. Normally there is only ADIF fields in there. Should we use user defined fields here @magicbug?

magicbug:
Yep keep them separate, also make sure the table is named something obvious :)

phl0:
Test with separate table also successful here :)

magicbug:
Do we all agree this is OK to merge into dev branch for further testing before going public?

phl0:
I have tested to the latest commit. Merge is ok from my point of view. But maybe we should ask @hugosilvaparagramanet if there is something left to add before :)

hugosilvaparagramanet:
@magicbug @phl0 I've just uploaded a new commit that fixes a duplication while "marking as uploded". 
From my tests the thing is ok, hopefully...

Issue #2036, opened 2023-03-17
Backup function breaks if you configure Cloudlog for index.php to be removed


m0urs:
If you configure Cloudlog to have the index.php to be removed from the URLs (like described here: https://github.com/magicbug/Cloudlog/wiki/Cloudlog.php-Configuration-File#remove-indexphp-from-your-url) then the backup functionality breaks.

You are no longer give the possibility to create a backup but you just get the list of all files in the backup directory:

![image](https://user-images.githubusercontent.com/2588362/225860697-266df76c-6b5a-42ec-969a-0a254f58e2f5.png)

By the way: I am sure we had already discussed that somewhere but I did not find an open issue for it: Currently backup does only make a backup of the current active station location. I would prefer that backup backs up all QSO regardless in which location thy are in.

m0urs:
I guess we can solve this by adding an exception for "/backup/" in .htaccess so that this URI is also rewritten to "/index.php/backup" like this:

```
# If you want to hide 'index.php' from the URL, rename this file to '.htaccess' on your server
# Then change the following variable in /application/config/config.php
#   $config['index_page'] = '';

RewriteEngine On

RewriteCond %{REQUEST_URI} ^/backup/$
RewriteRule ^(.*)$ /index.php?/$1 [L]

RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule ^(.*)$ /index.php?/$1 [L]

```
At least it seems to work for me. I will do some more tests and ten make a PR for that.

m0urs:
Fixed by https://github.com/magicbug/Cloudlog/pull/2037

PR #2041, opened 2023-03-20
Fix: KML export also no longer working if Cloudlog configured to remove index.php

m0urs:
Similar issue as with the backup menu option described in https://github.com/magicbug/Cloudlog/issues/2036


AndreasK79:
@m0urs I renamed it to kmlexport so that this is no longer a problem.

PR #2043, opened 2023-03-20
Mark QSOs as uploaded in batch to improve performance

hugosilvaparagramanet:
Instead of running QSOs one by one, build a batch of 500 and do a multiple insert, as it should be faster.
@phl0 can you test it on you excellent dataset?


phl0:
Surely I will yest that. The "excellent dataset" is the DV of my prod system. No risk no fun 🤣👋

PR #2044, opened 2023-03-20
Show nice message if the user doesn't have QO-100 Dx Club export conf…

hugosilvaparagramanet:
Shows an user friendly message if the user doesn't have API export configured.
@magicbug since you're a native speaker, you'll probably want to correct the copy :-)
Thanks for @AndreasK79 and @phl0 for the suggestion


AndreasK79:
@hugosilvaparagramanet I see you made a PR with it :) I just thought it would make more sense like this. I was just looking through everything, and had to figure out why station locations were empty.

PR #2061, opened 2023-04-10
[Contest logging] Moved contest session from local storage to database

AndreasK79:
Several people requested to be able to use have the contest session on several machines.

I moved the contest session from localstorage in the browser over to a database table to make this possible.

Works ok when testing locally, but issues might exist.


phl0:
Yeah. Serial No migration and increment now works. Tested with two different browsers.

phl0 (approved):
Go for merge from my end (to gain some user feedback on this).

PR #2062, opened 2023-04-11
Make LotW interactions a little more verbose (and less error prone)

phl0:
This addresses issues with LotW downloads as for example mentioned in https://github.com/magicbug/Cloudlog/discussions/1948.
The code now checks whether a file already exists and is writable and if not reports accordingly. If the file does not exists it checks whether the directory is writable and aborts if not. 
And last but not least error messages were improved. The code now checks the file contents for a valid LotW report before importing it and failing. Also error messages for simply empty list of matches has been refactored.
This is for @AndreasK79 for his kind review before merge. Tests and reports more than welcome even before this gets merged.

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -359,7 +359,6 @@ class Lotw extends CI_Controller {
 			|	Download QSO Matches from LoTW
 			*/
 			echo "<br><br>";
-			echo "LoTW Matches<br>";
 			echo $this->lotw_download();
 
 	}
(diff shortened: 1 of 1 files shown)


phl0:
I cannot request a review. So please check @AndreasK79 and @magicbug :)

AndreasK79:
As I do not use LoTW with Cloudlog, I can't test all this. However looking at the code, it does look ok to me. Better error handling is much needed when using LoTW.

Go for merge here.

PR #2066, opened 2023-04-19
Zone lookup

phl0:
I think we could use a little more interactive zone lookup tool here. Tnx to DK5UR for implementing direct links.
(Until @AndreasK79 is ready with the ITU zone implementation).


AndreasK79:
I like this more interactive map. A solution built in can be built, but not sure we need it though.

Go for merge!

PR #2067, opened 2023-04-19
Use DXCC ID pulled from LotW cert

phl0:
We have various issues like https://github.com/magicbug/Cloudlog/discussions/2065 where people seem to set the DXCC for their LotW certs wrong and upload fails. Luckily there is a simple solution for that: We should use the DXCC id that is contained in the LotW cert instead of asking the user to choose the (wrong) DXCC during cert upload.

Screenshot from 2023-04-19 10-44-01

The DXCC/ADIF ID can be found in the additional attributes (OID 1.3.6.1.4.1.12348.1.4). So we go ahead and use that instead of deriving it from dxcc_entities. Special case with ID 0 (i.e. no DXCC as /mm for example) is also handled and displayed accordingly.

Keen to hear your feedback @AndreasK79 and @magicbug.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 117;
+$config['migration_version'] = 118;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -153,20 +153,20 @@ class Lotw extends CI_Controller {
 			}
 
 			// Check to see if certificate is already in the system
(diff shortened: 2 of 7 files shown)


phl0:
So basically the input drop down menu for DXCC is removed and information is derived from the uploaded LotW certificate :)

AndreasK79:
I like this. One place less for users to make error. Uploading certificate works as it should. Sync is not tested, as I don't use LoTW with Cloudlog, but code looks ok! Go for merge.

phl0:
Thank you for your feedback @AndreasK79. I have tested this with some debug statements to prevent actual upload to LotW and then compared the resulting files to be uploaded. No differences :)

phl0:
I was keen and tested this state of the branch on my prod system after making one QSO with JA7BXS:

Screenshot from 2023-04-20 08-14-42

Upload worked:

Screenshot from 2023-04-20 08-12-45

And LotW processed successfully:

Screenshot from 2023-04-20 08-13-28

So all good from my point of view.

Issue #2076, opened 2023-04-24
Warning --> Undefined array key "manual" 


pa8s:
**Describe the bug**
Warning --> Undefined array key "manual" 

**To Reproduce**
Unable to recall performed steps :(

**Expected behaviour**
No error in logfile

**Logfile**
ERROR - 2023-04-24 21:05:34 --> Severity: Warning --> Undefined array key "manual" /mnt/www/application/views/interface_assets/footer.php 2397

**Server**
OS: FreeBSD 13.1
Server: nginx/1.24.0
PHP: 8.1.18
Codeigniter: 3.2.0-dev
Git branch: dev
Git commit: https://github.com/magicbug/Cloudlog/commit/d6139439734f92fbd81cfcb85c3a616fff322832
Git Tag: https://github.com/magicbug/Cloudlog/releases/tag/2.3.1

phl0:
Happens upon editing contests. Fix will be included with https://github.com/magicbug/Cloudlog/pull/2078.

PR #2078, opened 2023-04-24
Ignore manual parameter on editing contests

phl0:
Fixes: https://github.com/magicbug/Cloudlog/issues/2076.


AndreasK79:
Tested ok.

PR #2085, opened 2023-04-26
Show badge and hints about deleted DXCCs

phl0:
And here is the next one. Was based on the former patch to remove the redundant DXCC information. Still need to figure out how best show the badge in advanced logbook though.


magicbug (changes requested):
Looks good I'd suggest "Deleted DXCC"

phl0:
> Looks good I'd suggest "Deleted DXCC"

QSL on that. Fixed.

magicbug (approved):
Perfect

PR #2090, opened 2023-04-27
Make stats publicly usable with API key

phl0:
Tuned the statistics function so that it allows for public usage (i.e. not logged-in) with an API key that is at least read-only.

Was requested in https://github.com/magicbug/Cloudlog/discussions/1697. So I kindly ask @M0TZO for a test and feedback here.

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -672,14 +672,14 @@ class API extends CI_Controller {
 	*
 	*/
 
-	function statistics() {
+	function statistics($key = null) {
 		header('Content-type: application/json');
 		$this->load->model('logbook_model');
 
-		$data['todays_qsos'] = $this->logbook_model->todays_qsos();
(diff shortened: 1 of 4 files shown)


AndreasK79:
Tested working. Perhaps it's worth mentioning that url should be called like this: cloudlogurl/index.php/api/statistics/apikey

Substitue cloudlogurl with the correct url, and apikey with the readonly key generated in api.

phl0:
True. @M0TZO reported an error last night. Details missing. So waiting for his detailed feedback here. I guess the error was within his python client but not sure.

AndreasK79:
Right. Well, I got numbers on each point here. They were null values before I added apikey,.

phl0:
> Right. Well, I got numbers on each point here. They were null values before I added apikey,.

Yupp. Same here.

M0TZO:
Once updated, can confirm it's working now. Thanks.

Issue #2055, opened 2023-04-05
[Feature Request] Configure default value for "Power"


m0urs:
I would like to have the value for the transmit power as a config setting in a station location. So whenever a new QO is created for this location, this value should be the preset value (which you would be able to overwrite).

Currently after I start a new Cloudlog session I need to manually set the value every time (and often forget to set it). If you have set the value and saved at least one QSO, the value will be remembered already for the rest of the session, which is ok.

Unfortunately I still have not really understand how the programming model of Cloudlog works, so I can sometimes do small changes by myself, but others I am not able to do :-(

phl0:
At https://github.com/magicbug/Cloudlog/pull/2091 there is a PR that we would love to hear your feedback on. ;-)

m0urs:
Hi Florian, that looks fine to me. I was able to define my 10W into all my locations and it appears for new QSOs now.  Thanks a lot!

phl0:
Anyone else to test here before I undraft this PR? @AndreasK79 @magicbug?

AndreasK79:
Tested ok here.

PR #2091, opened 2023-04-27
Add option for default station power

phl0:
This addresses https://github.com/magicbug/Cloudlog/issues/2055 by adding a parameter to station_locations that holds a value for a default power. This is of course overwritten by CAT or user entered values. Please test and report.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1086,7 +1086,6 @@ $(document).on('keypress',function(e) {
 
 	$('#stationProfile').change(function() {
 		var stationProfile = $('#stationProfile').val();
-      console.log("TEST "+stationProfile);
 		$.ajax({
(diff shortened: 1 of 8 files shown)


application/views/interface_assets/footer.php
```diff
@@ -1084,6 +1084,22 @@ function searchButtonPress(){
 	});
 <?php } ?>
 
+	$('#stationProfile').change(function() {
+		var stationProfile = $('#stationProfile').val();
+      console.log("TEST "+stationProfile);
```
AndreasK79, line 1089: Remove this line first :)

phl0: That was just a test if you are paying attention :D

AndreasK79: :)

PR #2097, opened 2023-04-29
[Map options] Added options for map on dashboard and in logbook

AndreasK79:
Added the map options from the PR that @cats-shadow made, but these can be set in global options.


phl0:
Tested and works ok on my end :) GO for merge.

AndreasK79:
Thanks for testing.

PR #2102, opened 2023-05-01
{Advanced Logbook] Added sat filter

AndreasK79:
By popular request, I added a filter for sats.


phl0:
Tested and works ok.

AndreasK79:
Thanks. Merge time.

PR #2103, opened 2023-05-01
[ADIF Backup] Added so that backup can be made with a read-only api key

AndreasK79:


phl0:
I am getting an error here when calling the backup function within Cloudlog (i.e. logged-in):

```

Type: ArgumentCountError

Message: Too few arguments to function Backup::adif(), 0 passed in /var/www/cloudlog/system/core/CodeIgniter.php on line 485 and exactly 1 expected

Filename: /var/www/cloudlog/application/controllers/Backup.php

Line Number: 23
```

AndreasK79:
Thanks @phl0 for testing and fix.

phl0:
Works here after latest changes.

phl0:
Now also handles completely unauthorized access properly :D

PR #2106, opened 2023-05-02
[Advanced logbook] Added columns for  QSL/LoTW/eQSL confirmations

AndreasK79:
Added the same indications as in the other logbook:
image


phl0:
Works as designed on my end. Seems a good addition to advanced logbook view!

PR #2107, opened 2023-05-03
[Notes backup] Added support for backing up notes with an API key

AndreasK79:


phl0:
Works. All ok from my side :)

PR #2108, opened 2023-05-03
Show radio CAT status above QSO panel if available

phl0:
Addresses https://github.com/magicbug/Cloudlog/discussions/2101 by showing a green hint if CAT data is available and not timed-out.


phl0:
Screenshot:

signal-2023-05-03-16-05-16-169

AndreasK79:
Nice addition :)

PR #2111, opened 2023-05-05
[eQSL] Added cron job for eQSL

AndreasK79:
Call it with cloud-url/eqsl/index.php/upload

Thanks to @phl0 for testing.


phl0:
The URL is /index.php/eqsl/upload :-)

PR #2114, opened 2023-05-07
Fix PHP error while updating SOTA csv file

m0urs:


phl0:
Good find 👍🏼  GO for merge :)

PR #2125, opened 2023-05-09
[User selectable columns] Add Location to possible columns

dg0tm:
508d63dc Make user_column5 optional
0cf0b590 Add Location (station_profile_name) to possible columns


phl0 (approved):
Works as expected.

PR #2126, opened 2023-05-09
Remove useless file writing

phl0:
This addresses https://github.com/magicbug/Cloudlog/discussions/2123.
The file is written for no purpose. The content is saved to a variable which is then send downstream to the client. The file is written in parallel but not used anywhere. So imho we can remove that part of the code. Opinions @magicbug @AndreasK79?

### DIFF
--- a/application/controllers/Kmlexport.php
+++ b/application/controllers/Kmlexport.php
@@ -90,17 +90,8 @@ class Kmlexport extends CI_Controller {
 		$output .= "</Document>";
 		$output .= "</kml>";
 
-        if (!file_exists('kml')) {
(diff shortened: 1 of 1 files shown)


magicbug:
If its not used then yes definitely doesn't need written to the system

phl0:
> If its not used then yes definitely doesn't need written to the system

At least not for the kml export/download function.

magicbug:
I have a feeling this was early days and for a few events I used the KML file to plot qsos live

AndreasK79:
So I guess we agree to remove then?

magicbug:
done

PR #2131, opened 2023-05-10
Qsl view resize

cats-shadow:
Initially, the QSL card image was displayed using a w-100 class and had to occupy the entire width of the QSO block. 
When I upload the scanned image of the QSL card in high resolution, the QSL image does not fit within the borders of the QSO block. This is true for normal themes as well as for "wide". 

Default theme:

Default_wide theme:

After adding a separate class to control the width of the QSL card image (w-qsl, in the corresponding bootstrap.min.css), it was possible to edit this parameter without affecting the other properties of the theme. In my case, the value of 55% of the block width turned out to be optimal.

Default theme:

Default_wide theme:


AndreasK79:
This does look better, but remove the 2 css files. The wide themes are not a part of Cloudlog.

You should be able to add the css to general.css. That way it will be valid for all themes.

phl0:
And this also contains changes made in https://github.com/magicbug/Cloudlog/pull/2130. May result in merge conflict.

magicbug (changes requested):
Most things have already been pointed out, my only comment is regarding over-stretching an image beyond what its capable of, but if everyone using wide is ok then that's great.

- application/language/russian/general_words_lang.php needs removed

cats-shadow:
> Most things have already been pointed out, my only comment is regarding over-stretching an image beyond what its capable of, but if everyone using wide is ok then that's great.
> 
>     * application/language/russian/general_words_lang.php needs removed

I hope all is OK now. Could you check, please?

cats-shadow:
> The problem is that you made this PR based on your changes made in the lang branch. Thats why this PR also contains the previous changes now leading to the merge conflict.

It is understandable. I hope all is OK now.

phl0:
The PR itself is ok now from a merge-conflict point of view. But in my case it wastes much of the vertical space now. 
Your PR:

Screenshot from 2023-05-12 12-01-03

Original Code:
 
Screenshot from 2023-05-12 12-00-43

AndreasK79:
@cats-shadow I agree with @phl0 that is wastes a lot of space.

When I tried to find a better solution a while back, I didn't find anything that worked well enough.
After doing some digging though, I managed to get this one: https://github.com/magicbug/Cloudlog/pull/2133

Tried that solution out and see if it works better for you.

AndreasK79:
> @cats-shadow I agree with @phl0 that is wastes a lot of space.
> 
> When I tried to find a better solution a while back, I didn't find anything that worked well enough. After doing some digging though, I managed to make this one: #2133
> 
> Try that solution out and see if it works better for you.

cats-shadow:
> The PR itself is ok now from a merge-conflict point of view. But in my case it wastes much of the vertical space now.

You are right. Of course it looks better if the QSL card image takes up the whole width of the QSL block.
But I haven't found a more versatile option for different browser window sizes. How it originally looked in my case is shown above.
Perhaps it would look better if the image was centred?

phl0:
> @cats-shadow I agree with @phl0 that is wastes a lot of space.
> 
> When I tried to find a better solution a while back, I didn't find anything that worked well enough. After doing some digging though, I managed to make this one: #2133
> 
> Try that solution out and see if it works better for you.

That solution by @AndreasK79 looks nice. I guess we should use that one rather than the fixed 55% size of this PR tbh.

AndreasK79:
I merged my solution, which is more responsive.

PR #2136, opened 2023-05-12
Eqsl card download

phl0:
As preparation for a mass-download script for eQSLs this adds a view function to the view_qso popup, a list of all eQSLs received in the QSO menu and a list of all not yet downloaded eQSL cards in usre menu. A mass-download function is still under dev.


phl0:
@m0urs Are you able to test the current state? We don't have to many eQSL subscribers in the dev team :)

m0urs:
1. View eQSL card in View QSO popup works fine
2.  QSO -> View eQSL works fine too
3. " a list of all not yet downloaded eQSL cards in usre menu": Seems to work fine as well for me

PR #2138, opened 2023-05-14
Better language support

AndreasK79:
1. Removed the mod I did to controller for loading menu translation. Put this in autoload instead.
2. Autoload for language helper -> shortens the translation string from $this->lang->line to lang Thanks @dgdavid and @leroydiazg for the idea.
3. Replaced all translation strings from using $this->lang->line to lang.
4. Also added language files to autoloader.


phl0:
Yeah. Seems to work well. Clicked various pages for test :) No issues so far (and if some arise we will fix them)

PR #2144, opened 2023-05-15
Update satellite_data.json with new LOTW names for TEVEL sats

ve1cwj:
Matches TQSL configuration file 11.24 released 2023-05-15


phl0:
This is only half of the party. We need to add a DB migration to also tename already logged QSOs and reset LotW sent so that they get uploaded with correct name. If needed I can supply code.

magicbug:
This isn't needed, just needs remapping in the lotw controller and a migration file to set lotw sent to no.

phl0:
> This isn't needed, just needs remapping in the lotw controller and a migration file to set lotw set to no.

This only fixes LotW upload but leaves ADIF exports for example with the old non-LotW-compliant SAT_NAME. See https://github.com/magicbug/Cloudlog/pull/2145#issuecomment-1549060168.

phl0:
> Good catch @phl0! So maybe two options: 1. Change the sat_name in satellite_data.json (and rename/reupload those already in the logbook) or 2. Apply the same remapping strategy used in the LOTW upload on ADIF export?

Sorry but I think the correct way would be:

1. Change the sat_name in satellite_data.json
2. Update DB and rename COL_SAT_NAME for TEVEL-x to TEVELx and set COL_QSL_LOTW_SENT to N
3. Apply the remapping in LotW Controller (if really needed, because we now have all SATs with LotW name in DB)

Just as in https://github.com/magicbug/Cloudlog/pull/1812.

I cannot see a reason why we should manually remap instead of correcting the DB content itself. It is less error-prone and just the same as we did for GREENCUBE -> IO-117. But maybe @magicbug can shed a light why he is of different opinion.

magicbug:
Greencube was issued OSCAR IO-117.. TEVEL-1 is still TEVEL-1 officially its just LOTW cant handle long names.

phl0:
> Greencube was issued OSCAR IO-117.. TEVEL-1 is still TEVEL-1 officially its just LOTW cant handle long names.

Which has been proven to be wrong by one of the tqsl devs. But that is another story. But still this would mean we export ADIFs with TEVEL-1 while uploading TEVEL1 to LotW ... So for the case that someone exports ADIF and uses tqsl he will fail.

magicbug:
Yep thats his fault not mine :)

phl0:
So we stay with ADIF exports differing from LotW exports?

magicbug:
Yes

PR #2145, opened 2023-05-16
Remap & Reupload Tevel Sat QSOs to LOTW

ve1cwj:
To match sat_names in TQSL configuration file 11.24 released 2023-05-15.


phl0:
This is targeting the wrong branch. Should be PRred against dev :)

PR #2147, opened 2023-05-16
[Visual decorations] Small decoration for header of the log overview …

cats-shadow:
Added bold font decoration to the station logbook and station location names in the header. This makes the names more visible than before.
Before:

After:

### DIFF
--- a/application/views/view_log/index.php
+++ b/application/views/view_log/index.php
@@ -2,7 +2,7 @@
 
 	<h2><?php echo lang('gen_hamradio_logbook'); ?></h2>
 	<?php if ($results) { ?>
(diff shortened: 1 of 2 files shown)


magicbug:
I've no issue with visuals, but would this be better handled in CSS, would probably require either a id or a class for each

cats-shadow:
> I've no issue with visuals, but would this be better handled in CSS, would probably require either a id or a class for each
> 
> Happy to be told I'm wrong.

Done! :)
CSS classes (station_logbook_decoration, station_location_decoration) added to general.css.

magicbug:
This is far to overbearing and not keeping in with the overall design of Cloudlog 

image

AndreasK79:
@magicbug and what is too overbearing? The increase of font size? Would have been nice to get an exact explanation and a chance to correct.

PR #2148, opened 2023-05-17
[Translation][Russian] Statistics (partially), User, Options views

cats-shadow:


AndreasK79:
@cats-shadow great work with the translation!

PR #2154, opened 2023-05-21
German translations

phl0:
add German translations for recent changes.


phl0:
wrong bnrach. closing

PR #2155, opened 2023-05-21
German translations

phl0:


phl0:
Ready to go here for German.

PR #2158, opened 2023-05-22
eQSL bulk download function

phl0:
This adds a function to download all eQSL cards that have not been synced to Cloudlog yet.

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -182,16 +182,16 @@ class eqsl extends CI_Controller {
 		$status = "";
 		
 		// begin script
-		$ch = curl_init(); 
+		$ch = curl_init();
 
 		// basic curl options for all requests
-		curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1); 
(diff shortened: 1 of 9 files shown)


magicbug:
This doesn't actually work when trying to bulk import 

Type: Error

Message: Call to undefined function str_contains()

Filename: /home/twom0sql/domains/logbooks.2m0sql.com/public_html/application/controllers/Eqsl.php

Line Number: 568

phl0:
> This doesn't actually work when trying to bulk import
> 
> An uncaught Exception was encountered Type: Error
> 
> Message: Call to undefined function str_contains()
> 
> Filename: /home/twom0sql/domains/logbooks.2m0sql.com/public_html/application/controllers/Eqsl.php
> 
> Line Number: 568

Which PHP version pls?

magicbug:
I sit on the minimum recommended which is 7.4.33

phl0:
> I sit on the minimum recommended which is 7.4.33

Then this should work: https://github.com/magicbug/Cloudlog/pull/2159

PR #2170, opened 2023-05-27
Make contest table use padding for SRX and STX

phl0:
Use a little padding for SRX and STX.

Screenshot from 2023-05-27 14-32-06


magicbug:
👍🏻 

@phl0 if you have 5mins in this section could we have some top padding above the search box so its not tight to the grey banner

phl0:
> 👍🏻
> 
> @phl0 if you have 5mins in this section could we have some top padding above the search box so its not tight to the grey banner

Likethis?

PR #2172, opened 2023-05-27
Lotw information

AndreasK79:
This PR makes LoTW user information more visible in the advanced logbook:
image

A search has been added to find unconfirmed LoTW QSOs where QSO partner has uploaded after QSO date:
image

Information about LoTW user and last upload is displayed in the QSO popup:
image

Need to run http://cloudlog-url/index.php/update/lotw_users to populate table. This can also be run as a cron job.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 122;
+$config['migration_version'] = 123;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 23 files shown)


phl0:
Working fine here. Really like this feature. Though LoTW csv file seems to be updated rather infrequently. :(

magicbug:
Looks good, agree with Flo the lotw csv is updated not enough by them!

phl0:
> Looks good, agree with Flo the lotw csv is updated not enough by them!

I am watching the file since Thursday and no change. Should we add a disclaimer somewhere? But where?

magicbug:
In the debug section we could say when the file was last updated?

PR #2180, opened 2023-05-31
Sort eQSL cards in desc order

phl0:
Think its nicer to list eQsl cards in descending order.


magicbug:
Ha yes was annoying me scrolling to the bottom lol

PR #2186, opened 2023-06-02
Fix qsl marking

phl0:
Using the mark QSL function in Logbook -> Overview does mark eQSL as received using JS. That is because the code does rely on counting columns which does not make sense as the number can vary (i.e. works correct if eQSL is disabled). We fix this by using a fixed ID.

https://github.com/magicbug/Cloudlog/assets/7112907/698caf03-4bf7-4530-b3d7-5ac54d9bf587


phl0:
Wrong branch. Closing here.

PR #2193, opened 2023-06-12
Fix the OQRS System

DeveloperNeon:
I made some quick modifications to the OQRS system so it works properly. Below you can find a sumamry of what I did.

application/views/oqrs/index.php
-- 
Move a div so it doesnt get deleted when another div is reset.

assets/js/sections/oqrs.js
--
Set the station id so it cant be deleted when the field is deleted, and move some QSO arrays around so they can be referenced from anywhere in the function

- KD9NSC


AndreasK79:
@magicbug I need to take a look. I was so sure things were working, so I don't know how this was missed. Anyway, it will get fixed.

@DeveloperNeon thanks. This is very much appreciated. However, it is only a partial fix. 

1. The callsign dropdown should remain when displaying station info. This is when search is not grouped.
2. Grouped searched shows result double. If I remember correct, the grouped search will add some div's with id's. Guess that's the reason.

PR's must also be based on the dev branch.

I will push a fix, just need to dig into the code again.

phl0:
I did some research here. The OQRS un-grouped search (i.e. based on station location) broke with commit https://github.com/magicbug/Cloudlog/commit/932171c51eae48d601f865c25bdf07c18c031533. I am not sure why that div was moved around. Still in research phase with @AndreasK79 :)

magicbug:
@AndreasK79 Yep thought I'd leave this one to you to triple check :-) feel free to merge it as and when you feel ready to

DeveloperNeon:
> @magicbug I need to take a look. I was so sure things were working, so I don't know how this was missed. Anyway, it will get fixed.
> 
> @DeveloperNeon thanks. This is very much appreciated. However, it is only a partial fix.
> 
> 1. The callsign dropdown should remain when displaying station info. This is when search is not grouped.
> 2. Grouped searched shows result double. If I remember correct, the grouped search will add some div's with id's. Guess that's the reason.
> 
> PR's must also be based on the dev branch.
> 
> I will push a fix, just need to dig into the code again.

Ok, cool. I just wanted to make sure you were all aware, and helped out where I could. If you dont mind, whenever you get a fix, could you please just reply to this PR?

Thanks.

AndreasK79:
@DeveloperNeon all help is welcome. I re-used the js changes you did. Had to make some changes to the html. Still a few bugs to fix, but will let you know when it's done.

AndreasK79:
@DeveloperNeon I tagged you in the other PR. If you are able to test it out, please do. I'm closing this one.

PR #2195, opened 2023-06-12
Fixes broken OQRS

AndreasK79:
It seems I managed to break OQRS at one point.

Thanks to @DeveloperNeon for letting us know about it, and for suggesting a fix. Parts of this fix comes from his PR.

In my test, this should fix the problems, but always good to have others verify.

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -14,7 +14,7 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 */
 
 $config['app_name'] = "Cloudlog";
-$config['app_version'] = "2.4.2";
+$config['app_version'] = "2.4.3";
 $config['directory'] = "logbook";
 
 $config['callbook'] = "hamqth"; // Options are hamqth or qrz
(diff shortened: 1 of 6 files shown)


AndreasK79:
@phl0 thanks for helping me test this PR at an early stage. Please verify that this works for you now.

DeveloperNeon:
I have gone ahead and tried it, and it doesnt seem to be submitting the request. Instead of sending me to the page that says the request has been submitted successfully, it just sends me back to the select station page, with the following url: `https://kd9nsc.com/log/index.php/oqrs?qslroute=D&message=ASD&mode=club%40odarc.org`

AndreasK79:
@DeveloperNeon thanks for the feedback. I had enter implemented at some point, but will take a look.

AndreasK79:
@DeveloperNeon I tried OQRS with enter, and for me it does work.
Could you specify a bit more where enter is not working?
I tried grouped search, and regular search. Both seems to work fine for me.

DeveloperNeon:
> @DeveloperNeon I tried OQRS with enter, and for me it does work. Could you specify a bit more where enter is not working? I tried grouped search, and regular search. Both seems to work fine for me.

@AndreasK79 
When I hit enter in the email field after entering my email, I get sent to the following link with no success message: `https://kd9nsc.com/log/index.php/oqrs?qslroute=B&message=test&mode=club%40odarc.org`. Nothing shows up in the OQRS Requests page as well.

AndreasK79:
@DeveloperNeon thanks. I guess I tabbed through the form instead of using enter. I've added so that enter works from the email input.

PR #2203, opened 2023-06-15
Check that station_id belongs to API key owner

AndreasK79:


AndreasK79:
Tested working by @phl0 and me.

PR #2204, opened 2023-06-16
[Email] Added email address option

AndreasK79:
Now must define email address in options.

@magicbug should we define name too? They are hardcoded right now.
Forgotten password: "Cloudlog"
OQRS: "Cloudlog OQRS"


magicbug:
Hmm, hard-coded doesn't seem bad for a name, I mean both those look good and make it identifiable but could just have it so it's uniform everywhere.

AndreasK79:
I guess since we set the subject, the sender name could be what you specify in options. I'll add another option.

magicbug:
Yep I can live with that :-)

AndreasK79:
@magicbug yep, think this should be it then.

phl0:
Nice. Works as expected here. Go for merge :)

PR #2205, opened 2023-06-16
Update FINNISH lang-files

oh1jw:
Updated some fields and created new files based on English language files.


magicbug:
Merged into dev branch will be in master soon

PR #2213, opened 2023-06-19
Fix LoTW users for activated grids

phl0:


phl0:
Same as previous PR but some other occurrences in the code.

PR #2216, opened 2023-06-20
added distance as choosable column

Werzi2001:
Based on the new stored distance I added it as a choosable column. I noticed that the distance stored is always in kilometers and therefore I added the fixed suffix "km". It might be good to do that based on the user setting (miles, nautic miles, kilometers). Is there a method already to convert between these units?


phl0:
> I noticed that the distance stored is always in kilometers and therefore I added the fixed suffix "km". It might be good to do that based on the user setting (miles, nautic miles, kilometers). Is there a method already to convert between these units?

That is because ADIF spec says the value is to be store in km. See http://adif.org.uk/314/ADIF_314.htm#QSO_Field_DISTANCE. If we want to show it in different units we have to convert upon display.

Werzi2001:
> If we want to show it in different units we have to convert upon display.

Yes that's what I meant. Do we want that? And if so is there a conversion method already? Or are you fine with the output being always in km?

AndreasK79:
Tested ok. Merging.

PR #2218, opened 2023-06-21
Calculate distance based on grid

phl0:
As qrz.com and hamqth.com do not provide distance via API we have to calculate it on our own.


AndreasK79:
This was a bug I found this morning. Merge time.

PR #2220, opened 2023-06-21
[Gridmap] Added gridsquare map with filtering

AndreasK79:
image
Can now filter gridsquare map on band/mode/sats/confirmation.

image
Added a legend with numbers.

image
And a separate menu item.


phl0:
Voting for this option. We could think about dropping the old one in favor of this.

magicbug:
Ageed seems to make sense that this is a replacement seems to do everything better

phl0:
I guess we would need a(nother) solution for the public maps. Just keep the existing map for public and the gridmap code for logged-in users?

AndreasK79:
So just remove the old menuitem and merge?

magicbug:
Yep

PR #2226, opened 2023-06-27
[Advanced Logbook] Added links for QRZ and HamQTH after callsign

AndreasK79:
Added clickable links for QRZ and HamQTH after the callsign in the Advanced Logbook:
image


phl0:
Yeah. Looks nice and works as expected on my end.

PR #2227, opened 2023-06-28
[Avanced Logbook] Can print labels

AndreasK79:
Added a few more buttons to the Advanced Logbook:
image

Added collapse for filters and actions:
image

image


phl0:
Works ok on my dev system.

PR #2232, opened 2023-06-30
Prevent .git being served by Webserver

int2001:
Securityissue:
Prevent .git-directory from being served/exposed by webserver


phl0:
Good find. This is totally needed for people running the instance directly of github checkouts.

PR #2233, opened 2023-06-30
[Advanced logbook] Added more filters and lookup

AndreasK79:
Added an extra filter row.
QSL / LoTW / eQSL are now on their own line:
image

Clickable links are added to DXCC, CQ Zone and IOTA to display the lookup table when you click on them.


phl0:
Yeha. I like the more advanced filters and especially the popups for DXCC/CQ/IOTA :) Very handy.

PR #2235, opened 2023-07-01
LOTW QSO end date issue

hugosilvaparagramanet:
When importing a LOTW certificate the "QSO End Date" was being stored in the database with time ```00:00:00```, but in fact the validity is a date so it should work until ```23:59:59```. This causes QSOs made in the last day that the certificate covers to not be uploaded to LOTW.
The fix is simply to add ```23:59:59``` to the data when importing the certificate.
I have not created a fix to the existing data since this is very close to an edge case and only people that ask for certificates with very tight date limits should be affected.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -21,7 +21,7 @@ $config['migration_enabled'] = TRUE;
 | be upgraded / downgraded to.
 |
 */
-$config['migration_version'] = 124;
+$config['migration_version'] = 125;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


phl0:
Good point. Though it is an edge case we can fix it. I will have a look.

hugosilvaparagramanet:
Updated with migration from @phl0 to correct existing data in ```lotw_certs``` table

phl0:
Yeah. Worked on my end. Maybe @AndreasK79 can cross-check the DB stuff with his WAMP instance?

AndreasK79:
Tested working here :)

PR #2237, opened 2023-07-02
[Gridsquare Map] Added display of coordinates, gridsquare, distance and bearing when mouseover

AndreasK79:
image


magicbug:
love it

AndreasK79:
Let's merge it then 😊

Issue #2238, opened 2023-07-02
eQSL- / LotW-Import doesn't care about station_id


int2001:
**To Reproduce**

Prequisites:
* Working and logging with special Event Call (let's call it "DL0TESTING"). Station-Profile is set up to DL0TESTING
* You are doing a qso with another station (let's call it "DL0FOO") and logging it.
* DL0FOO asks you to log also your personal call. You accept and log it in another Cloudlogaccount also for your personal Call (let's call it DL1BAR)
* Now two qso exists. One DL0TESTING <-> DL0FOO, and one DL1BAR <-> DL0FOO. Same time_off, same grid, etc. - distinguishable via station_id

The Bug:
* You're getting an eQSL (and/or a LotW-Confirmation - that's the second part of the Bug) but only for the QSO between DL1BAR and DL0FOO.
* Cloudlog imports this one and doesn't take care of the station_id. So it updates BOTH QSOs in the Database. In other Words: Even if the Account/Station_id DL0TESTING doesn't have any eQSL or LotW-Credentials, the QSO will also be updated for DL0TESTING. This happens with eQSL and LotW

int2001:
Need help, perhaps from @AndreasK79 (?) or @phl0 (?), with the eQSL-Part.

eQSLs-confirmations aren't coming with a gridsquare. So one needs to iterate through the (valid) station_id of the origin-call given for the confirmation. The following functions need to be adjusted for that:
- find_correct_station_id (actually returns only one row) https://github.com/magicbug/Cloudlog/blob/06656bfef81fa9234f2e296950e48f514ec1dec1/application/models/Logbook_model.php#L3773
- of course import itself: https://github.com/magicbug/Cloudlog/blob/06656bfef81fa9234f2e296950e48f514ec1dec1/application/libraries/EqslImporter.php#L134
- import_check: https://github.com/magicbug/Cloudlog/blob/06656bfef81fa9234f2e296950e48f514ec1dec1/application/models/Logbook_model.php#L2472
- eqsl_dupe_check: https://github.com/magicbug/Cloudlog/blob/06656bfef81fa9234f2e296950e48f514ec1dec1/application/models/Eqslmethods_model.php#L176
- eqsl_update: https://github.com/magicbug/Cloudlog/blob/06656bfef81fa9234f2e296950e48f514ec1dec1/application/models/Eqslmethods_model.php#L159

Idea is to update every QSO, which belongs to station_ids (plural) where the ownercall is listed. 
[...]

PR #2239, opened 2023-07-02
LotW-Import now takes care of station_id

int2001:
See Issue #2238 - But only the LotW-Thing.
This Patch resolves, that the LotW-Import now takes care of the station_id the qso belongs to


int2001:
Will do a new PR for the whole package to fix this issue

PR #2241, opened 2023-07-03
Use monospace font to "fix" text and prevent jumping

phl0:
May I propose to use a monospace font here and apply some padding with spaces? This prevents the text from jumping left and right while hovering the map.

@AndreasK79 Your opinion on this is highly appreciated.

### DIFF
--- a/application/views/gridmap/index.php
+++ b/application/views/gridmap/index.php
@@ -31,6 +31,7 @@
     align-items: stretch;
 }
 .cohidden {
+    font-family: monospace;
     display:none;
 }
 </style>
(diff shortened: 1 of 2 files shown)


AndreasK79:
@phl0 I like the idea of the text not jumping around, but I find the font misplaced here. I tried to avoid a fixed width in the css (specify width in px), but not sure if we have any other solution without monospace font.

You can set col-md-2 etc, but then you get a lot of air around the columns.

@magicbug any tips here?

magicbug:
setting columns inside a row col-md-2  is probably best then give it an ID and change some of the padding settings probably the only real way of fixing it

phl0:
This will not solve the issue with jumping text. I will keep the patch for myself. Maybe I can discover a better method without monospace font later.

PR #2245, opened 2023-07-04
Load gridmap on page load

phl0:
Show gridmap on page load using default form settings.


AndreasK79:
I would very much like this being optional. 99% of the time, I'm not interested in seeing the default map.

int2001:
is working for me.
Tested user-config and gridview.

phl0:
Added text for all bands. Tnx

AndreasK79:
Tested working. And with default values set, nothing to complain about. Merge time.

PR #2249, opened 2023-07-04
Fixes #2238

### LINKED ISSUE #2238: eQSL- / LotW-Import doesn't care about station_id

int2001:
Fixes Issue https://github.com/magicbug/Cloudlog/issues/2238

Implements a dropdown for Stationlocations when importing eQSL manually
eQSL: Checks if Station_callsign of the QSO matches with Call provided by dropdown (and User)
LotW: Checks if station_id of the QSO matches with Call and grid provided by LotW-Download
Some background-checks (mantra: Never trust userinput)

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -15,6 +15,7 @@ class eqsl extends CI_Controller {
         $folder_name = "images/eqsl_card_images";
         $data['storage_used'] = $this->sizeFormat($this->folderSize($folder_name));
 
+
         // Render Page
         $data['page_title'] = "eQSL Cards";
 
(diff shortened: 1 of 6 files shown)


AndreasK79:
@int2001 from what I can tell, the code looks ok. Can't really test the import stuff myself, but I guess @phl0 can report back.

magicbug:
Agree looking at the code it seems OK, I'd want to triple-check that lotw manual changes has zero effect on the cronjob

int2001:
@magicbug : Checked the eQSL-Sync-Cron (i think you meant the eQSL-Downloader, there's no LotW-Cron for download) a few minutes ago.
Works well. Grabbed an eQSL-Card and updated the QSO here.

Upload also works well - i didn't changed anything regarding the upload (neither LotW nor eQSL).
But lets wait for @phl0 and @AndreasK79 - triple check, you know ;)

Issue #2198, opened 2023-06-14
Stop syncing to qrz.com if Key is expired


int2001:
if one enters a qrz.com-api-key, cloudlog syncs to qrz.com (Great Feature!)
But if the qrz-subscription expires, cloudlog will try the sync **for each (unsynced) QSO** every time.

Suggestion: If the first Request failed with the following Errormessage, then stop trying to upload the rest of the QSOs.

`--> QRZ upload failed with the following message: STATUS=AUTH&REASON=user does not have a valid QRZ subscription to use this function`

Saves a lot of traffic.

magicbug:
This seems like a good reason, if that's returned simple solution might be to delete the API key

int2001:
Yes, one can do that. But the "normal User" isn't aware of that. He's confused about "no sync" while the server is establishing more and more connections.
Extreme situation: Fresh Account. User imports everything via ADIF-Import und has a key. But key is invalid.

So CL tries every croncycle to sync (happened here: More than 200 MB every hour)

int2001:
Fix with PR: https://github.com/magicbug/Cloudlog/pull/2252

PR #2252, opened 2023-07-05
Qrz failure fixing - Fixes https://github.com/magicbug/Cloudlog/issues/2198

int2001:
Fixes the following:
- When uploading to qrz with an invalid or expired key, Cloudlog stops immediatly uploading for **that** station, so only one QSO fails, not thousands (in the bad case)
- When uploading to qrz and qrz tells us, that is is a dupe (already uploaded / by hand / whatever), Cloudlog simply marks the QSO as uploaded. So there are no infinity retries on that. This is also applied on live-logging.

Tested for mass_upload (via cron), Frontend-Upload (QRZ Logbook -> Upload Logbook) and Livelogging


magicbug:
Thanks for doing this

PR #2254, opened 2023-07-05
Added MODE to matching for LotW/eQSL Confirmations

int2001:
Old behaviour:
- QSO logged with Mode "FT8"
- eQSL/LotW confirmation comes with Mode "SSB"
-> QSO will be confirmed in the Cloudlog-database, but when trying to fetch the eQSL-Image it fails, because eQSL tells us "no QSO found with mode...". Furthermore it is no real match at LotW (for awards)

New behaviour:
- QSO logged with Mode "FT8"
- eQSL/LotW confirmation comes with Mode "SSB"
-> QSO will **not** be confirmed in the Cloudlog-database. Error (when syncing via Frontend) is shown as "QSO not found"

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -497,7 +497,7 @@ class Lotw extends CI_Controller {
 					$record['qsl_rcvd'] = $config['lotw_rcvd_mark'];
 				}
 
-				$status = $this->logbook_model->import_check($time_on, $record['call'], $record['band'],$record['station_callsign']);
+				$status = $this->logbook_model->import_check($time_on, $record['call'], $record['band'], $record['mode'], $record['station_callsign']);
 				$skipNewQso = $this->input->post('importMissing'); // If import missing was checked
 
 				if($status[0] == "No Match" && $skipNewQso != NULL) {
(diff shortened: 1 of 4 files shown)


int2001:
tested this - with the help of @AndreasK79 (tu) - against LotW and eQSL.
In both servies col_mode is the master. There's no qsl with the submode in the mode-field. So checking mode against mode is the right way.

Crons / Imports are still working (double-checked it)

magicbug:
Finding now that LoTW no longer matches it says its matched but the record isn't updated

image

magicbug:
OK found the issue  with find_correct_station_id running a like on gridsquare wont work for example in my account I had IO87IP and IO87JP locations and the like just picked off the first one.

```php
    function find_correct_station_id($station_callsign, $my_gridsquare) {
        $sql = 'select station_id from station_profile
            where station_callsign = "' . $station_callsign . '" and station_gridsquare = "' .$my_gridsquare. '"';

        $query = $this->db->query($sql);

        $result = $query->row();

        if ($result) {
            return $result->station_id;
        }
        else {
            return null;
        }
    }
```

int2001:
interesting. as far as i consider, i didn't change anything regarding the grid. Only added MODE-Check on top

magicbug:
Possibly happened in another recent commit :-)

PR #2259, opened 2023-07-06
Minor Fix: supress fetching empty eqsl-image at qso-detail-view

int2001:
Fixes following bug: 

if there's no eqsl-image at the database, qsl-detail-view will try to fetch /images/eqsl_card_images/ which produces a 403 and spams webservers-error-log


phl0:
Good find. Go merge it :-)

PR #2260, opened 2023-07-07
This one adds HRDLog-Support to Cloudlog

int2001:
Includes and tested so far (plz double-Check @phl0 / @AndreasK79 / @magicbug ):

* Add/edit station profiles with new HRDLog fields.
* Migration script
* Bulkupload
* Bulkmark
* Live upload
* Modify and update QSOs within Cloudlog and syncing them to HRDLog

Update: forgot to mention, that there's also a cronjob


int2001:
For those who are interested, here's the API [Documenation](http://www.iw1qlh.net/projects/hrdlog/HRDLognet_2.pdf)

application/controllers/Hrdlog.php
```diff
@@ -0,0 +1,176 @@
...
+
+/*
+	Controller to interact with the hrdlog.net API
+*/
+
+class Hrdlog extends CI_Controller {
+
+    // Show frontend if there is one
+    public function index() {
+        $this->config->load('config');
```
magicbug, line 11: Shouldn't need to load config here.

magicbug:
Oh this has been asked for a number of times and I step over it, I don't actually use HRDLog but I'll be able to test that it doesn't break anything it shouldn't :-)

PR #2262, opened 2023-07-07
[Advanced Logbook] Added SOTA, POTA, WWFF searches and refs

AndreasK79:
Added filters for SOTA, POTA and WWFF.
image

POTA and WWFF are also display in REFS now. Previously only SOTA was displayed.

@phl0 here you go.

### DIFF
--- a/application/controllers/Logbookadvanced.php
+++ b/application/controllers/Logbookadvanced.php
@@ -1,4 +1,4 @@
-<?php 
+<?php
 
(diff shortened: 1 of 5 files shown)


phl0:
> @phl0 here you go.

Will look after it as soon as I finished the gridmap user options. Tnx.

phl0:
Just pulled and tested. Works like a charm and lets one easily find QSOs with same ref (SOTA/POTA/WWFF). Go for merge!

PR #2270, opened 2023-07-09
Fixes CAT-Update when not logged in

int2001:
Old behaviour:
- You've setup CAT
- You've an open Browser-Tab, lets call it "1", (with CAT-enabled) to log a qso
- Everything works great
- You've another Browser-Tab or Window, where you log out
- Browsertab "1" now produces ongoing ajax-calls to radio/json/[radio_id]
- At controllers/Radio.php it isn't checked if User is logged in. Some SQLs are fired...

New behaviour:
- You've setup CAT
- You've an open Browser-Tab, lets call it "1", (with CAT-enabled) to log a qso
- Everything works great
- You've another Browser-Tab or Window, where you log out
- Browsertab "1" now produces ongoing ajax-calls to radio/json/[radio_id]
- Controllers/Radio.php is now checking if User is logged in and fills json with an error-message (this can also be used in future for other error cases)
- At Browsertab "1" the info is shown, that the user isn't login, with a link to the mainpage.
- If User logs in back again (on another Tab), the errormessage disappears and everything works as expected.

Also did formatting for the modified functions to fit the codestyle-guide

### DIFF
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -78,82 +78,95 @@
 
 	}
 
-	function json($id)
-	{
+	function json($id) {
 
-		header('Content-Type: application/json');
+		$this->load->model('user_model');
 
-		$this->load->model('cat');
+		// Check if users logged in
 
-		$query = $this->cat->radio_status($id);
+		if($this->user_model->validate_session() == 0) {
+			// user is not logged in
+			// Return Json data
+			header('Content-Type: application/json');
+			echo json_encode(array(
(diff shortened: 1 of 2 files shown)


int2001:
plz review this fix b4 merging
tested it by myself. at 1st sight everything is ok.

phl0:
Works as expected and hides the SQL errors. Go for merge :)

AndreasK79:
Tested and error is gone in the log. Instead it is replaced with this one in the gui:
image

PR #2276, opened 2023-07-12
Base LoTW matches on time and station_callsign

phl0:
Just discovered an issue with LoTW reports. I have two station_profiles with exact same grid and station_callsign (one has a POTA reference and the other a GMA reference) but both are in the same grid.
Now the LoTW import fails because `find_correct_station_id($station_callsign, $my_gridsquare)` would find both of them and always only use the first.

I did some research with @int2001 and @AndreasK79 and we came to the conclusion that matches should be based on station_callsign and time_on as those two values are delivered by lotwreport.adi upon syncing. lotwreport contains the exact time on you logged even if the QSO partner logged some minutes off (tested with @int2001) so we are sure that we can use the time for exact matches.


phl0:
`skipNewQso` aka `importMissing` has been removed as auto-creating QSO from LoTW matches should not occur imho. The code was hidden from the GUI anyway.

magicbug:
I agree with you @phl0 auto creating from QSO matches isn't a good idea, either the qso is in the log or it isnt

phl0:
Ok this one is go for merge then. Tested by @int2001.

int2001:
QSL!
testet this thing on my dev-env (imported DB from a multiuser-env).
couldn't find any bug.

PR #2277, opened 2023-07-12
Always update with LoTW confirmed grid

phl0:
As we are updating several infos like IOTA, state, county I guess we should also update grid from LoTW confirmations. The code introduced in https://github.com/magicbug/Cloudlog/commit/614452b07e83828ea5c9d27d0a9ea823b4188c87 obviously has a problem here.

I logged a QSO with @int2001 and put his grid wrong (e.g. JO31OO). LoTW confirmation comes in and contains the correct grid (JO30xx). The code would not update the QSO's grid because it is longer than 4 chars but not longer than 6 chars which leads to an SQL statement that does not give a match. So no update on the grid here.

Imho the LoTW confirmation is the most specific grid info here so we should just update the QSO with grid info from LoTW instead of trying to determine which is more specific. Also this would prevent updates if a user logged a wrong but more specific grid ...

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -2693,21 +2693,6 @@ class Logbook_model extends CI_Model {
       $this->db->where('date_format(COL_TIME_ON, \'%Y-%m-%d %H:%i\') = "'.$datetime.'"');
       $this->db->where('COL_CALL', $callsign);
       $this->db->where('COL_BAND', $band);
-	  $this->db->group_start();
-	  $this->db->where('COL_VUCC_GRIDS');
-	  $this->db->or_where('COL_VUCC_GRIDS', '');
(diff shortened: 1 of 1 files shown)


magicbug:
Yeah updating grid seems a sane thing todo, catches errors does this handle VUCC @phl0 or just the main gridsquare field.

magicbug:
Ha, i dont think it did.. OK shall we mege this?

phl0:
Yes please and I need to look at VUCC_GRIDS separately.

phl0:
> Yeah updating grid seems a sane thing todo, catches errors does this handle VUCC @phl0 or just the main gridsquare field.

See https://github.com/magicbug/Cloudlog/pull/2278. There is the implementation for VUCC_GRIDS.

PR #2280, opened 2023-07-13
Lotw per user / fixes lotw-download

int2001:
Old behaviour:
- LotW-Download only worked for one User (not for multiusers), regardless if GUI or cron-triggered.
- When a user triggered LotW-Upload via GUI, every (not uploaded) QSO (of every! user which had a lotw-cert) was uploaded. All results were shown to this user (even the results of other users)
- lotw_upload triggered by cron also exposed every information. (even to users which are not logged in / simply called the url)
- last_download lotw-QSL was taken from the whole database, not from user or station. so (together with the bug 1st mentioned here) some QSOs of some users were never getting the chance of LotW-Confirmations

New behaviour:
- LotW-Download works for current user (if logged in)
- LotW-Upload works for current user (if logged in)
- LotW-Download / Upload is for whole Database when triggered by cron/curl
- LotW-Download / Upload is not longer exposing informations, when triggered by cron/curl without session
- last_download lotw-QSL is now per user. so if you have not enabled your cron and every user syncs by pressing the sync/import button on the gui, the sync takes care of the last received lotw-qsl

pls doublecheck @AndreasK79 / @phl0 / @magicbug

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -355,8 +355,11 @@ class Lotw extends CI_Controller {
 			/*
 			|	Download QSO Matches from LoTW
 			*/
-			if ($this->user_model->authorize(2)) {	// Only be verbose if we have a session
+			if ($this->user_model->authorize(2)) {
 				echo "<br><br>";
+				$sync_user_id=$this->session->userdata('user_id');
+			} else {
+				$sync_user_id=null;
 			}
 			echo $this->lotw_download($sync_user_id);
 	}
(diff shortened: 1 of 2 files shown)


int2001:
Tested and passed done so far:
- Import LotW-Report manually (ADIF Import)
- Cron lotw-sync
- Single-User sync up & down (logged in user)

phl0 (approved):

phl0:
to be a little more verbose: My tests where based on a multiuser system, i.e. two user accounts with two different callsigns and LoTW certificates. cron syncs both whereas GUI only sync the logged-in user.

magicbug:
Worked on the dev box, lets merge into the branch and try it on my real log.. sure its fine there too :D

PR #2282, opened 2023-07-13
Public search

phl0:
This re-adds a public search box which can be enabled on a per-logbook basis. Result table is sortable. Awaiting feedback by @AndreasK79 and @int2001 before undrafting this.


AndreasK79:
Tested PR. Working as intended as far as I can see.
Enable/disable works. Displays calls correct, and gives message with nothing found when call not found.
Go for merge.

PR #2283, opened 2023-07-13
Adds LotW-Gradient to QSO-Page

int2001:
Contents:

- Adds the LotW-Info-Gradient, known from logbook and logbook-advances to the QSO Page
- Adds also tooltip where last-lotw-post is shown to that.
- Refactored Logic, so that this function is also using the database instead of the file

plz tst.


phl0:
Works as expected. I like it!

Issue #1505, opened 2022-06-29
[station_logbooks] Search bar not displayed on public page


m0urs:
I have set "Public Search Bar" to "enabled" in the Global Options and I would expect that the search bar is now visible on the public page. Unfortunately it does not seem to be the case. Is this a bug or did I get it wrong for what this option is good for?

[https://cloudlog.urspringer.de/index.php/visitor/dg3nab](https://cloudlog.urspringer.de/index.php/visitor/dg3nab)

I am also missing the possibility to scroll through the QSOs list on the public page. I think think was possible in the old version? Currently only a fixed numbers of QSOs are listed on the public page.

phl0:
@magicbug Should there be a public search bar in station_logbooks? I did dig through the code but could not find a hint that it was implemented or removed in station_logbooks. Or is that just that it has not yet been (re-)implemented for station_logbooks?

magicbug:
The answer is yes, reason its not is because I hadn't written the search function for the new public dashboard code :)

phl0:
We have a solution for that in https://github.com/magicbug/Cloudlog/pull/2282.

m0urs:
Does work for me.
However, I would suggest, that we can search for parts of a call e.g. to find "DG3NAB/P" also if we search for "DG3NAB". What  do you think?

phl0:
> Does work for me. However, I would suggest, that we can search for parts of a call e.g. to find "DG3NAB/P" also if we search for "DG3NAB". What do you think?

Good idea. Implemented with https://github.com/magicbug/Cloudlog/pull/2285.

m0urs:
> Good idea. Implemented with #2285.

This does not work for me? Could it be that a "%" as wildcard is needed after a LIKE statement in your SQL where clause?

phl0:
> This does not work for me? Could it be that a "%" as wildcard is needed after a LIKE statement in your SQL where clause?

It is the same search statement as in the "internal" search. Does it work there for you? On my machine:

![Screenshot from 2023-07-14 08-59-53](https://github.com/magicbug/Cloudlog/assets/7112907/f5a65c0c-d7c4-421f-90c3-3951a8f792ec)

The executed SQL statement is:

```
SELECT *                                                            
FROM `TABLE_HRD_CONTACTS_V01`                                                                            
JOIN `station_profile` ON `station_profile`.`station_id` = `TABLE_HRD_CONTACTS_V01`.`station_id`
[...]

PR #2285, opened 2023-07-14
Refactor public search

phl0:
Made a few adjustments on the public search:

- Make search a little less picky (as requested by @m0urs in https://github.com/magicbug/Cloudlog/issues/1505#issuecomment-1634926285)
- Added CSV export button to result table
- Fixed a JS error (i.e. not load the map for result page)
- Made search input field upper case only on focus (requested by @AndreasK79)
- Use search term for pre-setting the input field on the result page
- Added page title to result page
- Make search not execute if user searches for empty string

### DIFF
--- a/application/controllers/Visitor.php
+++ b/application/controllers/Visitor.php
@@ -467,7 +467,7 @@ class Visitor extends CI_Controller {
 	}
 
 	public function search() {
-		$callsign = $this->security->xss_clean($this->input->post('callsign'));
+		$callsign = trim($this->security->xss_clean($this->input->post('callsign')));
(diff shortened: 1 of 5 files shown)


int2001:
Checked:
- CSV export / works
- JS error is gone
- Searchfield case / works
- Search term stays there after doing a search
- Title is there in result page
- Search doesn't execute if user searches for empty string

Read2merge

int2001:
tested tooltip and checking for empty-input. 
Works.

int2001:
tested trimming: Works

ready 4 merge

PR #2286, opened 2023-07-14
Station call check on import

int2001:
Old behaviour:
- Chosing a station_profile at import ADIF
- Upload an ADIF with a different Call than that in the station_profile
--> inconsistent Data in Database

New behaviour:
- Chosing a station_profile at import ADIF
- Upload an ADIF with a different Call than that in the station_profile
--> importer skips Records where station_callsign differs from that one in the profile

### DIFF
--- a/application/controllers/Adif.php
+++ b/application/controllers/Adif.php
@@ -206,7 +206,7 @@ class adif extends CI_Controller {
 
 
 				$custom_errors .= $this->logbook_model->import($record, $this->input->post('station_profile'),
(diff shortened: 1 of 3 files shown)


AndreasK79:
Tested working with ADIF with another call. No QSOs were imported, and got an error message.

phl0:
Tested and works ok here:

Screenshot from 2023-07-14 09-44-34

int2001:
also removed the check for NULL in Api.php. The else will never be reached, because script dies earlier when station_profile_id is omitted

PR #2293, opened 2023-07-15
Set Mode to SSB when inverted SSB transponder, otherwise to LSB/USB

int2001:
Old behaviour:
- Mode is always set to SSB, when working SATs with LSB **or** USB down/uplink
- When working SAT with bent pipe (like QO100) Up/Down is only USB. Mode is also set to SSB then.

New behaviour:
- Checks if Downlink/Uplink differ and if they're LSB/USB. Sets Mode to SSB
- When working SAT with bent pipe (like QO100) Up/Down is only USB. in this case USB is set
- When working SAT with bent pipe Up/Down is only LSB. in this case LSB is set

plz chk


phl0:
Looks of for me and is working fine.

PR #2297, opened 2023-07-15
Fixes #2295 // Fill station_callsign when provided empty by adif

### LINKED ISSUE #2295: ADIF - If importing and station_callsign isn't provided check if OPERATOR is
If importing an ADIF file and station_callsign isn't provided check if OPERATOR is and use that instead thus suppressing unnecessary erroring.

photo_2023-07-15_10-58-52

Get rid of those errors! :)

Possibly might want to change the error to say that said tag was missing or incorrect rather than just wrong

Thanks G0ABI for reporting

int2001:
Populate station_callsign with chosen profile-call when ADIF has empty station_callsign.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -2754,7 +2754,7 @@ class Logbook_model extends CI_Model {
 	}
 
 	if (($station_id != 0) && ($record['station_callsign'] != $station_profile_call)) {	// Check if station_call from import matches profile ONLY when submitting via GUI.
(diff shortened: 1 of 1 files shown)


AndreasK79:
Tested importing without station callsign ok.
Tested importing with correct station callsign ok.
Tested importing with different station callsign gave an error.

Go for merge.

AndreasK79:
@magicbug this was also merged to master.

magicbug:
oops!

magicbug:
pushed to dev too, seems to be working ok however

magicbug:
I shouldn't touch Github when i feel pants :D

AndreasK79:
It can happen to anyone.

PR #2300, opened 2023-07-16
[PHP8.2] Updated CI files to work with PHP8.2

AndreasK79:
I changed to PHP 8.2.4 on my dev machine.

Only small changes in 6 files were needed. I have clicked through looking for errors, but would appreciate if others could test as well.


int2001:
Checked with PHP 8.1:

tested (with 8.1):
- eQSL import (curl and XML)
- LotW import (curl and XML)
- maps 
- DXCC-Updated
- Logbook / Logbook advanced

all working without issues with PHP 8.1.

phl0:
Tested and works ok on PHP7.4, PHP8.1 and PHP8.2 here

PR #2301, opened 2023-07-16
Dxcluster lookup

int2001:
Implemented simple DXCluster-Lookup.

See discussion and Video at https://github.com/magicbug/Cloudlog/discussions/2299

Per default it's disabled, you can enable it by putting the Base-URL to config.php (dxcluster_provider)
Needs small service which can be installed from here: https://github.com/int2001/DXClusterAPI

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -598,16 +598,3 @@ $config['rewrite_short_tags'] = FALSE;
 | Array:		array('10.0.1.200', '192.168.5.0/24')
 */
 $config['proxy_ips'] = '';
-
-/*
-|--------------------------------------------------------------------------
-| DXCluster-API Base-URL
-|--------------------------------------------------------------------------
-|
-| If you have access to a DXCluster-API ( see https://github.com/int2001/DXClusterAPI for running your own )
-| put the base-URL for that API here
-|
-| Example:			'https://log.jo30.de/dxcache'
-| Leave blank to disable:	''
-*/
-$config['dxcluster_provider']='';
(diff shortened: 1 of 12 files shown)


AndreasK79:
I tried the PR by copying over the changes (I am on PHP8.2.4, so can't really change branch right now). QRG lookup for call and triggering lookup in Cloudlog worked as expected.

I don't know why those extra commits are listed here, so not sure if that will do anything if merged.

phl0:
Tested and works great. I like the idea. 
The extra commits probably because the dxcluster_lookup branch was updated to latest dev state by merging dev into dxcluster_lookup with https://github.com/magicbug/Cloudlog/pull/2301/commits/ba5a97b5dd7a5138f97a60a06466cc33b230e4bb.

int2001:
@magicbug what's your opinion about this one? Need some guidance from you.
I would like to know if we should pursue this further (bandmap is in the works), or if it's a silly idea

magicbug:
Definitely not a silly idea, I'm constantly asked for dx cluster integration think bandmap is the key one and perhaps just a view of the last x spots coming down.

Sorry not been to active last couple of days

magicbug:
I'd suggest we hold out, i'll check the nodejs script and try get something deployed so people have something to connect it too.

Another thing is that dxcluster_provider needs to be in the options table in the database rather than the config file so it can be changed in the global options area in the interface trying to get away from static configs.

int2001:
so i refactor that to options-table and Admin-UI and make a new PR?
no problem.

magicbug:
Yep, you can keep it as this PR if you want and just keep adding to it.

int2001:
@magicbug before i place my own instance again into the languagefiles as an example. Do you already have an URL for me, which i can put into the hints?

int2001:
@magicbug Moved that to admin-page. Can you check?
If ok, you can undraft and pull 😊

phl0:
Updated and tested the current state of the code. Works fine for me and is very useful at times. 👍

PR #2303, opened 2023-07-16
[Distances Worked] When clicking bars, QSOs are displayed in a popup

AndreasK79:
Now that distances are stored in the database, we can easily select on these QSOs.
I added so that the bars in Distances Worked can be clicked on to display all QSOs with the clicked distance.


int2001:
Working great.
No JS-Errors or other bugs.
Tested:
* SAT-Distances
* HF-Distances
* Switching around (SAT Mode / Distance-Window / etc.)
* Details

PR #2309, opened 2023-07-18
Display Transmit Power in capital letter (W), added space

rtoenniges:
Display the Unit Watt in capital letter 'W' with space. 
Like: "5 W"


phl0:
I cannot add a review/approval here. Just checked and found this be the only occurence where TX power is displayed. So ok to merge I think.

PR #2314, opened 2023-07-22
Adjusted caption of contest-session-reset

int2001:
Changed caption on the contest page.

The old caption was a bit misleading.

If you forget to "reset" and enter contest mode, you don't see the last QSOs from the previous contest (which can give you the hint to reset).

You have to make the first QSO. After adding the old QSOs are visible and you are annoyed not to have reset before.

So "Start new contest Session" instead of "Reset contest Session" will be better.


magicbug:
Agree makes way more sense

PR #2315, opened 2023-07-22
Override 'environment' with environment variable 

ashleyprimo:
### **Brief Description**:
This is a simple change that allows the default `ENVIRONMENT` (of `development`) to be overridden using an [environment variable](https://www.php.net/manual/en/reserved.variables.environment.php).

### **Why?**:
This is specifically very useful for two cases (that I can think of at-least):
1. **Container Runtimes** (such as docker, crio etc). It's not possible/and or practical to modify the 'index.php' within a container, it's furthermore not a persisted change. If you did persist it, it would complicate tasks such as updates. 
2. **Updating in general**. It's possible that the 'index.php' may need to be updated in the future, changes to this file may be overwritten. That said, I am new to this project, I have not had time to review how updates are done (etc).

All the best
Ashley (M7JNR)

### DIFF
--- a/index.php
+++ b/index.php
@@ -53,9 +53,7 @@
  *
  * NOTE: If you change these, also change the error_reporting() code below
  */
-	#define('ENVIRONMENT', isset($_SERVER['CI_ENV']) ? $_SERVER['CI_ENV'] : 'development');
-	define('ENVIRONMENT', 'development');
-
+        define('ENVIRONMENT', isset($_ENV["ENVIRONMENT"]) ? $_ENV["ENVIRONMENT"] : 'development');
 /*
  *---------------------------------------------------------------
  * ERROR REPORTING


magicbug:
This would sadly cause lots of issues for others that's the biggest worry

ashleyprimo:
@magicbug could you expand, this should not impact any other users - if the environment variable is not set, it'll stick to the default 'development'

phl0:
I will try to explain as I just did some research on this. For some reason (that I don't know) the original line defining the environment has been commented out in the past and environment is hard-coded to development. Some instances run for users are automatically updated using the update_cloudlog.sh script in this repo (at least I think this is how it works). But their environments are (manually?) set to production. Now whenever we change this line the update script will fail pulling from github because of a merge error. It simply expects the line to ne set to development and of not will fail with a merge error. This (probably) is the rrason for Closing this PR (and others in the psst).

I just did an experiment to re-set the env further down the code which would prevent a merge error on updating. But this fails as PHP dies not allow changing or deleting constants. So we are a little stuck here. 

Maybe someone has a clever idea how to solve this without merge errors?

ashleyprimo:
Thanks for that explanation, makes a lot of sense - I can see why this change would cause some users issues. I will look at a solution that considers this issue and report back :)

PR #2319, opened 2023-07-24
Cut off an s in case we have only one QSO

phl0:


magicbug:
This is the kind of small charge that I love. the details 👍🏻 thanks Flo

PR #2324, opened 2023-07-25
Fixed Submode/Mode issue when checking if worked

int2001:
Old Behaviour:

Consider II1WRTC worked on 20m with logged Mode "USB"

When trying to log him again, DXCC, Grid and Call is shown red (Never worked). This results in comparing the Submode (USB) to the Database (which holds only the Mainmode - in this case "SSB")

New behaviour:
Mode is always converted to mainmode before checking. if it already is mainmode - no problem.
So in the above case II1WRTC with grid and dxcc will be shown as "worked" when Mode ist "USB", "LSB" or "SSB".


phl0:
Makes sense imho.

PR #2326, opened 2023-07-26
Bandmap allbands

int2001:
Added "All"-Bands to Bandmap and Bandlist.
Refactored sorting at list


int2001:
Withdrawing this PR, since it collides with https://github.com/magicbug/Cloudlog/pull/2325
Will re-PR it, when https://github.com/magicbug/Cloudlog/pull/2325 is merged.

PR #2331, opened 2023-07-26
Bandmap2qso

int2001:
Click on Call at BandList and see what happens.

If there's a qso-Window anywhere open:
- Call, Band, Frequency will be copied to the Window, so everything for the qso is prepared

If there's no qso-Window open:
- Qso-Window will be opened (yes, in tab, because the 90s destroyed opening "popUps")
- After a graceful time of 2,5s Call, Band, Frequency will be copied to the Window, so everything for the qso is prepared

The feature is only available at BandList (perhaps later in Bandmap, too - we will see)

Please check this and give some Feedback.


int2001:
Tested the following things:
* Cross-Instance: No Data leak between Dev and prod (Prod fills prod, dev fills dev)
* Tested with and without CAT. With CAT the Band/QRG jumps back to TRX-Settings, but Call is prefilled. That's ok (for the moment)
* Tested in Safari and Chrome
* Tested with closed QSO-Tab/Window before and after QSO
* Tested with already opened Tab/Window
* ...

phl0 (approved):
Tested and works as intended (after two minor changes submitted minutes before).

AndreasK79:
Tested working ok. Nice feature.

PR #2335, opened 2023-07-28
Label fix ascii

int2001:
Refactored label-printer.

Old behaviour:
* Only Text on the label
* No Split, when worked different SATs
* No sorting by time, mode when multiple QSOs on one label

New:
* fancy ASCII-Table (tested - without gaps!)
* Split to single labels when working different SATs or having mixed SAT/HF-QSOs with one partner
* sorting improved
* Removed support for other Fonts, since table is only working with one Font!

See Screenshot:
Screenshot of two printed labels with the new Feature


AndreasK79:
This looks much better I would say. And no gaps in the vertical lines. Tested working with new labels/editing labels, and printing labels with one or more QSOs. Go for merge.

PR #2336, opened 2023-07-28
SatMode label fix

int2001:


phl0:
This needs to be applied to other places as well where SAT mode is shown to the user. At least application/views/view_log/qso.php needs that patch.

PR #2339, opened 2023-07-28
Add code to calculate 4, 6, 8 char grid squares

magicbug:
I'd like this to be sanity checked please, but basically change to the Qra.php library that brings support for 4,6,8 char grid squares

As I keep getting asked to support it

@int2001 @AndreasK79 @phl0


int2001:
?

PR #2340, opened 2023-07-28
try to fix some things...

int2001:
left comments for you in the code @magicbug 

tried it with my own location. Seems to work. Pse chk.


magicbug:
Seems to be OK

PR #2344, opened 2023-07-29
Split SAT mode in (adv) logbook and QSO view

phl0:
If SAT mode is stored with only 2 chars (e.g. "UV") display it with slash (i.e. "U/V") just like in the SAT input box during logging.


int2001 (approved):
Works on my machine...

Issue #2346, opened 2023-07-29
Advanced Logbook Filter is missing qsos


DL3EL:
you can filter qsos, if they are cionfirmed with LOTW or not. I have several qsos, which are not shown, when you select LOTW = NO. 
A closer look at the qso data shows, that these qso do not contain the field "<LOTW_QSL_RCVD:1>N".

Could that be the reason?

phl0:
> A closer look at the qso data shows, that these qso do not contain the field "<LOTW_QSL_RCVD:1>N".

What exactly do you mean by "does not contain the field"? Are you talking about an ADIF im- or export?

phl0:
Oh advanced logbook. My fault for bot reading correctly. @AndreasK79 already has an idea. So this should be fix-able.

phl0:
I proposed a patch with https://github.com/magicbug/Cloudlog/pull/2347. @DL3EL could you please verify this works on your system?

AndreasK79:
It's merged now, so pull latest dev.

DL3EL:
@AndreasK79 with that it works, thanks

PR #2347, opened 2023-07-29
Look also for QSOs where QSL column is NULL or empty

phl0:
This adresses https://github.com/magicbug/Cloudlog/issues/2346. To be tested and verified by @AndreasK79 prior to merging.


AndreasK79:
Tested working.

phl0:
So then ... undrafted

PR #2353, opened 2023-08-01
Added Menu for maintenance

int2001:
if there are QSOs without station_id only admin can reassign those QSOs to existing station locations.
Therefore a new Menu-Item at "Admin" was createt. "Maintenance".

One can choose the target-Location by station_callsign and reassign.

The old "reassign to station-id 1" at Station Locations for Users isn't shown any longer. if logged in as admin there's a hint / link for the Admin to reassign those QSOs at Admin->Maintenance.

Picture of Admin-UI for reassigning QSOs

### DIFF
--- a/application/controllers/Backup.php
+++ b/application/controllers/Backup.php
@@ -10,7 +10,7 @@ class Backup extends CI_Controller {
 	public function index()
 	{
 		$this->load->model('user_model');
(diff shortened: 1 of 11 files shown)


AndreasK79:
Easy test: update logbook set station_id = null;

Now I got the warning. I went into the maintenance menu. Since I had 2 different station callsigns in the log, I was presented with a choice for each callsign. Two clicks and all was fixed.

This would not have worked in the old solution where everything was set to id = 1.

Thumbs up.

PR #2354, opened 2023-08-01
[Advanced Logbook] Added dialog to set label offset when printing labels

AndreasK79:


int2001:
Tested:
* Label-printing (with offset) from Label-Menu --> **works**
* Label-printing (with offset) from Logbook-Advanced --> **works**

Great feature!

PR #2358, opened 2023-08-01
integrate Winkey

magicbug:
This implements the Winkey into Cloudlog using WebSerial


magicbug:
What I was hoping was buttons and stuff only show "after" connected but yes a nice turn on in account would be ace!

PR #2360, opened 2023-08-02
User-specific language setting

int2001:
From now on every User can set his own language (Fallback is english).
To achieve that, one MUST edit his config.php (sample included)

Hint to edit config.php is also implemented at Maintenance-Menu (Admin).

Once enabled you can set your language at Accounts --> Language


phl0 (approved):
Ok (by just reading through the changes).

PR #2361, opened 2023-08-02
Implemented Userlanguages with a lot of help from @AndreasK79

int2001:
From now on every User can set his own language (Fallback is english).
To achieve that, one MUST edit his config.php (sample included)

Hint to edit config.php is also implemented at Maintenance-Menu (Admin).

Once enabled you can set your language at Accounts --> Language


AndreasK79:
Tested working here after adding necessary part to config.php:
1. Edit user
2. Add user
3. Change language
4. Log in and out

PR #2362, opened 2023-08-02
Updated Simplified Chinese translation

Lingluoluo:


phl0 (changes requested):
This PR does delete Logbook_model.php! This will crash Cloudlog!

phl0:
I think this PR is based of the master branch instead of dev. This needs to be corrected (before merge).

int2001 (changes requested):
As @phl0 already mentioned the deletion of Logbook_model.php will break CloudLog.

PR #2363, opened 2023-08-02
SV Swedish translation

OuiWii:


magicbug:
Thanks @OuiWii are the ones in /application/languages for Swedish all OK too? suspect they might be a bit behind compared to the English one

OuiWii:
> Thanks @OuiWii are the ones in /application/languages for Swedish all OK too? suspect they might be a bit behind compared to the English one

I will check that, currently I am just translating the /system/ files

phl0:
Wouldn't it be better to only have a look at application/language and PR the changes to CI stuff upstream?

oh1jw:
Hi, 

@OuiWii these files are actually from **Codeigniter**, so I think it would be better to translate them to there? 
Or what do you think @magicbug

magicbug:
@oh1jw correct! much better suited.

OuiWii:
Will translate over there. Thanks for all! 73 SA0WII6 aug. 2023 kl. 23:17 skrev Peter Goodhall ***@***.***>:﻿
Closed #2363.

OuiWii:
Please ignore the text after my callsign, that stuff happens when I use email to reply..

PR #2366, opened 2023-08-03
Some minor grammatical changes for the German translation

205er:
- Email <> E-Mail; E-Mail is the correct German word for electronic mail
- US "Grafschaft" is not commonly used in German. It's better to use "County", also because "County" is used elsewhere in the German translation.


phl0 (changes requested):
Then we probably should use the correct plural as "Counties" here.

205er:
HaHa! You won't believe it, but in this case _Countys_ is correct German.
Source:
https://www.duden.de/rechtschreibung/County
Wikipedia article
https://de.wikipedia.org/wiki/County_(Vereinigte_Staaten)

phl0:
Well then I learned something new and withdraw my statement.

magicbug:
So is there an agreement for merging this in?

phl0:
> So is there an agreement for merging this in?

I would still rename to "US-Bundesstaat".

int2001:
There's a difference between US-Bundesstaat like "Alabama, Florida, etc." and US-Counties.

Counties are much smaller than states: e.g.: "Falls Church" is a county within the state "Virginia".

ARRL also differences that. There's an award for "worked all states" which targets states "US-Bundesstaaten" like "Alaska", "Virginia" and so on.

Just my 2 Cents

phl0:
Yeah I know that. Just wanted to change USA-Bundesstaat to US-Bundesstaat. I guess no German says "USA-Bundesstaat" ...

int2001:
ah. misunderstood that. thought you wanted to name the states to "countys"

205er:
I just wanted to make it more inconsistent with the existing entry "_USA County_"

image

But you are right and I agree: US-Bundesstaat and US-County would be the better choice in this case. I'll add this change to the CR.

205er:
Just one more comment on the State/County-discussion:
My original decision to use _USA County_ and _USA Bundesstaat_ was based on the English template, where _**USA** County/State_ is used (and not _**US** State/County_).
Don't know if this also should be changed in the English version? (or probably in other language versions too)

PR #2370, opened 2023-08-04
Finnish update

oh1jw:
Finnish language files, updated from **DEV** and translated. Pls check for conflicts (if this causes any) - I had no rights to update to DEV...  I'm not sure about the policy of branches. SRI!


magicbug:
Yep to make life easier for me i've blocked PRs to Master branch and only the dev as that's the branch we develop on.

Made some adjustments but should be ok now

PR #2371, opened 2023-08-04
[QSLPrint] Fixed js. Added button to mark as sent

AndreasK79:
Added a button for marking single QSOs as sent:
image
Fixed javascript when inserting from QSO List. Problem was that it didn't insert all table cells.

Split javascript to it's own file, reducing footer.php by 100 lines.


int2001 (approved):
tested. works

magicbug:
My bad, thanks for fixing

PR #2372, opened 2023-08-04
[Winkeyer] Fixed url when Cloudlog is in a subfolder

AndreasK79:
Settings can't load or save with /index.php when Cloudlog is located in a subfolder.


int2001 (approved):
tested it on my dev - works

PR #2373, opened 2023-08-04
Update of Czech translation

ok1cdj:


phl0:
This is also causing issues because it was branched of master and not dev. :)
@ok1cdj could you please base your changes on dev branch? Let me know if you need github support.

ok1cdj:
OK, now problem. Will switch to the DEV.

magicbug:
Based on that will close this PR for now

PR #2375, opened 2023-08-05
Added (optional) grid to label-printing

int2001:
Tnx @AndreasK79 for UI.

Label dialog

Label with grid enabled


magicbug:
Fantastic, saves extra time when out doing multiple grids :D

PR #2380, opened 2023-08-06
Updated some Fin and Swe files

oh1jw:
Updated some Fin files, especially added DXcluster lines to options_lang.php


magicbug:
@oh1jw thankyou!

PR #2382, opened 2023-08-07
[QSO Entry] Cleanup and tighten up db lookup from QSO entry

AndreasK79:
All the lookup code for a callsign checked the whole database. It was not filtered at all.

Now the database search are filtered on the user, and the inputs are run through xss_clean.


phl0 (approved):
Tested and working ok.

int2001 (approved):
Tested this with qrz.com and hamqth.
Working.

Tested also enrichment of e.g. "Name" in a first QSO.
At the 2nd QSO the "new" Name is fetched.
But ONLY (as it should be) from the QSOs the logged in User made so far.

PR #2383, opened 2023-08-07
Added "Already imported to clublog" to ADIF-Importer

int2001:


int2001:
Tests made (and working):
- ADIF Import (with and without new checkbox for Clublog ticked)
- API-Upload

PR #2388, opened 2023-08-07
PR to create stable release 2.4.6

magicbug:
As per the title really create a new stable release of 2.4.6 as its been a few weeks.


magicbug:
@phl0 @int2001 @AndreasK79 

Any Objections.

int2001:
Still testing things here. If everything is going well i'll approve tonight or tomorrow early in the morning.

magicbug:
Thanks, think overall it looks stable, within reason its the age old thing of sometimes just having to risk it

phl0:
I will check and test tonight. QRX

AndreasK79:
I think we should merge in the last German language translations, then release,

And if some bugs show up that we haven't found, we'll fix it.

magicbug:
In that case with that vote go for merge :)

PR #2389, opened 2023-08-07
Prevent QSO being updated twice or more when LoTW-Cnf arrives

int2001:
Otherwise Clublog and qrz.com will get updates the whole day.

LoTW always delivers the full day, so all LoTW-Confs are updated all day long (until midnight)


int2001:
Tests done here:
- Imported LotW things again via UI (as expected after patch: No set to "M" at QRZ or Cloudlog)
- Imported LotW things again via UI (as expected after patch: No set to "M" at QRZ or Cloudlog)

PR #2392, opened 2023-08-08
removed a lot of unused (and dangerous) functions

int2001:
in progress. will undraft after finished


AndreasK79:
Removing unused stuff is good for keeping the code clean and the attack surface smaller.

Can't see that these function are used in Cloudlog, so it looks good to me.

magicbug:
Thanks for going through all this :-) I mean the codebase dates back beyond 2011 so wont be pretty in places

int2001:
Yes. i am seeing ghosts meanwhile.
so better take a break and undraft this thing

PR #2398, opened 2023-08-10
[Logbook Advanced] Fixed callbook lookup

AndreasK79:
Seems I broke the callbook lookup when adding things to the advanced logbook.
This should fix the lookup.


int2001 (approved):
tested this. works and resolves a bug!

PR #2401, opened 2023-08-12
Logbook advanced: QSL slideshow

AndreasK79:
Advanced Logbook:
I added a filter to search for QSOs with uploaded QSL images.
If you then mark those QSOs, you can click on the button QSL Slideshow.
A popup with those QSLs will then appear.


int2001 (approved):
Tested this. Works fantastic!

PR #2403, opened 2023-08-14
Fixed EQSL-Col naming

int2001:
EQSL-Filters didn't work at Logbookadvanced. Fixed that


AndreasK79:
Thanks for fixing these :)

PR #2405, opened 2023-08-14
Added detail-view to todays QSLs

int2001:


phl0 (approved):
Nice feature and works perfectly.

PR #2407, opened 2023-08-15
Only show the spinner for clicked button

phl0:
Fixes an issue where all buttons show spinners during QRZ upload instead of only the clicked one.

signal-2023-08-14-093121_002


int2001 (approved):
Tested / works as expected...

PR #2417, opened 2023-08-17
Fix LoTW hints and make them a link

phl0:
This fixes the missing LoTW hints in advanced logbook. Also makes the hint a direct link to LoTW lookup of last upload from the LoTW page.


int2001 (approved):
Works. Nice Feature

PR #2421, opened 2023-08-20
Added Operatorcallsign to add QSO-Form

int2001:
Added Sessionvar for "operator_callsign" so import ADIF ("Use login-call"-checkbox) is still working


phl0 (approved):
Looks fine to me :)

AndreasK79:
Tested working. You can now override operator callsign in the QSO entry form.
Operator callsign override also survives when visiting another page in Cloudlog, then go back to QSO entry to log some more.

magicbug:
I'm not overly keen on yet another input, however until additional users can access logbooks I'll let it go in for now.

PR #2423, opened 2023-08-22
Added OperatorCall to Contesting

int2001:
Simple Field at contesting for the OP-Call.

So multiple OPs are able to work together with a ClubCall or ContestCallsign.
Simply share credentials for the Login, enter your OP-Callsign and do contesting.

Warning: When leaving Contest-Mode, and re-entering Contest-Mode you have to type in your Call again. This is intended, because assumption is: When you're doing contest, you're doing contest.

Screenshot of OP-Call-Field at Contest-Logging

### DIFF
--- a/application/views/contesting/index.php
+++ b/application/views/contesting/index.php
@@ -30,6 +30,11 @@
 									} ?>
                                 </select>
                             </div>
+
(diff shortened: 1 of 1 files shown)


AndreasK79:
Tested working.

PR #2427, opened 2023-08-24
We need to allow , in URI chars to enable vucc grid lookups

phl0:
VUCC grid lookup fail with HTTP 400 because commas are not allowed as URI characters:

Screenshot from 2023-08-24 23-06-38


magicbug:
Better solution is to probably change the look up request code

phl0:
> Better solution is to probably change the look up request code

U mean replacing comma by something permitted on lookup?

phl0:
Close in favor of https://github.com/magicbug/Cloudlog/pull/2432.

PR #2431, opened 2023-08-25
Of course we need to load the model before using it ...

phl0:
I mixed up logbook_model and logbooks_model. The former needs to be loaded before using it. Forgot in my previous commit/PR. Sorry.


int2001 (approved):
now it's working again (500 is away)

PR #2432, opened 2023-08-25
[WIP] Convert bearing lookup to POST request

phl0:
The bearing and distance functions fail for gridline and gridcorners because comma is not a permitted URI character. Instead of changing the config we convert the AJAX lookups to POST requests. Overcomes https://github.com/magicbug/Cloudlog/pull/2427 which will be closed instead.

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -917,7 +917,9 @@ class Logbook extends CI_Controller {
 	}
 
 	/* return distance */
-	function searchdistance($locator, $station_id = null) {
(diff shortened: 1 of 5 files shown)


phl0:
This is ready to be tested by @AndreasK79, @int2001 and also of course @magicbug. Would like to hear about your test results before I un-draft this PR. Feedback welcome.

int2001 (approved):
works as designed

phl0:
@magicbug What is your opinion on this?

PR #2440, opened 2023-08-27
FIX missing semicolons in Czech translation

petrkr:
Fixes error when user switch account to Czech language

```
Type: ParseError

Message: syntax error, unexpected variable "$lang"

Filename: /var/www/html/application/language/czech/account_lang.php

Line Number: ...
```


petrkr:
Sorry, rebased to dev tree

PR #2441, opened 2023-08-27
Distance details lookup fix

int2001:
## Old behaviour:

- Analytic / Distances worked: Distances are computed from every QSO fetched from the DB
- When clicking details, the QSOs were fetched by DISTANCE at DB

Since not every QSOs had a (computed or imported) distance, the details are not always show the same results than the graph.

## New behaviour
- Analytic / Distances worked: Distances are computed from every QSO fetched from the DB. If distance at DB differs from computed distance, it's being updated.
- When clicking details, the QSOs were fetched by DISTANCE at DB

So, same result at details / graph

### DIFF
--- a/application/models/Distances_model.php
+++ b/application/models/Distances_model.php
@@ -4,17 +4,16 @@ if (!defined('BASEPATH')) exit('No direct script access allowed');
 class Distances_model extends CI_Model
 {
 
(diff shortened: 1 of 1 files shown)


AndreasK79:
Tested working with no errors. The regular user will not notice anything special, as the changes are done in the backend.

PR #2442, opened 2023-08-27
IMPORTANT Security fix: eQSL uses HTTP (not encrypted) communication

petrkr:
Since eqsl using HTTPS, there is no single reason (maybe just for debugging purposes) to use un-encrypted HTTP connection, especially if there are passwords in requests.

This PR replacing endpoints to HTTPS. What it does NOT do is update in database table "config", which was seeded from migration script 004. I am not familiar with this PHP database framework. So please add there change of

`
		$sql = "UPDATE config SET eqsl_download_url = 'http://www.eqsl.cc/qslcard/DownloadInBox.cfm' WHERE id=1";
`
to 
`
		$sql = "UPDATE config SET eqsl_download_url = 'https://www.eqsl.cc/qslcard/DownloadInBox.cfm' WHERE id=1";
`

I tested it on my server and communication is TLS based and works fine.

Before patch:
image

After patch
image

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 138;
+$config['migration_version'] = 139;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -277,13 +277,13 @@ class eqsl extends CI_Controller {
 		return $table;
 	}
 
-	// Build out the ADIF info string according to specs http://eqsl.cc/qslcard/ADIFContentSpecs.cfm
+	// Build out the ADIF info string according to specs https://eqsl.cc/qslcard/ADIFContentSpecs.cfm
 	function generateAdif($qsl, $data) {
 		$COL_QSO_DATE = date('Ymd',strtotime($qsl['COL_TIME_ON']));
 		$COL_TIME_ON = date('Hi',strtotime($qsl['COL_TIME_ON']));
 		
 		# Set up the single record file
-		$adif = "http://www.eqsl.cc/qslcard/importADIF.cfm?";
+		$adif = "https://www.eqsl.cc/qslcard/importADIF.cfm?";
 		$adif .= "ADIFData=CloudlogUpload%20";
 		
 		/* Handy reference of escaping chars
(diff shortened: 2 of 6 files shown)


magicbug:
Yes would be nice if eQSL didn't want the password passed to it, it's worse that even they just store it in plain text so using a password just for the platform is important.

magicbug (changes requested):
Overall looks fine but migration files needed.

petrkr:
Yes I know, but I do not know how to write migration files in this framework. So please if you can do...

I totally agree eQSL should use some API / REST with API keys, it do not need be OAuth2, simple API key will be enough, but users are not loud enough to push them to do that :(

Also problem is, old-school loggers/apps will NOT use those API, as nobody will put there such support I guess. But they could have it as two ways "old-insecure" and new secure.

phl0:
> Yes I know, but I do not know how to write migration files in this framework. So please if you can do...

Look in application/migrations. There is various examples. Should not be too hard.

petrkr:
Well.. I will just stop doing my job and learn new framework :) I though someone help me with this as-well... Anyway, I just did some best effort without framework knowledge, I do not know if that is right way, how to do such job in this and I really do not know how to run migrations or how to test it. Or how to run downgrade (because I found there down function, so I suppose it is for downgrading db version).

phl0:
Nah not needed. I will do that then.

petrkr:
I tried to do some, but can not test, because I am running master version and there is some one more migration between dev and master branch, so I can not test it on mine server unless I will set up development environment aswell

AndreasK79:
Migration script tested working ok. No errors, and url was changed on up- and downgrade.

phl0:
> @phl0 did u checked i I've done it right way? Thanks

Yepp. Just tested. Sorry for delay. I would remove the echo output on downgrade as this will most probably not happen to any user and if they would only see it once on reloading and then gone ...

Screenshot from 2023-08-29 12-49-04

phl0:
Just talked to @AndreasK79 about this and we came to the conclusion that best would be to just leave downgrade empty. This does not even switch back to insecure HTTP at all. And would not require such notice.

magicbug:
> Yepp. Just tested. Sorry for delay. I would remove the echo output on downgrade as this will most probably not happen to any user and if they would only see it once on reloading and then gone ...

Yep agreed @phl0

petrkr:
> > Yepp. Just tested. Sorry for delay. I would remove the echo output on downgrade as this will most probably not happen to any user and if they would only see it once on reloading and then gone ...
> 
> Yep agreed @phl0

Check latest PR

Issue #2139, opened 2023-05-14
Live Contest Logging: New QSOs do not appear in QSO list after saving


m0urs:
If you are in the Live Contest Logging view and you enter a new QSO I would expect the QSO to appear below in the "Contest Logbook" list after saving.

However the list is not updated.

It is only updated after I leave the Live QSO logging view and reenter it.

Same is true for the "Post Contest Logging" view.

AndreasK79:
I can't reproduce this problem here. It will filter on current callsign when entering a call, but once that is logged, all QSOs will show in the table.

Do you get any error messages in devtools (F12) ?

phl0:
Cannot reproduce here either ... Keen to see what your dev tools reveal.

phl0:
This is how it looks here:

https://www.youtube.com/watch?v=IJxMAdVLX8s

(I did log two test QSOs in that contest before starting the recording).

m0urs:
I guess I know what is happening. As long as you do not leave the Contest Log view, everything is ok. Just leave the view and then return to that view later and start to enter new QSOs.

Now the QSOs are displayed successfully but ... they are added to the END of the view:

![image](https://github.com/magicbug/Cloudlog/assets/2588362/46d1c49d-fffd-48bd-b634-cbbacba69685)

I did not look at the end but only at the beginning of the view and therefore though they are not displayed at all.

So my expectation would be that new QSOs are always displayed at the top of the QSO list even after you left the view between two sessions.

AndreasK79:
@m0urs #2161 should fix this. At least it did in my tests. Could you give it a go?

m0urs:
> @m0urs #2161 should fix this. At least it did in my tests. Could you give it a go?

Seems to work for me now correctly. Thanks.

However: I do no longer get a "duplicate" warning!? I switched back to the current "dev" branch and I would say, in this branch there is the same issue. So it seems that it is not you last change which broke it but something before?

m0urs:
Steps to reproduce th eissue:

- Add 2 Contest QSOs (the duplicate warning appears correctly)
- Switch to the normal live QSO view and add a normal (non-contest) QSO
- Switch to the Live Contest QSO view
- Add a duplicate callsign (worked before in the contest): The duplicate warning does not appear.

AndreasK79:
I've merged the table fix.

I can't duplicate the worked before problem. I tried doing exactly what you described, and I get worked before. Not sure what is going on here. Did you reset the contest session?

[...]

PR #2447, opened 2023-08-28
[Contesting] Removed seconds from checkIfWorkedBefore()

magicbug:
Removes seconds from checkIfWorkedBefore() to try resolve error in #2139

### DIFF
--- a/application/models/Contesting_model.php
+++ b/application/models/Contesting_model.php
@@ -217,8 +217,8 @@ class Contesting_model extends CI_Model {
 		if ($contest_session) {
 			$qsoarray = explode(',', $contest_session->qso);
 	
-			$date = DateTime::createFromFormat('d-m-Y H:i:s', $qsoarray[0]);
-			$date = $date->format('Y-m-d H:i:s');
+			$date = DateTime::createFromFormat('d-m-Y H:i', $qsoarray[0]);
+			$date = $date->format('Y-m-d H:i');
 	
 			$this->db->where('STATION_ID', $station_id);
 			$this->db->where('COL_CALL', xss_clean($call));


petrkr:
It probably will need more investigation, because it randomly sends some datetime, which really do not know where it got, because time already passed... And few requests was sent with seconds too.

I tried chrome and firefox, so it is not browser dependent.

I trying read javascripts where they get those time value, because in Live Contest QSO after page loads, there is visible onle HH:MM, but then some javascript start "ticks" with seconds too.

In Post Contest QSO is only HH:MM, but user is able to add seconds.

I'll try this patch in my devlog environment

petrkr:
Well,, I littlebit investigate..

difference when it is with and without seconds depends when you did "start new contest session"..

If you did it on Live contest, it is with seconds,
If you did it on Post contest, it is without seconds because user not added them and it is pre-filled without seconds

Also if you does NOT change exchange type after "Start new session" date time is empty string

And also sometimes if do "start new contest" and not change exchange type and start type callsign,, sometimes instead of empty array it's null (because getSession return null)


well... 
if you click "start new session" it will call "deleteSession" and "setSession" at once, sometimes set is faster than delete, so result is null in getSession
sometimes is delete faster and set slower... That is that javascript async hell

```
<h4>An uncaught Exception was encountered</h4>

<p>Type: Error</p>
<p>Message: Call to a member function num_rows() on null</p>
<p>Filename: /var/www/devcloudlog/application/controllers/Contesting.php</p>
<p>Line Number: 192</p>
```

magicbug:
I'l let @AndreasK79 chime in as he knows the contesting code better than me

AndreasK79:
@magicbug nah, this breaks the qso loading in the table for me. The sessions stores the time in that format. Removing seconds breaks the code here.

AndreasK79:
I see another thing that might be a problem as well. If session is never set, the code still tries to check for QSOs, so I guess there are multiple ways for failure here.

PR #2449, opened 2023-08-28
Updating PHP to version 8.0

iu2frl:
I have upgraded the base PHP image to 8.0

This version is supported up to 26 Nov 2023, i read in the Wiki that 8.1 and later are not recommended, is this still true?


AndreasK79:
@iu2frl latest version supports 8.2

iu2frl:
Thanks, I will move to that one

petrkr:
Maybe squash that two commits into one, i think that meanstep with 8.0 is useless.

I am running it on 8.2 ar debian 12 native.

PR #2452, opened 2023-08-30
[Advanced Logbook] You can choose which columns to show

AndreasK79:
Added option to choose visible columns in the advanced logbook:
image


phl0 (changes requested):
The code shows empty logbook and error messages right after upgrade:

Screenshot from 2023-08-30 12-54-40


This is due to user_options being empty. Fix proposed in https://github.com/AndreasK79/Cloudlog/pull/18.

phl0:
Saving and refreshing works fine. PR probably needs a rebase re migrations as we just have 2.4.8 been released :)

PR #2461, opened 2023-09-03
Leave DXCC out of tweets if none

phl0:
Only include DXCC if it is not "- None -" in tweets and toots.


int2001 (approved):
Working as intended

PR #2462, opened 2023-09-03
Apply selected cols also to qso-form

int2001:
Let Usersettings (for columns) also apply for last-contacts and previous contacts at QSO-Form


phl0 (approved):
Tested and works as designed. A good idea imho.

PR #2466, opened 2023-09-06
[LBA] Use station profile table for MY_xxx info

phl0:
LBA shows references from MY_xxx columns of contacts table istead of the joined station_profile table. This leads to wrong information display if you change the station location for a QSO after logging.

Correct data in QSO view:
Screenshot from 2023-09-06 09-16-57

Wrong/old data in LBA view:
Screenshot from 2023-09-06 09-17-22


phl0:
> Tested working as it should here.

phl0:
Lets merge this. Requires further work re MY_xxx fields with redundant data however. We will cover that later.

PR #2474, opened 2023-09-11
Activated grids to gridmap layout

phl0:
Activated grids should also use the nice new gridmap page made by @AndreasK79. This is wip. Basics work. But popup is not filtering correctly yet.


phl0:
This is basically ready. Would love to receive some feedback before undrafting. Maybe @int2001 or @AndreasK79 are willing to test and tell me if it needs some more love?

int2001:
Works as designed. I really like that.

This one also fixes a security issue - at the "old" activator-view QSOs from other Users were leaked.

int2001 (approved):

PR #2482, opened 2023-09-14
API: Add additional output to QSO-call

tudacs:
Two return fields added for improved reporting:
* Messages returned from Logbook-import
* Number of QSO's added without message

To discuss:
* Number of QSO's depends on import-function returning empty string for successful import - maybe add documentation there
* Maybe add strings of errornous ADIF's to output?

I appreciate ideas/comments...

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -164,6 +164,9 @@ class API extends CI_Controller {
 
 		$this->load->model('stations');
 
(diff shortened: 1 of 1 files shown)


phl0:
Please change PR to dev branch as merge target.

tudacs:
Thank you for the hint :). As I see no way to change the branch, I close this one.

PR #2483, opened 2023-09-14
API: Add additional output to QSO-call

tudacs:
Two return fields added for improved reporting:

    Messages returned from Logbook-import
    Number of QSO's added without message

To discuss:

    Number of QSO's depends on import-function returning empty string for successful import - maybe add documentation there
    Maybe add strings of errornous ADIF's to output?

I appreciate ideas/comments...

p.s. Sorry for this duplicate PR - I saw no way to change the PR to dev-branch :(


magicbug:
As its a seperate record in the json should be ok, do we think the field names OK for the data its representing?

int2001:
Good point, Peter.
i think messages is OK.
But instead of "created" i suggest: "created_cnt". Its more representative for the value

magicbug:
Yeah or imported_count perhaps i'm happy with either then it can be merged

int2001:
Yes, same view as Peter. @tudacs can you rename the param pse?

tudacs:
Sure, thank you for your comments and suggestions!

magicbug:
I'm good with it now!

PR #2495, opened 2023-09-17
Replace slashed zero

phl0:
We may want to replace slashed zeros upon storing QSOs to the database.
This currently only replaces upon DB insert. Lookups during entering the QSO is still b0rken.


magicbug:
Frontend might be easier to just add a bit of JS validation to stop it even being entered

phl0:
I was more like replacing it. The op in https://github.com/magicbug/Cloudlog/discussions/2493 copied the callsign from elsewhere. I am just looking into changing the lookup codes to use the un-slashed zero.

magicbug:
ah yeah need to catch it coming in too then :(

PR #2497, opened 2023-09-19
use the same logic for usa county as in create function

dg0tm:
Use the same logic for usa county as in create function
otherwise for empty USA Award input Fields, an single comma
were insert into DB.


dg0tm:
wrong branch :(

PR #2507, opened 2023-09-20
Ignore Stationcallsign at import as option

int2001:
@phl0 / @AndreasK79 can u please check this?

checked here:
- Upload with ticked new box and without. Behaves like it should
- WSJT-X Logging (Behaves as before)


AndreasK79:
Tested with a modified ADIF having the wrong station callsign. 
1. Error when box not checked.
2. Imports ok when box checked.

Go for merge.

PR #2511, opened 2023-09-23
add test egual 0, for update distance

abarrau:
Add the test "egual 0" for update_distances function


phl0 (approved):
Tested and works ok for me.

PR #2521, opened 2023-09-28
Backport cookie destroy function gfrom CI 3

phl0:
Proposed fix for https://github.com/magicbug/Cloudlog/issues/2516#issue-1912141938. Awaiting test by the op before un-drafting this.


int2001:
tested on my dev - works. r4m

int2001 (approved):

PR #2522, opened 2023-09-28
Change http query build

abarrau:
With the previous version, i have systematicly this error : "Error: No such Username/Password found"
when i delete QTHNickname and change username by nickname, credential and download are OK .


abarrau:
Hello, 
IGNORE THIS PULL REQUEST !
I am sorry but it's a mistake ... My QTHNickname was bad ... 
Aurélien

PR #2528, opened 2023-09-30
Fix Gridsearch / QSO-Details

int2001:
Behaviour:

Search for Grid (e.g. awards/VUCC) "CM" shows also QSOs where "CM" is at the end (FN75CM).
But it should only show Grids which START with "CM"

Please check this, before merging @AndreasK79 or @phl0


phl0 (approved):
Makes sense to me.

PR #2535, opened 2023-10-02
eqsl: add date force update

abarrau:
Add a field to add a date if necessary, in order to resynchronize the eQsl to an earlier date.

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -69,6 +69,7 @@ class eqsl extends CI_Controller {
 				redirect('eqsl/import');
 			}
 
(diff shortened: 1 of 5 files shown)


int2001 (changes requested):
Tested this. It works as propagated. Like that feature.

But: Can you please adjust the view? Seems a little "unformatted". I suggest an extra row for it. See Screenshot.

image

abarrau:
Hello, 
Do you prefer it like this?

image

abarrau:
ok, I added a alt text

abarrau:
@int2001, i update import view file, you can check if it's ok for you now ;)

magicbug:
Yep thats better

PR #2541, opened 2023-10-04
Logbookadvanced enhanced QSL Filter

### LINKED ISSUE #2459: Select QSL requests by category (Electronic, Bureau, Direct)
I have again a problem with QSL management and Cloudlog.

We had the discussion regarding setting QSL sent status to "Requested" vice "Queued". I used that in the past to differentiate between Bureau requests (set to "Requested") and electronic requests by mail (set to "Queued").

The outcome of the discussion was that we will only use "Requested" from now on, because you do have the possibility to differentiate the request with the "Sent Method". 

So I used now only "Requested" and now wanted to create the QSL cards.

Here are my problems:

- If I am using the "Print Requested QSLs" option I do get all my requests listed and I can see if these are "Electronic", "Direct" or "Bureau" requests. That is fine.

However, if I then select e.g. only the electronic cards and click on "Export requested QSLs to ADIF file" I do get ALL requested QSLs exported (as the name of the button is saying). I do not have the possibility to create an ADIF file with only the selected ones. So I cannot use this way.

- Then I tried the "Advanced Logbook"

Here I can select all QSOs with status "QSL sent" set to "Requested", but I do not see the "Sent Method" so I am unable to select e.g. only the direct or only the Bureau QSLs.

At least I did not find that option (maybe I missed something?)

There is a possibility to select "QSL via" and that gives me a selection of "B", "D","E", and "M" but if I do set it to "E" e.g. I do not get all electronic requests. "QSL via" should also means something different ...

So my question is now:

What do I miss here? How can I now get an ADIF file with just the QSLs I want to have, means selected by "Direct", "Bureau" and "Electronic"?

HB9HIL:
Hello there

I try to fix #2459 and try to add follwing 

- adding two new Filter Options "QSL Sent Method" and "QSL Recv Method" to make search even more precise
- removed datalist from "QSL via" Form because "QSL via" should be a search for QSL Managers and not the Method 

My actual Problem is, that I can't get the filtering to get working. 

I added in `application/models/Logbookadvanced_model.php` 

```
if ($searchCriteria['qslSentMethod'] !== '') {
	$condition = "COL_QSL_SENT_VIA like ?";
	$conditions[] = $condition;
	$binding[] = $searchCriteria['qslSentMethod'].'%';
}

if ($searchCriteria['qslReceivedMethod'] !== '') {
	$condition = "COL_QSL_RECV_VIA like ?";
	$conditions[] = $condition;
	$binding[] = $searchCriteria['qslReceivedMethod'].'%';
}
```
		
to make a search possible with the vaules in `application/views/logbookadvanced/index.php`

```
            <div class="form-group col-lg-2 col-md-2 col-sm-3 col-xl">
                <label for="qslSentMethod"><?php echo lang('filter_qsl_sent_method'); ?></label>
                <select id="qslSentMethod" name="qslSentMethod" class="form-control form-control-sm">
                    <option value=""><?php echo lang('general_word_all'); ?></option>
                    <option value="B"><?php echo lang('general_word_qslcard_bureau'); ?></option>
                    <option value="D"><?php echo lang('general_word_qslcard_direct'); ?></option>
                    <option value="E"><?php echo lang('general_word_qslcard_electronic'); ?></option>
                    <option value="M"><?php echo lang('general_word_qslcard_manager'); ?></option>
                </select>
            </div>
            <div class="form-group col-lg-2 col-md-2 col-sm-3 col-xl">
                <label for="qslReceivedMethod"><?php echo lang('filter_qsl_recv_method'); ?></label>
                <select id="qslReceivedMethod" name="qslReceivedMethod" class="form-control form-control-sm">
                    <option value=""><?php echo lang('general_word_all'); ?></option>
                    <option value="B"><?php echo lang('general_word_qslcard_bureau'); ?></option>
                    <option value="D"><?php echo lang('general_word_qslcard_direct'); ?></option>
                    <option value="E"><?php echo lang('general_word_qslcard_electronic'); ?></option>
                    <option value="M"><?php echo lang('general_word_qslcard_manager'); ?></option>
                </select>
            </div>
```

Probably I miss something.


AndreasK79:
Depending on what you are searching for, this will only search for something that begins with the string:
image

AndreasK79:
If you are searching for something that is later in the string, it will not show. I suggest adding '%' as the beginning of the binding too.

HB9HIL:
At the moment I stuck a little bit. 

The Syntax of the searchCriteria in 'application/models/Logbookadvanced_model.php' is exactly the same like the existing for example 'lotwSent'

New Criteria not working

```
		if ($searchCriteria['qslSentMethod'] !== '') {
			$condition = "COL_QSL_SENT_VIA = ?";
			$conditions[] = $condition;
			$binding[] = $searchCriteria['qslSentMethod'];
		}

		if ($searchCriteria['qslReceivedMethod'] !== '') {
			$condition = "COL_QSL_RECV_VIA = ?";
			$conditions[] = $condition;
			$binding[] = $searchCriteria['qslReceivedMethod'];
		}
```

Existing and working:

```
		if ($searchCriteria['lotwSent'] !== '') {
			$condition = "COL_LOTW_QSL_SENT = ?";
			if ($searchCriteria['lotwSent'] == 'N') {
				$condition = '('.$condition;
				$condition .= " OR COL_LOTW_QSL_SENT IS NULL OR COL_LOTW_QSL_SENT = '')";
			}
			$conditions[] = $condition;
			$binding[] = $searchCriteria['lotwSent'];
		}
		if ($searchCriteria['lotwReceived'] !== '') {
			$condition = "COL_LOTW_QSL_RCVD = ?";
			if ($searchCriteria['lotwReceived'] == 'N') {
				$condition = '('.$condition;
				$condition .= " OR COL_LOTW_QSL_RCVD IS NULL OR COL_LOTW_QSL_RCVD = '')";
			}
			$conditions[] = $condition;
			$binding[] = $searchCriteria['lotwReceived'];
		}
```

I just removed the inner 'if .... == N' because the QSL Method does not need the "No" Option. 

The Search Parameters Look like this in 'application/views/logbookadvanced/index.php'

```
            <div class="form-group col-lg-2 col-md-2 col-sm-3 col-xl">
                <label for="qslSentMethod"><?php echo lang('filter_qsl_sent_method'); ?></label>
                <select id="qslSentMethod" name="qslSentMethod" class="form-control form-control-sm">
                    <option value=""><?php echo lang('general_word_all'); ?></option>
                    <option value="B"><?php echo lang('general_word_qslcard_bureau'); ?></option>
                    <option value="D"><?php echo lang('general_word_qslcard_direct'); ?></option>
                    <option value="E"><?php echo lang('general_word_qslcard_electronic'); ?></option>
                    <option value="M"><?php echo lang('general_word_qslcard_manager'); ?></option>
                </select>
            </div>
            <div class="form-group col-lg-2 col-md-2 col-sm-3 col-xl">
                <label for="qslReceivedMethod"><?php echo lang('filter_qsl_recv_method'); ?></label>
                <select id="qslReceivedMethod" name="qslReceivedMethod" class="form-control form-control-sm">
                    <option value=""><?php echo lang('general_word_all'); ?></option>
                    <option value="B"><?php echo lang('general_word_qslcard_bureau'); ?></option>
                    <option value="D"><?php echo lang('general_word_qslcard_direct'); ?></option>
                    <option value="E"><?php echo lang('general_word_qslcard_electronic'); ?></option>
                    <option value="M"><?php echo lang('general_word_qslcard_manager'); ?></option>
                </select>
            </div>
```


But the search still does not work. I think maybe there is another Dependence I forget or I don't know about it..

AndreasK79:
@HB9HIL to add a new parameter, you need to add it to the javascript function in logbookadvanced.js:
$('#searchForm').submit(function (e) {
Under data, you have all the searchcolumns.

After that, you need to add to the logbookadvanced.php controller in function search, and the final step is in the model.

I think you are missing the javascript part.

HB9HIL:
@AndreasK79 Thank so soooo much. That was the Millon Dollar Hint 😆

HB9HIL:
I've added the new Filters with your help.
The QSL Sent Method or QSL Recv Method works without Problems with the other Filters. 

Since the "QSL Sent/Recv **Electronic**" Filter means only the manual one. eQSL, LoTW isn't meant by that. Perhaps a note needs to be added. 

I'll add a additional Action Button "Received (Electronic)" to complete this task but this has nothing to do with the new Filter.

application/views/logbookadvanced/index.php
```diff
@@ -280,14 +300,7 @@ class="form-control" data-target="#dateTo" />
...
                 <label for="qslvia"><?php echo lang('filter_qsl_via'); ?></label>
-                <datalist id="qslvia" name="qslvia">
-                    <option value=""><?php echo lang('general_word_all'); ?></option>
-                    <option value="B"><?php echo lang('general_word_qslcard_bureau'); ?></option>
-                    <option value="D"><?php echo lang('general_word_qslcard_direct'); ?></option>
-                    <option value="E"><?php echo lang('general_word_qslcard_electronic'); ?></option>
-                    <option value="M"><?php echo lang('general_word_qslcard_manager'); ?></option>
-                </datalist>
-                <input type="search" list="qslvia" name="qslviainput" class="custom-select custom-select-sm">
+                <input type="search" name="qslviainput" class="form-control form-control-sm">
```
HB9HIL, line 303: just a text field now

AndreasK79:
Tested working!

@HB9HIL you should add the new filters to the mapping functions, as it uses the same filters. Check mapqsos in the js-file and in the controller.

HB9HIL:
> Tested working!
> 
> @HB9HIL you should add the new filters to the mapping functions, as it uses the same filters. Check mapqsos in the js-file and in the controller.

Done in js. In the controller I already added them

HB9HIL:
I'm done :-) #2459 should be fixed with that 👍

AndreasK79:
@HB9HIL please remove the break tags in the language files.
image

The buttons text will automatically wrap when it does not have enough space. And by having them, they will always wrap when you do have a wide enough screen. That defeats the whole point of trying to have a responsive design.

HB9HIL:
> @HB9HIL please remove the break tags in the language files. image
> 
> The buttons text will automatically wrap when it does not have enough space. And by having them, they will always wrap when you do have a wide enough screen. That defeats the whole point of trying to have a responsive design.

I understand. 

But they do not wrap.. Neither in dev mode nor in production mode (if this affects it)

grafik

grafik

magicbug:
Any adjustments to the buttons can be done with CSS

AndreasK79:
@HB9HIL I use my own custom theme, it might be that I have done some adjustments there. I'll see if I can figure out the difference and get back to you.

AndreasK79:
@HB9HIL the default theme does not do the wrapping, The other themes does the button wrapping I see.

HB9HIL:
> Any adjustments to the buttons can be done with CSS

Makes Sense...

HB9HIL:
I added it directly to the <style> in the logbookadvanced/index.php for the Buttons and also added the same for the column "name" in the table. 

Or would it be better to add this in the theme overrides?

HB9HIL:
> I added it directly to the <style> in the logbookadvanced/index.php for the Buttons and also added the same for the column "name" in the table.
> 
> Or would it be better to add this in the theme overrides?

The column "Name" is the potential longest in the whole table I think. I loaded over 1000 QSO's in my Test System and I have some Amateur Radio Clubs from South America with reaaaaaly long names 😆

magicbug:
Other option for the column is to truncate it to a set length

HB9HIL:
> Other option for the column is to truncate it to a set length

Isn't it the same problem again if a Name is longer then the (fixed) set length that it needs wrapping to don't cut the end of the text?

HB9HIL:
grafik This is how it looks actually with the default theme.

HB9HIL:
grafik
With the buttons we are talking about

AndreasK79:
@HB9HIL the button wrapping works in all but the default theme. If it is just that style you need, put it in overrides.css for the default theme. The css is:
`.btn {
	white-space: normal;
}`

application/views/logbookadvanced/index.php
```diff
@@ -38,12 +38,6 @@
   float: left;
   margin: 0 8px 0 0;
 }
-.btn {
-    white-space: normal;
-}
-#qsoList td:nth-child(11) {
```
HB9HIL: I added this before

HB9HIL:
> @HB9HIL the button wrapping works in all but the default theme. If it is just that style you need, put it in overrides.css for the default theme. The css is: `.btn { white-space: normal; }`

Moved `.btn { white-space: normal; }` from `index.php` to `overrides.css`

AndreasK79:
@HB9HIL it is better to put styles in the css files, than in the view. Thank you for moving it. This makes it easier for others to modify the themes to their liking.

HB9HIL:
> @HB9HIL it is better to put styles in the css files, than in the view. Thank you for moving it. This makes it easier for others to modify the themes to their liking.

Makes sense :-) I still see issues in the responsive Design of the Table itself. But this maybe some other time in another PR 👍

AndreasK79:
> Makes sense :-) I still see issues in the responsive Design of the Table itself. But this maybe some other time in another PR 👍
Yes, smaller PR's are usually better. 

Which issues are you seeing in the table?

HB9HIL:
The Table allover width is not responsible. 

I loaded over 1000 QSO's to my Test-Machine and can see, that the "Name" Column is the potential longest. So because of some overrides in the different themes it doesn't truncate the table width to the display width. 

I think the best table design has the theme "cosmo"

grafik
The table is cut at the right end in default theme.

AndreasK79:
Some people like everything on one line, instead of columns wrapping, others like to see the whole table but columns need to wrap to fit. The best thing is to find a theme you like, and perhaps alter that.

HB9HIL:
Thats true. Skip that for the moment. 👍😊

HB9HIL:
What do you guys think about the filter?

int2001:
Working well for me. i like

AndreasK79:
Tested working! Great job. Go for merge.

magicbug:
Looks good here too

PR #2548, opened 2023-10-06
Contest Translation

HB9HIL:
I added several multilanguage supports for Contesting and the QSO Data Window. 

Also corrected config/autoload to add the missing language files to that list.


HB9HIL:
Question: What is the best solution to add multilanguage support to the bootstrapDialog (`assets/js/bootstrapdialog/js/bootstrap-dialog.js  Ln 211 to 223`)?

Maybe this in the the view files like I did it in `application/views/contesting/add.php Ln 72`:
```
<script>
	var lang_admin_contest_add_contest = '<?php echo lang('admin_contest_add_contest'); ?>';
	var lang_admin_close = '<?php echo lang('admin_close'); ?>'
</script>
```
?

AndreasK79:
Yes, because you can't use PHP in the JS files, so you need to output the language text to javascript variables, and then use them later in the functions.

HB9HIL:
I think these changes are good now. Could you test it?

AndreasK79:
Tested the PR. No errors found as far as I can tell. Seems safe to merge.

@HB9HIL good job!

HB9HIL:
I noticed that this warning appears instantly when opening the "Add Contest" Dialog. Shouldn't the warning only appear after an attempt has been made to click "Create"?

grafik

The same Warning in "Edit Contest" does not appear instantly. I can't find the problem actually if this is a Problem.

HB9HIL:
> I noticed that this warning appears instantly when opening the "Add Contest" Dialog. Shouldn't the warning only appear after an attempt has been made to click "Create"?
> 
> grafik
> 
> The same Warning in "Edit Contest" does not appear instantly. I can't find the problem actually if this is a Problem.

Same behaviour in `master` branch

magicbug:
They really shouldn't show unless they are empty on submit.

HB9HIL:
Can you check this? I wasn't able to find out whats the Problem here..

AndreasK79:
I think that problem has nothing to do with this PR. It has probably been there since I implemented that contest stuff. It probably has to do with the dialog being loaded. Let us focus on this PR right now.

@magicbug agree on merge?

int2001:
We are all very very happy with your translations, @HB9HIL 
Do you plan to continue that? Especially the station_profile pages need translation...

only a question - no pressure ;)

HB9HIL:
> We are all very very happy with your translations, @HB9HIL Do you plan to continue that? Especially the station_profile pages need translation...
> 
> only a question - no pressure ;)

Absolutely :-) my plan is to do the whole code :-)

HB9HIL:
> We are all very very happy with your translations, @HB9HIL Do you plan to continue that? Especially the station_profile pages need translation...
> 
> only a question - no pressure ;)

https://github.com/magicbug/Cloudlog/pull/2564 😉

PR #2551, opened 2023-10-07
Also lookup VUCC_GRIDS for hams.at grids

phl0:
We should also regard VUCC_GRIDS as worked for hams.at passes (currently only looking up GRIDSQUARE).


magicbug:
Nope your right total oversight on my part!

Issue #2473, opened 2023-09-11
Feature request: quick way to log a new QSO


jholubinka:
Hello,

Adding a new entry in the log should be an essential feature of a logging app. In Cloudlog it's hidden in the menu - it needs 2 clicks (or 3 on mobile view) and it's not even in the first place. I made a quick and dirty hack which replaces the search box with a form that collects a callsign and redirects to the "Post QSO" page with populated callsign. I'm fully aware that this patch is awful, but I hope it will give you a little bit of inspiration for doing it as it should be ;)

Kind regards,
Jakub SP9VIK

```diff
diff --git a/application/views/interface_assets/footer.php b/application/views/interface_assets/footer.php
index 262abec5..00569f46 100644
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1052,6 +1052,7 @@ $(document).on('keypress',function(e) {
     });

      // Callsign always has focus on load
+      $("#callsign").focusout();  // Force reload of a list of previous QSOs
       $("#callsign").focus();

       if ( ! manual ) {
diff --git a/application/views/interface_assets/header.php b/application/views/interface_assets/header.php
index 234a479f..0b3e6736 100644
--- a/application/views/interface_assets/header.php
+++ b/application/views/interface_assets/header.php
@@ -195,10 +195,11 @@
     </ul>

      <?php if($this->optionslib->get_option('global_search') != "false" || $this->session->userdata('user_type') >= 2) { ?>
-               <form method="post" action="<?php echo site_url('search'); ?>" class="form-inline">
-               <input class="form-control mr-sm-2" id="nav-bar-search-input" type="search" name="callsign" placeholder="<?php echo lang('menu_search_text'); ?>" aria-label="Search">
+               <form method="get" action="<?php echo site_url('qso'); ?>" class="form-inline">
+               <input class="form-control mr-sm-2" id="nav-bar-search-input" type="search" name="callsign" placeholder="New QSO" aria-label="New QSO">
+                <input type="hidden" name="manual" value="1">

-               <button title="<?php echo lang('menu_search_button'); ?>" class="btn btn-outline-success my-2 my-sm-0" type="submit"><i class="fas fa-search"></i></button>
+               <button title="Add QSO" class="btn btn-outline-success my-2 my-sm-0" type="submit"><i class="fas fa-plus"></i></button>
                </form>
        <?php } ?>

[...]

PR #2554, opened 2023-10-08
Quicklog Feature

HB9HIL:
Maybe this could a nice additional Feature 

Reference to #2473


HB9HIL:
@magicbug What is you opinion in general to a "Quicklog" Feature? Is it worth it to work on that?

HB9HIL:
I've been thinking about your feature request, and I think it's going to be pretty tough to make sure it works well on every page.

One option would be to put the Quicklog field in the header, but that's kind of like your two-in-one idea, which goes against the usual design rules.

Another idea could be to add a setting in the user options that lets you choose whether you want to see the Quicklog field instead of the search field. However, this would mess with the design a bit and make it harder to find the search function (we'd need to add another option to the dropdown).

Right now, I'm not sure what the best way is to meet your request without messing up the basic design principles.

HB9HIL:
Note:

In the header.php I removed 

```
<?php if ($this->optionslib->get_option('global_search') != "false" || $this->session->userdata('user_type') >= 2) { ?>
```
since this is not in use anymore (Ref: [Commit e36daaa](https://github.com/magicbug/Cloudlog/commit/e36daaa6bd5d17ba893789239f6be3c8be78eb52) )

AndreasK79:
Gave it a test. On/off option works. Search works, quicklog works. Seems to me this is good to go.

HB9HIL:
Problem to use Enter if Quicklog is enabled

HB9HIL:
Added a second Quicklog Option so the User can choose what shall happen when pressing "Enter" in the Quicklog Field

HB9HIL:
Ready for Testing and awaiting Feedback 😄

PR #2556, opened 2023-10-09
[migrations] remove duplicate file

dg0tm:
i think we could remove the one file with space in the filename...


magicbug:
Completely didnt spot that file!

PR #2557, opened 2023-10-09
Keep unmodified callsign for LoTW/qrz/hamqth lookups

phl0:
With https://github.com/magicbug/Cloudlog/commit/1bfb44268b87ec4af59e83ed3362a380119df83f the lookup and link functions to LoTW / qrz.com / hamqth.com broke for prefixed and suffixed callsigns like DJ7NT/P:

Screenshot from 2023-10-09 13-35-24

Screenshot from 2023-10-09 13-35-00

So we need an unmodified copy of the callsign for those.

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -504,6 +504,7 @@ $("#callsign").focusout(function() {
 		}
 
 		var find_callsign = $(this).val().toUpperCase();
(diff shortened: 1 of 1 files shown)


int2001:
tested this. it's working. tnx 4 fixing @phl0 - ready 2 merge

PR #2561, opened 2023-10-11
Pattern-matching on QRZ-APIKey

int2001:
Some Users type passwords or sth. else than the qrzkey into the Field.
This should prevent it


phl0 (approved):
Makes sense. Patterns everywhere :D

int2001:
image

PR #2562, opened 2023-10-11
Add link to eqsl.cc user profile

phl0:
Needed to lookup some users last uploads for eqsl. Thought this might be useful for others as well.


int2001 (approved):
good idea. its working

PR #2563, opened 2023-10-11
Search for "Base-Call" instead of Prefix+Call

int2001:
Old behaviour:

Lets say i worked DF2ET while he was at PA. So i logged PA/DF2ET
After a few weeks, DF2ET is QRV from HB0, so i enter "HB0/DF2ET". No previous QSO with DF2ET will show up. And no QSO will show up at Search-Bar.

New behaviour:
Lets say i worked DF2ET while he was at PA. So i logged PA/DF2ET
After a few weeks, DF2ET is QRV from HB0, so i enter "HB0/DF2ET". All previous QSOs with DF2ET will show up at search, the last 5 at logging

@AndreasK79 / @phl0 : plz have a look / test this. tnx in advance

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -135,7 +135,18 @@ class Logbook extends CI_Controller {
 		];
 
 		$return['dxcc'] = $this->dxcheck($callsign);
-		$return['partial'] = $this->partial($callsign);
+		$split_callsign=explode('/',$callsign);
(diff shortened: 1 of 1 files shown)


AndreasK79:
Yeah, it's a great idea. Tested working here.

int2001:
so ready 4 merge @magicbug

PR #2566, opened 2023-10-15
Generalize band and also use in QSO entry

phl0:
Addresses https://github.com/magicbug/Cloudlog/discussions/844 by using default band also for QSO entry. Also some cleanup in view.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 144;
+$config['migration_version'] = 145;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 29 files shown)


phl0:
Not sure if I should also rename variables and especially SQL table column names?!

HB9HIL:
If you change files in application/language/.. you have to change these files in all available languages in order to stop getting the languages unsynced. If you change just english and german the other languages wont work anymore ;-)

HB9HIL:
> If you change files in application/language/.. you have to change these files in all available languages in order to stop getting the languages unsynced. If you change just english and german the other languages wont work anymore ;-)

just put in english in the other language files

phl0:
Good point. Done.

AndreasK79:
> Not sure if I should also rename variables and especially SQL table column names?!

I think that should be done. Makes it less confusing for others looking at the code.

phl0:
Updates branch and renamed both variables in code and database. Please review and comment :)

HB9HIL:
tested and worked :-)

PR #2570, opened 2023-10-16
We should not lie about dupes on the success page ...

phl0:
Result page of ADIF import says dupes were skipped even if we forced to insert dupes.


int2001 (approved):
that's true.

AndreasK79:
The ADIF File has been imported. Dupes were inserted!

Merge time

PR #2573, opened 2023-10-16
Multilanguage Support for ADIF Import/Export Menu

HB9HIL:


phl0:
Tested and works ok on my end :-)

PR #2576, opened 2023-10-17
Lba dupes

int2001:
Adds dupe-search to logbook-advance


AndreasK79:
Tested working. Good idea! Go for merge.

PR #2586, opened 2023-10-18
Beautify DCL sync output and add some more useful information

phl0:
Also completes content which had to be translated so @HB9HIL should be happy now :)


HB9HIL:
> Also completes content which had to be translated so @HB9HIL should be happy now :)

Thank you 😄 👍

int2001 (approved):
tested this. great & beautiful :)

PR #2587, opened 2023-10-18
Multilang QSO Deletion Warning

HB9HIL:
The translation process is still a work in progress and goes on and on and on 😄


int2001 (approved):
Tested this. Works as expected. Ready 2 merge.

PR #2588, opened 2023-10-18
Correct default bandgroup data

HB9HIL:
Default Bandgroup is now the correct one for 160m, 23cm and 13cm

160m is 'mf' not 'hf'
23cm is 'uhf' not 'shf'
13cm is 'uhf' not 'shf'


magicbug:
This is OK for new installs but doesn't resolve it for existing

PR #2597, opened 2023-10-19
Gridmaster map

phl0:
Add map view for US Gridmaster


phl0:
Does render unneeded grid which slows down the map. Need to check.

magicbug:
Yep was going to say loading takes a decent amount of time, but looking good flo

phl0:
Made some bugfixes and some more styling. Ready to integrate now. Keen to read your feedback also :)

PR #2598, opened 2023-10-20
Fix quick search

AndreasK79:
This should take care of quick search not working if columns are hidden


phl0 (approved):
Works as expected :)

int2001 (approved):
works as designed

PR #2607, opened 2023-10-22
JA gridmaster and a more generalized approach serving as blueprint for

phl0:
future gridmaster map pages


AndreasK79:
Tested working :)

PR #2615, opened 2023-10-24
CQ Award Multilanguage Support

HB9HIL:
Added multilanguage support for the CQ Award View

also:

- changed german Translation from "Auszeichnungen" to "Diplome", shorter and fits bit better the use case
- added general language to footer for Map views to make later translation process easier
- added some margin to the cqmap and the cq table at the bottom

### DIFF
--- a/application/views/awards/cq/index.php
+++ b/application/views/awards/cq/index.php
@@ -1,8 +1,7 @@
 <style>
     #cqmap {
-	height: calc(100vh - 430px) !important;
+	height: calc(100vh - 480px) !important;
 	max-height: 900px !important;
-    margin-bottom: 30px !important;
 }
 /*Legend specific*/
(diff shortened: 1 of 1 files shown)


application/views/awards/cq/index.php
```diff
@@ -2,6 +2,7 @@
     #cqmap {
 	height: calc(100vh - 430px) !important;
 	max-height: 900px !important;
+    margin-bottom: 30px !important;
```
AndreasK79, line 5: I think it's better to adjust the height above instead like this: 
height: calc(100vh - 460px) !important;
Can you test that instead and see if that works better?

HB9HIL:
Now it should be better. My mistake

AndreasK79:
Agreed. Merge time.

Issue #2612, opened 2023-10-24
Add QSL-VIA to label when available


magicbug:
Add QSL-VIA to label when available

PR #2618, opened 2023-10-25
Added sorting like printing and callbook-lookup to Label-List

int2001:
Added sorting like printing and callbook-lookup to Label-List.
You can now click on a link to qrz.com / hamqth or eQSL to check the via. It's often placed within the Text on the page.

One step to resolve https://github.com/magicbug/Cloudlog/issues/2612


AndreasK79:
Tested working!

Issue #2555, opened 2023-10-09
Add validation to the date/time fields in QSO


ntilley905:
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the forum see readme for details**

**Describe the bug**
Entering a QSO with an extra digit in the time produces a database error.

**To Reproduce**
Steps to reproduce the behaviour:
1. In the time section on a post QSO, enter a typo in the field such as "718:07"
2. Finish the remainder of the log and submit it
3. Error described below occurs

**Expected behaviour**
Input sanitation to verify that the entry can be accepted by the software.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux (Pop OS)
 - Browser: Firefox
 - Version: 118.0.1 (64-bit)

**Additional context**
Error log:
```

A Database Error Occurred

Error Number: 1292

Incorrect datetime value: '2023-10-08 718:07' for column `cloudlog`.`TABLE_HRD_CONTACTS_V01`.`COL_TIME_ON` at row 1

[...]

PR #2620, opened 2023-10-25
Mark invalid time and date with red border

phl0:
Addresses https://github.com/magicbug/Cloudlog/issues/2555 by setting a pattern for date and time and mark input fields if pattern does not match.

Screenshot from 2023-10-25 09-33-25


int2001 (approved):
works as designed / expected

magicbug:
Thanks Flo

PR #2632, opened 2023-10-27
Qso time

phl0:
This adds a separate input field for QSO end time.


HB9HIL:
> tested:
> 
> * Live QSO: Working
> * Post QSO: Working
> * Live-Contest: Working
> * Post-Contest: Working
> * checking details of qso in detail-view: Working

copy that

AndreasK79:
I have just given it a quick spin, but one thing I would like to see, is an option to always set start_time = end_time. It needs to be a user option, as this must be set on user basis.

magicbug:
> I have just given it a quick spin, but one thing I would like to see, is an option to always set start_time = end_time. It needs to be a user option, as this must be set on user basis.

Yep this!

phl0:
Merge conflict here due to Swedish language being updated in the meantime. So closing here and new PR with squashed commits in https://github.com/magicbug/Cloudlog/pull/2643.

PR #2634, opened 2023-10-28
No of unassigned qsos

phl0:
Besides showing the number of QSOs we also show the basic information about QSOs without station id. Also we let the user select which QSOs to assign to the selected profile.


AndreasK79:
Tested working. Nice addition :)

PR #2624, opened 2023-10-25
Simple Fast Log Entry

HB9HIL:
It was topic in the discussions and is a great efficient way to get the paper logs from the outdoor sessions into cloudlog. 

'Fast Log Entry', or simply 'FLE' is a system to log QSOs very quickly and efficiently. Due to its syntax, only a minimum of input is required to log many QSOs with as little effort as possible.

FLE was originally written by DF3CB. He offers a program for Windows on [his website](https://df3cb.com/fle/). [Simple FLE](https://github.com/ok2cqr/sfle) was written by OK2CQR based on DF3CB's FLE and provides a web interface to log QSOs.

Now this cool tool should find its place in Cloudlog.

**It's still in progress. The current status is:** 

- Basic functionality is given
- Added Info box about FLE
- Removed 'Sample Log' - Button and added 'Syntax Help' - Button (~~Syntax Help Text is pending~~)
- Operator Call is prepopulated but editable
- Multilanguage Support is missing

Still working on that. 👍 

Any contribution is welcome 😄

PR #2640, opened 2023-10-31
SimpleFLE Improvements

HB9HIL:
In this additional PR to #2624 I added following:

- added POTA and IOTA support
- fixed WWFF logging (was to sig info instead of wwff_ref)
- improved regex expressions to cover all sota, pota, iota and wwff refs (there was an issue with numbers in the ref like W6/CT...)
- commented out the adif download function since we do not need that anymore

For the PR #2624 I was a few minutes too slow 😄


magicbug:
If we don't use it, then rather than comment it perhaps better to just delete it to save a few bytes

HB9HIL:
Deleting code feels soo.... final .. 😆 no problem :) is done

PR #2653, opened 2023-11-04
Add timer reset buttons

phl0:
This adds two buttons for the post QSO log window. The start button sets start and end time to current UTC time. The end button sets end time to current UTC time.

Screenshot 2023-11-04 153720


HB9HIL (approved):
works as expected 👍

PR #2656, opened 2023-11-05
Translation Work

HB9HIL:
Added multilanguage-support for following pages:

Analytics:
- Statistics
- Gridsquare Map
- Activated Gridsquares
- Gridsquare Activators
- Distances Worked

Also I added Translation for the 'Close' Buttons in whole Cloudlog. As usual for my Translation Work this is a bigger PR


magicbug:
Seems to be OK at first test will mere into dev

PR #2657, opened 2023-11-06
Quicklog Bug Fixes

HB9HIL:
This PR should fix 

- on Android Phones the Quicklog Field is marked as Search Field
- Callsign Lookup should kick in after switching to qso window automatically
- Reset Button doesn't work properly when coming from Quicklog Field


HB9HIL:
Quicklog Bug Fixes are ready for review 👍 please feedback

PR #2661, opened 2023-11-06
[DRAFT] load map / leafembed.js (new code, for use this function with argument)

abarrau:
this PR maintains compatibility with function calls from the footer. 

a 2nd step will then allow the code in footer to evolve to offer configuration to the user

PS: perhaps requires the cache delete of your browser


abarrau:
I will do a new project ... with all files
I close this

Issue #2623, opened 2023-10-25
Contest logging: Worked stations do not appear on list.


205er:
Hi,

I use Cloudlog for all my Ham radio activities, including contest logging. And I am very pleased with Cloudlog! Thanks for your effort and good work!
Unfortunately, I discovered a minor glitch, when logging contest QSO:
Already worked stations appear on the bottom list, as soon as I enter a call sign. This only works, when I open "live contest logging" freshly. All stations that have been entered during a "session" (without accessing this page again through the menu) will never appear in this list. Nevertheless, the warning message "Worked before!" will be displayed, even if the QSO entry in the list is missing. This is somehow confusing. The attached screenshot will show the issue:
![Screenshot of Contest log](https://github.com/magicbug/Cloudlog/assets/14996169/83147993-a7e4-4350-b16f-82cdd53dce30)
Is this a bug or just a issue of my server configuration?

Thanks & br!

dk5cr:
Same issue with my version. Was the same before updateing to 2.4.11 (2.4.5 before). Also the "Worked before" sign with white letters in red box doesn't appear when entering a Callsigh, which is worked before. Both is SSB and 80m, but no "match". The worked station (M6T) is visible in the list of all QSO and it's in the same contest. 
Problem occurs everytime you switch to another menu item in CloudLog and then switch back to (live) contest logging. Contest is - of course - correctly selected.

205er:
The "Worked before" warning appears as soon as you leave the call sign field (switch to next input field). It might take a moment, but that always works for me.
It's only the non-appearance of QSO in the Contest logbook (on the bottom), that bothers me.

dk5cr:
Thanks for the reply, but after typing in the callsign and leaving the field, drinkin' a beer nothing appears like "worked before". 
Perhaps it will be shown, if the worked callsign is in the list under the entry fields but... 

So I hope for a quick solution to this issue

phl0:
Hej @205er can you please have a look the JS console of your browser if any errors are reported?

205er:
Hi @phl0,

I just checked the contest logging fuctionality again and found the following errors in my browser's dev tool:
![Browser console](https://github.com/magicbug/Cloudlog/assets/14996169/0eae0961-3c27-4c5e-9082-34cb166c7ac6)
![Browser console](https://github.com/magicbug/Cloudlog/assets/14996169/495b8f9e-a1e3-4901-969d-fc64cdf8577d)

[...]

PR #2668, opened 2023-11-08
fix if cat_timeout_interval is empty

abarrau:
fix an js error, when cat_timeout_interval is empty

JS Error

case to reproduce : go to "live contesting" ; add a new qso --> error on JS browser

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1421,8 +1421,9 @@ $(document).on('keypress',function(e) {
 					  $("#selectPropagation").val(data.prop_mode);
 
 					  // Display CAT Timeout warning based on the figure given in the config file
-					  var minutes = Math.floor(<?php echo $this->optionslib->get_option('cat_timeout_interval'); ?> / 60);
-
+					  var cat_timeout_interval = "<?php echo (!empty($this->optionslib->get_option('cat_timeout_interval')))?$this->optionslib->get_option('cat_timeout_interval'):0; ?>";
+                      var minutes = Math.floor(cat_timeout_interval / 60);
+					  
 					  if(data.updated_minutes_ago > minutes) {
 						  $(".radio_cat_state" ).remove();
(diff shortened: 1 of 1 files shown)


abarrau:
this fix allows you to close the issue #2623

int2001 (approved):
Tested that. Works.

But one Question: Why should "cat_timeout_interval" at DB be NULL or empty? That smells a little like a migration-script or sth. else failed...

magicbug:
Something is odd as the default is 1800 inserted into the table in migration file 068

abarrau:
For my part, this option does not exist in the database.

During a new installation this value is not added by default (not exist in install.sql). 
So this field remains empty, if we do not use CAT and we do not need to go to the options to add a value.
Migration script 68 is not run during a new installation.

Furthermore, I must not be the only one, because this fix fixes issue #2623.

In principle, any case that happens at least once, you have to plan for the case where it could happen again ;)

73

phl0:
Could only reproduce the error by deleting the option from the DB. Imho you should check why the timeout interval is not existent. The migrations should run after CL is installed.

int2001:
Yes Full ACK to @phl0 - without running Mig-Scripts there's nothing populated. Only a very raw and rudimentary Database. Cloudlog will fail everywhere without those scripts.

magicbug:
It shoulds like your install of Cloudlog is broken if the migration files arent running correctly as they are here in test environments and for new setups I've been doing.

abarrau:
I understand your feedback.
adding security, even if the case is unlikely, does not commit anything ;)
You can close in this case
73

phl0:
@abarrau What is the value of "version" in your "migrations" table?

magicbug:
Note:

That cat_timeout_interval is set by default using the migration script **068_add_cat_timeout_to_options** thus PR suggests that this fix is related to an incorrectly setup instance of Cloudlog rather than a code issue.

PR #2670, opened 2023-11-09
Fix lookups for empty contest sessions

phl0:


magicbug:
What @int2001 said seems to be resolve the problem

dk5cr:
Hi, seems to me, that there's a  } missing. First if opens with a {, second if has also its {, but before return there's only one closing bracket }. (Is this the right name for this sign?) 

	header('Content-Type: application/json');
		if ($result->num_rows()) {
		if ($result && $result->num_rows()) {
			echo json_encode(array('message' => 'Worked before'));
		}
		return;

Thanks

PR #2671, opened 2023-11-09
Skip rove table if empty and display info message

phl0:
Show a nice message to the user instead of only table headers (which can be quite confusing).


int2001 (approved):
Tested. Looks great

PR #2672, opened 2023-11-10
Make rove comment a separate column

phl0:
Would like to have the comment in a separate column to make it easier to see.

Before (only on hover):

Screenshot from 2023-11-10 07-31-20

After:

Screenshot from 2023-11-10 07-34-46


int2001 (approved):
like it

PR #2673, opened 2023-11-10
Tools : manage language by web page

abarrau:
Tools for dev/admin
- User can change value of tag language by a web page
- if tag not exist on language file, tag is create

### DIFF
--- a/application/controllers/Admin_check.php
+++ b/application/controllers/Admin_check.php
@@ -1,0 +1,133 @@
+<?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
+
+class Admin_check extends CI_Controller {
+
+	function __construct() {
+        parent::__construct();
+        $this->load->model('user_model');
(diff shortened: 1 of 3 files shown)


abarrau:
Hi,
I propose this PR for dev/admin (it's a 1st version, but some functionnaly can add)

How to use (hide URL, not link from cloudlog interface, but must be added) : 
- https://URL/cloudlog-gith-dev/index.php/admin_check/language_check 

- script list all file in language folder ; the folder reference is "english"
home list

- choose 1 ou more language, and check if file exist and compare data between english and other language
list of file by language folder

- click on file name, all tag of this file are visible
tag list

- click on the tag value for language, that you would change
- if tag value not exist return "-"
change tag

- change value and save
- if tag value not exist, it will be create, event if, it will update (show in green color after save)
save tag

abarrau:
this tools, integrate in cloudlog, is not the good way, 
i will propose an other way for this tools (out of cloudlog) for helping translater people.

PR #2674, opened 2023-11-11
Fix POTA regex

rsammelson:
POTA identifiers are allowed to have 5 numbers after the hyphen, see for example https://pota.app/#/park/K-10000.


HB9HIL (approved):
Even if I'm pretty careful with formatting files this looks good to me 👍 
Thanks @rsammelson for that fix. I wasn't aware of that there could be 5 numbers  😆

PR #2677, opened 2023-11-11
Pass error message given by qrz.com to the user

phl0:
There is an issue in the qrz lib line 74. This should catch errors with empty results but does not work because qrz.com always gives an XML result. In case of an error or not found callsign this XML contains an error message. But as it is not empty the result page for looking up a non existent callsign looks weird:

Screenshot from 2023-11-11 21-33-41

This PR parses also the error message contained in the XML result an passes it to the user on the results page:

Screenshot from 2023-11-11 21-34-40

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -919,9 +919,14 @@ function worked_grid_before($gridsquare, $type, $band, $mode)
 						}
 
 						$data['callsign'] = $this->qrz->search($id, $this->session->userdata('qrz_session_key'), $this->config->item('use_fullname'));
-						$CI = &get_instance();
-						$CI->load->model('logbook_model');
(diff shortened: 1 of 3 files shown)


HB9HIL:
Throws an error if hamqth is configure in the config

phl0:
> Throws an error if hamqth is configure in the config

Tnx will check.

HB9HIL (approved):
approved :)

phl0:
Need to check / decide whether it is worth to implement the callbook search function also for hamqth.com ... Opinions?

magicbug:
In theory would be nice if it supported HamQTH but only if its not a pain in the arse

phl0:
Merge this one and I will look into hamqth.com separately.

PR #2678, opened 2023-11-12
[BUG FIX] Empty results if click on "More QSOs" in QSO Windows

HB9HIL:
This fixes a bug, if it shows empty results if click on "More QSOs" in the QSO Windows. 

The POST reference was renamed since Quicklog to avoid form fields with the same ID

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -924,8 +924,8 @@ function searchButtonPress(){
 
 $(document).ready(function(){
 
-  <?php if($this->input->post('searchBar') != "") { ?>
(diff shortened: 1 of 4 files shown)


HB9HIL:
@magicbug Wait with that pls
Probably It needs something more

HB9HIL:
Changed the id's back to callsign since this doesn't cause a buggy behaviour of quicklog anymore (since this is using localStorage now instead of GET).

int2001 (approved):
checked click/search from:
- Dashboard: Works
- LBA: Works
- Overview: Works
- Search: Works

Without that previous QSOs are not searchable by click

HB9HIL:
@magicbug now it's good 😄

phl0 (approved):
Tested. Works ok.

PR #2679, opened 2023-11-12
Show latest Migration in Database in the Debug View

HB9HIL:
This adds a new model and shows the latest run migration (value from the database) in the Debug View.

This could help to debug issues in user installations if we easily know which was the latest migration.

db_migration

-- If there is more than one value in the database, there is definitely something wrong. So, show an error.

db_error_warning


magicbug:
Thanks for making the change to a model for the db calls!

PR #2685, opened 2023-11-13
Add a minimum random Offset to Latitude at map

int2001:
Add a minimum random Offset to Latitude at map, because otherwise the pins of HV/1A0/I are not visible/overlapping.
Same happens at other "Micro-DXCCs"

### DIFF
--- a/assets/js/sections/dxccmap.js
+++ b/assets/js/sections/dxccmap.js
@@ -136,8 +136,10 @@ function addMarker(L, D, mapColor, map) {
         html: `<span style="${markerHtmlStyles}" />`
     })
 
+    var rndoff=(Math.floor(Math.random() * 10)/250);
+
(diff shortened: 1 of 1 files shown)


int2001:
@HB9HIL creates a ticket at the data-supplier to fix the coords there. Is more elegant than this PR.
Will leave the PR open/as draft as fallback if supplier wont react...

phl0 (approved):
I have to admit I don't like it very much but it works as interim solution to un-hide the DXCC.

magicbug:
Yeah I'm not keen on this to be honest this adds offsets to everything and makes things less precise which was never my plan for the maps.

Sorry

int2001:
i am also not happy with that. So let's wait for the answer from the ClubLog-People

PR #2686, opened 2023-11-13
Sub menus

phl0:
This allows the navbar menu to contain sub menu items. Also makes the dropdown menu active on hover (saves a click).

Screenshot from 2023-11-13 23-56-47


HB9HIL:
I love this menu. But I also have to say that the whole menu now acts pretty weird on mobile phones. For mobile use and touch display's it's not as good as the one before.

phl0:
Marked as as draft as needs some tweaking re usage on mobile (i.e. iOS) devices.

phl0:
@AndreasK79 will come up with a better approach. So closing here.

PR #2688, opened 2023-11-14
Focus in QSO Live View if no Quicklog Value is set

HB9HIL:
In Live QSO View the focus jumps to mode even if no Quicklog Value is set. 

This fixes that


AndreasK79:
Yep, this fixes it. Thanks.

PR #2690, opened 2023-11-14
Bugfix Resettime

HB9HIL:
There is an Issue that the Reset Time Button in QSO View only works if a dxcluster url is configures. This because of the placement of the javascripts in the particular line of the footer  (inside the brackets `if ($this->optionslib->get_option('dxcache_url') != ''){ ?>`

To reproduce this behaviour:

Delete the dxcluster cache url in the admin menu. Then go back to qso window and enter a callsign and move to the next field to trigger the callsign lookup. After a few seconds try to reset the start time -> doesn't work. 
So if a User does not have configured a dxcluster cache he won't be able to use the reset time button.

Moved the javascript to `assets/js/sections/qso.js` because the reset time buttons are only available in qso window.

@phl0 Can you check this one please? The QSO Timers are your baby.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1000,37 +1000,6 @@ $(document).on('keypress',function(e) {
 					}
 				);
 			});
-			$('#reset_time').click(function() {
-				var now = new Date();
-				var localTime = now.getTime();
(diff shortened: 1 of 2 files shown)


phl0 (approved):
Confirmed the bug. What a dumb mistake. The fix works fine. Go for merge.

PR #2691, opened 2023-11-15
Prevent fetching image from non-existent qrz.com entry

phl0:
We have introduced a hand full of problems with passing the qrz.com lookup information to the user. Collecting fixes here for now.


phl0:
Ok basically it was just an issue with (setting and) passing parameters to the results view. This is fixed now so that the lookup function delivers results regardless if there are previous QSOs or the callsign was looked up on qrz (and found or not). Ready for merge.

HB9HIL (approved):
works great. thx for this fix :)

PR #2695, opened 2023-11-15
HamQTH Lookup for search

phl0:
HamQTH Lookup for search function implemented.

ToDo:

- HamQTH lookup for previous QSOs
- Tests


int2001:
checked. works.
- Live
- Post
- Search

AndreasK79:
Tested working live/post/search.

PR #2699, opened 2023-11-16
We should only check directories for being writable

phl0:
With https://github.com/magicbug/Cloudlog/pull/2675 we implemented a recursive permission check. The current code checks every file if it is writable by the web server. This leads to the case that debug info shows that directories are not writable:

Screenshot from 2023-11-16 09-37-42

Where the directories itself are writable indeed. But if the code finds a file that is not writable it returns false and pretends the directory is not writable which in fact is wrong.

So we need to only check directories and sub-directories for permissions but should ignore files here.


int2001 (approved):
yes. much better now

PR #2701, opened 2023-11-17
Add link to FG8OJ sked tool to hams.at listing

phl0:


int2001 (approved):
Great. i can "sked" now immediatly out of CL.
works

PR #2705, opened 2023-11-17
Fix comment quotation marks

### LINKED ISSUE #2704: Comments will get (partially) deleted when including quotation marks
Hi,

today I discovered that comments will get (partially) deleted after saving when the text including quotation marks. Here is an example:

First I enter the comment...
edit mode
...and save the QSO data set.

The comment is visible in viewing mode
viewing mode

But after reopening the QSO, the comment text partially disappeared.
disappeared comment

There seems to be a function that interprets quotation marks incorrectly.

I'm using Cloudlog 2.5.1

HB9HIL:
This should fix #2704 

Please check and test

### DIFF
--- a/application/views/qso/edit_ajax.php
+++ b/application/views/qso/edit_ajax.php
@@ -187,7 +187,7 @@
 
                                 <div class="form-group">
                                     <label for="comment">Comment</label>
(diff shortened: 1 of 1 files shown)


205er:
Bug fix works!

PR #2710, opened 2023-11-20
Fix QSO Start Time gets overwritten when editing a QSO in Live QSO Logging

### LINKED ISSUE #2703: Start Date/Time gets deleted when editing existing QSO
Hi all,

I found a bug, that is very annoying. I logged a QSO and afterwards I need to make some changes to it, so I edit the QSO and save my corrections. Unfortunately, the start date/time will then be set automatically to _0000-00-00 00:00:00_ (while the end date stays untouched)
Changed start date

Changed start date

This issue will only appear when changing an existing QSO from "Live QSO" (at the right side) 

I'm using Cloudlog 2.5.1

HB9HIL:
This should fix #2703 

Please test and confirm. Tested different QSO Edits and everything seems still working after removing those classes from the Form Input.

### DIFF
--- a/application/views/qso/edit_ajax.php
+++ b/application/views/qso/edit_ajax.php
@@ -55,12 +55,12 @@
                                 <div class="form-row">
                                     <div class="form-group col-sm-6">
                                         <label for="start_date">Start Date/Time</label>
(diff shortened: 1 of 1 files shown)


magicbug:
Works a great!

PR #2715, opened 2023-11-21
E-Mail Bugfixes

HB9HIL:
This is a bug-fixing PR that focuses on the email options. There were several bugs here which have now been fixed:

- SMTP encryption had no influence on the sending of emails, which caused problems with the setup for many users.
- The confirmation message after saving showed the set SMTP password as plain text, which is not nice from a security perspective.
- The confirmation message after saving always showed the same flashdata, namely the successful saving of the SMTP password. Even if this hasn't been changed.
- Some hints have been added to the form fields.
- As is often requested, there is now a button that triggers a test email. The email address that is stored in the account settings for the logged in user is used to send this mail to.

Thanks for testing and approving


HB9HIL:
This should fix issues in #2002

Issue #2706, opened 2023-11-17
Advanced Log: Search Dupes only worked for one time


dg9vh:
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the forum see readme for details**

**Describe the bug**
The button "Dupes" in "Advanced Log" only worked for exactly one time and now produces error-popup.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Logbook - advanced
2. Click on Dupes-Button
3. remove dupes by deleting them (the first found ones in the result-set of 250 logentries)
4. Do it a second time
5. from now on see the following error:
![Screenshot of the Error-Message "An Error occured while making the request](https://github.com/magicbug/Cloudlog/assets/13950650/81d4ceaf-f412-478f-9f8d-5861ca955123)

**Expected behaviour**
It should present a new list of dupes found. It seems, that it does not found any dupes and therefore throws this error.

**Screenshots**
![Output of the javascript-console](https://github.com/magicbug/Cloudlog/assets/13950650/540f6985-f038-4634-9a47-f1c8e808f2f1)

![more details](https://github.com/magicbug/Cloudlog/assets/13950650/8130a782-4d17-4bbb-b003-c3fa5fe758a6)

**Desktop (please complete the following information):**
 - OS: ChromeOS, but error is also reproduceable on Windows and Linux
 - Browser Chrome
 - Version 120.0.6099.25 

**Additional context**
If I should deliver more information, let me know.

Here are some lines from application/log

`
ERROR - 2023-11-17 17:00:57 --> 404 Page Not Found: Assets/js
ERROR - 2023-11-17 17:01:11 --> Severity: error --> Exception: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')

                        ORDER BY qsos.COL_TIME_ON desc, qsos.COL_PRIMARY_KEY desc
                        LIMIT 250' at line 12 /home/huebelon/public_html/dg9vh.de/cloudlog/system/database/drivers/mysqli/mysqli_driver.php 301
ERROR - 2023-11-17 17:01:12 --> Severity: error --> Exception: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')

                        ORDER BY qsos.COL_TIME_ON desc, qsos.COL_PRIMARY_KEY desc
[...]

PR #2720, opened 2023-11-23
Concat QSO IDs in PHP rather than SQL

phl0:
The group_concat function of mySQL is limited to 1024 chars (for old(er) DBs). So we should concat all dupe QSO IDs in PHP to prevent this.


phl0:
This addresses https://github.com/magicbug/Cloudlog/issues/2706.

int2001 (approved):
tested this. it works.
i don't know why - but it seems to be some kind of mySQL-Voodoo.
Good job. tnx!

PR #2725, opened 2023-11-24
Refactor LoTW imports

phl0:


phl0:
Simplified some checks (e.g. checking for empty username is obsolete because the previous SQL result provides only users with non-empty username) and corrected error handling. Errors are now displayed and loop is exited once an error occurred. The current code would allow two error messages to be concatenated in specific situations:

 
Screenshot from 2023-11-24 13-59-49

AndreasK79:
Looks good.

magicbug:
thanks Flo

Issue #2622, opened 2023-10-25
Unexpected Behavior for LOTW import qsl when using multiple calls


GregTroar:
**Describe the bug**
I use two different callsigns depending on my location or the radio I use, either XV9BPO or XV9Q.
I have created several different station locations with different call.

When I do an import of LOTW to get my confirmation it will only take into account the call which is defined in "Account".

So if I activate a location with my call XV9BPO while in account my call is XV9Q, the lotw import will always update only XV9Q QSL from LOTW not the XV9BPO.

In order to update the XV9BPO QSL from lotw I need to go in account and change my call from XV9Q to XV9BPO and then the lotw will work for XV9BPO.

I do not know if this is the expected behavior but I think it would be better that the LOTW is based on the callsign of the station location rather than the call defined in Account,

Hope my explanation is clear and if this is not considered a bug or if this is the normal behavior then you can close the issue but it will be a bit tricky then to update lotw qsl received when using different calls.

Thanks

HB9HIL:
Did you upload p12 Certs for all Callsigns, you created as Station Locations?

GregTroar:
you are right I did not upload p12 for XV9BPO.
So it works only if I change in account from XV9Q to XV9BPO but does not work if I keep XV9Q in the account but select the station location XV9BPO.
It seems to be expected as it is right now but not too convenient as I switch all the time between XV9BPO and XV9Q I would have to go to Account to change all the time between the two calls.

Thanks

HB9HIL:
> you are right I did not upload p12 for XV9BPO. So it works only if I change in account from XV9Q to XV9BPO but does not work if I keep XV9Q in the account but select the station location XV9BPO. It seems to be expected as it is right now but not too convenient as I switch all the time between XV9BPO and XV9Q I would have to go to Account to change all the time between the two calls.
> 
> Thanks

You have to upload p12 Certificates for all available Calls in the Station Locations. Cloudlog will automatically look for a (Call-) matching p12 Cert for the active Station Location. 

Choose XV9BPO as active Station Location. Add a p12 Certificate for this Call in the LotW Menu. Then it should work automatically without changing your Call in the Account Settings. 

Try that and give feedback please.

HB9HIL:
@GregTroar  Check this wiki Information Part "QSO Uploading" and "Background Processes"
https://github.com/magicbug/Cloudlog/wiki/LoTW-Import-&-Export-Documentation#qso-uploading

[...]

PR #2726, opened 2023-11-24
Allow for callsign selection upon LoTW download

phl0:
The GUI function for LoTW imports does not allow to select the callsign the confirmations should be downloaded for. So we add a dropdown and some code that allow to select a specific one or donwload for all user's callsigns if All is selected. The current code only downloads for logged-in user (whereas the cronjob downloads for all which is a bit of a difference).

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -746,9 +746,9 @@ class Lotw extends CI_Controller {
 			$lotw_url .= "&qso_qslsince=";
 			$lotw_url .= "$lotw_last_qsl_date";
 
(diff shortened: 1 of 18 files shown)


phl0:
This addresses https://github.com/magicbug/Cloudlog/issues/2622 (if the op confirms my assumptions re cron and GUI function).

phl0:
This obviously breaks cron sync to LoTW. Investigating ...

phl0:
False alarm. Cron still working as before. So ready for feedback here.

AndreasK79:
Tested working. Nice work.

PR #2730, opened 2023-11-25
QSL Confirmations

HB9HIL:
The QSL Confirmations in Awards for eQSL should be checked by default, same as QSL and LotW.
Same on Timeline.


phl0 (changes requested):
Tbh, I do not agree. At least for CQ, WAS and DXCC eQSL is not accepted. Therefore it was left out being checked.

HB9HIL:
Ah I didn't know that..

phl0:
No problem ✌️😇

PR #2731, opened 2023-11-26
Fix tooltips to new BS5 Syntax

int2001:
Tooltips are not working since syntax of BS5 has changed here.
This PR will fix that


int2001:
merging with @HB9HIL PR

PR #2732, opened 2023-11-26
RST in QSL Queue List

HB9HIL:
I added the RST (S) and (R) to the QSL Qeue List. This makes it easier if you write QSL Cards by hand. You don't have to open the QSO to check the RST.


int2001 (approved):
Works as expected

phl0 (approved):
Think this makes sense. Works as expected.

Issue #2696, opened 2023-11-15
Question: Development mode enabled by default?


bgritzmacher:
Is there a particular reason developer mode is hardcoded to be enabled? The information provided by it is not particularly useful to end-users unless you are actively debugging the application.

Best practices would be to use the CodeIgniter default that has been commented out - which is to set server variables to control its status, while assuming production mode by default. It is then up to the developer to make sure that their environment is setup correctly to do development tasks (even if it is just hardcode it for themselves.)

This is more annoying in Docker which requires one of numerous possible steps to disable:
- Install a text editor such as `nano` or `vim` within the container to edit the index.php file. This goes against docker best practices.
- `docker cp` the index.php file out of the container into the host file system, make the modifications...
  - then `docker cp` it back into the container.
  - or use a bind mount to the host copy.
  - or build my own local image that apply the changes.
  - or any other possible alternatives, I did say "numerous possible steps"...

And then this has to happen any time you down and up the container, as the changes get lost from the writable layer (unless you build a local image with the change in it.)

All for something that should be default, and able to be controlled at docker runtime, without modifying files within the container.

Everything I've seen within the code seems to indicate that this is an intentional choice, but I do not understand why.

iu2frl:
I see the line you are mentioning, it is:
```
#define('ENVIRONMENT', isset($_SERVER['CI_ENV']) ? $_SERVER['CI_ENV'] : 'development');
define('ENVIRONMENT', 'development');
```

bgritzmacher:
Correct. The default way CodeIgniter sets the ENVIRONMENT variable via the `$_SERVER['CI_ENV']` variable was commented out and just hardcoded to always be "development." I could argue that CodeIgniter's defaulting to "development" when CI_ENV isn't set isn't preferable either, but this isn't CodeIgniter's support 😉

A further suggestion to improve on your pull request and make it more complete for Docker support would be to add support for setting the CI_ENV value at runtime (either in `docker compose` in an `environment` section or by using `docker` with the `-e` switch.) This could then get set using SetEnv in a config file that gets loaded by apache in the /etc/apache2/conf-enabled folder.

[...]

PR #2733, opened 2023-11-26
Changing default environment to production

iu2frl:
Fixing #2696 

For future reference, if development or testing mode is needed, all you need is to set the `CI_ENV` variable in apache2 by creating a file called `.htaccess` in the root of cloudlog (where the main `index.php` is) containing:

```
SetEnv CI_ENV development
```

And then refresh the page, the new environment should be automatically set. Tested on Ubuntu 22.04

Documentation: https://www.codeigniter.com/user_guide/general/environments.html

### DIFF
--- a/index.php
+++ b/index.php
@@ -53,8 +53,8 @@
  *
  * NOTE: If you change these, also change the error_reporting() code below
  */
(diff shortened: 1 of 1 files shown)


magicbug:
Currently, I'm not keen to change this as it will break 100% of installs requiring them to run Git commands, might be prepared todo it further down the road but not yet.

iu2frl:
Makes sense!

PR #2734, opened 2023-11-26
Add user-selectable 'Name' column

DJ3CE:
This commit adds a user selectable 'Name' column to the 4 (5) user-configurable columns. There had been an 'operator'-option, which shows the operator name of the "own station", so with this change, there's also a way to show the remote operators name.

I put the option behind the 'operator'-entry and reused the 'general_word_name'-localization field.

Feedback appreciated :).


HB9HIL (changes requested):
option should be also added in `application/views/user/create.php`

DJ3CE:
Thank you for the hint. Found it, added it :).

HB9HIL:
Yeah my bad. said create.php but it is obviously add.php 😆

HB9HIL (approved):
Looks good to me. Works as expected.

int2001 (approved):
Works as expected. Nice one!

phl0 (approved):
Tested and works (both for editing a user and creating a new user).

PR #2738, opened 2023-11-27
Quick lookup bs5 fix

HB9HIL:


HB9HIL:
the form-select was doubled in some places. I removed them.

PR #2748, opened 2023-11-29
[MAP Custom] custom qso spot 

abarrau:
Version 1 of this new proposal functionality : 
* on user editor page (current user only) ; not available for other user (data are save on user_options, and this table use only  curent user)
* user can : 
   * add icon (house) for curent active station , and change color
   * change icon / color of QSO (default is the same like now (red spot)
   * change icon / color of QSO confirmed (qsl / lotw / eqsl)
   * active display gridsquare

If user not configure this feature (user options table empty for this feature), default spot is the red spot

this change is available for : 
* map on dashboard
* map on custom view
* map on qso logbook

if this feature is validate, i will make evolution for have color by band / mode (like on PSKreportor)

**some picture :** 
- edit user by default
config by default
- icon list for "station"
icon list for home
- different selection for 3 icon
config for 3
- default map view (with station, and confirmed qso display in default view)
map red dot
- view with qso confirmed colored
map with qso confirmed

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -145,6 +145,8 @@ class Dashboard extends CI_Controller {
 		echo "{\"markers\": [";
 		$count = 1;
 		foreach ($qsos->result() as $row) {
+			// check if qso is confirmed //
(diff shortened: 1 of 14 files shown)


abarrau:
i think i will never arrived with merge ... :'( ... again a mistake when i would like make correction for conflit general.css :(
i close it ...

PR #2749, opened 2023-11-30
Fix various errors in gridmap (in context of removal of old maps)

phl0:
Fixes various errors in gridmap:

- Hard coded homegrid removed for visitor map
- Added non-breaking space in coord divs on gridmap and activated gridmap
- Repair activated gridmap (needed to take care of newly introduced visitor variable)


int2001 (approved):
Confirm. Activators-Map is working again with this patch. tnx!

Tested: Activators Map and Visitors Map

PR #2755, opened 2023-11-30
[MAP Custom] custom qso plot (renew)

abarrau:
Version 1 of this new proposal functionality : 
* on user editor page (current user only) ; not available for other user (data are save on user_options, and this table use only  curent user)
* user can : 
   * add icon (house) for curent active station , and change color
   * change icon / color of QSO (default is the same like now (red spot)
   * change icon / color of QSO confirmed (qsl / lotw / eqsl)
   * active display gridsquare

If user not configure this feature (user options table empty for this feature), default spot is the red spot

this change is available for : 
* map on dashboard
* map on custom view
* map on qso logbook

if this feature is validate, i will make evolution for have color by band / mode (like on PSKreportor)

**some picture :** 
- edit user by default
config by default
- icon list for "station"
icon list for home
- different selection for 3 icon
config for 3
- default map view (with station, and confirmed qso display in default view)
map red dot
- view with qso confirmed colored
map with qso confirmed

### DIFF
--- a/assets/js/leaflet/leafembed.js
+++ b/assets/js/leaflet/leafembed.js
@@ -1,28 +1,17 @@
-
-
 var map;
-var ajaxRequest;
-var plotlist;
 var plotlayers=[];
+var iconsList={'qso':{'color':'#FF0000','icon':'fas fa-dot-circle'}};
 
-var greenIcon = L.icon({
-    iconUrl: icon_dot_url,
-    iconSize:     [10, 10], // size of the icon
-});
+var stationIcon = L.divIcon({className:'cspot_station'});
+var qsoIcon = L.divIcon({className:'cspot_qso'}); //default (fas fa-dot-circle red)
+var qsoconfirmIcon = L.divIcon({className:'cspot_qsoconfirm'});
+var redIconImg = L.icon({ iconUrl: icon_dot_url, iconSize: [10, 10] }); // old //
 
 var osmUrl = $('#leafembed').attr("tileUrl");
 
-function initmap(ShowGrid = 'No', MapTag = 'map') {
-    // set up AJAX request
-    ajaxRequest=getXmlHttpObject();
-    if (ajaxRequest==null) {
-        alert ("This browser does not support HTTP Request");
-        return;
-    }
-    
+function initmap(ShowGrid='No', MapTag='map', options={}) {
     // set up the map
     map = new L.Map(MapTag);
-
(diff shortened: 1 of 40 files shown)


assets/js/leaflet/leafembed.js
```diff
@@ -1,28 +1,17 @@
...
 
 var osmUrl = $('#leafembed').attr("tileUrl");
 
-function initmap(ShowGrid = 'No', MapTag = 'map') {
-    // set up AJAX request
-    ajaxRequest=getXmlHttpObject();
-    if (ajaxRequest==null) {
-        alert ("This browser does not support HTTP Request");
-        return;
-    }
```
int2001, line 21: Don't know if that is a good idea to remove this.
I think it was there for a reason. It checked if Browser was/is up2date and throws an error if the browser doesn't support ajax-requests.

abarrau:
> i really like the view. Great.
> 
> But: Did you check the visitor-View (when calling the public-slug without login) - seems like this PR breaks the View.

No tested ... i just make test and is KO :( but i know how fix it 

During my test, i find a core error ; the javascript visitor variable not defined 
js error

visitor variable is defined in php (true/false), but not in javascript.
I will propose a fix

abarrau:
I add a test for the visitor variable (considered the PR2757 is ok)
Now the plot is ok on visitor view
visitor can see only plot on "default" (red plot), and can not see the station position

-- user view : 
user view

-- visitor view :
visitor view

int2001:
> visitor variable is defined in php (true/false), but not in javascript. I will propose a fix

Yes, we noticed that, too.
There's a little Work in Progress @phl0 - maybe you wait with your fix until the visitor (and some things under that) is fixed. I see pain coming up due to possible Merge-Conflicts.

phl0:
> > visitor variable is defined in php (true/false), but not in javascript. I will propose a fix
> 
> Yes, we noticed that, too. There's a little Work in Progress @phl0 - maybe you wait with your fix until the visitor (and some things under that) is fixed. I see pain coming up due to possible Merge-Conflicts.

The issue will be solved as soon as https://github.com/magicbug/Cloudlog/pull/2761 is merged.

phl0 (approved):
Tested and works OK for me.

HB9HIL (approved):
tested and works good. I really like this one

int2001 (approved):

PR #2763, opened 2023-12-02
[MAP] reduce highlight map line on dark theme

abarrau:
This fix is available for page : 
- qso edit
- analytics (gridsquare, activated, custom)

For the 4 dark theme. 

Example before : 
global map KO
qso edit map KO

After : 
global map OK
qso edit map OK

By the way, i think we can reduce the ID map at only 3 (map, qso edit, gridsquare), ... but it's a other exchange...

### DIFF
--- a/assets/css/blue/overrides.css
+++ b/assets/css/blue/overrides.css
@@ -47,7 +47,10 @@ path.grid-worked {
 }
 
 #map,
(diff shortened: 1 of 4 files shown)


HB9HIL (approved):
Looks good to me 👍

PR #2769, opened 2023-12-04
Version Dialog

HB9HIL:
This is the Version Dialog we talked about. 

It's not finished yet but main functionality is give. You can test and give some feedback. 

When I undraft it I'll propose a detailed description of the functionality.


HB9HIL:
The Version Dialog is now ready for review/testing. 

version-dialog-on-dashboard

Basic Functionality:

The Version Dialog shows up to every user after the next release. It uses already existing datatables 'user_options' and 'options', so no migration is needed. 

The Version Dialog shows the latest release info's fetched directly from github.

If the user reloads the page the first time the needed rows are automatically generated in the database (for the logged in user). Because the code is placed in the footer to show up the dialog on every page we need to autoload the user_options_model, which becomes handy in future to use this for further user options. It reduces code in the controllers.

The admin of a multiuser instance has some settings for this version dialog

version-dialog-admin-menu

There also is an option to set a custom text within the version dialog. (release notes, custom text or both)

custom-text

vd-with-custom-text

The admin also have the possibility to trigger the version dialog for all users again if there is the need for an important message to all users. 

If users clicked on "Don't show again", they still have the option to see the version dialog by clicking on the menu point

vd-menu-button

Please test and give some feedback

PR #2771, opened 2023-12-04
Naming of subdivisions

phl0:
Currently the COL_STATE column is always named "State" which only is correct for DXCC 291 USA. This is a try to outsource this into a lib lookup function depending on DXCC.


AndreasK79:
Tested working here.

PR #2772, opened 2023-12-04
[QSO] Fix TimeOff on qso post

abarrau:
2 fix : 
1/ check QSO time off not lower than time off (accepted only for time on = 23:xx and time off = 00:xx)
- check doing on submit form (in JS)
timeoff less timeon

2/ when  time on = 23:xx and time off = 00:xx, change day before insert bdd value
- Before save (no error, timeoff < timeon accepted here)
add day if timeoff in 00:xx
- result on edit page 
1 day as add

Tested in : SLFE and Contest (not regression)
2nd fix NOT Tested in "live QSO" ... (time can be changed)


abarrau:
last comit fix a problem view with icon near qrz field
-- before 
prblm

-- after
fix

HB9HIL:
I still  can add QSO, if time off is less then time on

timeonoff

HB9HIL (changes requested):
please use the new bootstrap 5 classes "ms", "me", "mt" and "mb" for margin instead css commands in general.css

abarrau:
> please use the new bootstrap 5 classes "ms", "me", "mt" and "mb" for margin instead css commands in general.css

hello, done

HB9HIL (approved):
works for me. @phl0 could you check that too?

AndreasK79:
@abarrau good work. Merge time :)

PR #2773, opened 2023-12-05
Fix LBA Issue where POTA causes Problems

int2001:


phl0 (approved):
Works as expected.

AndreasK79:
Tested working as it should.

PR #2779, opened 2023-12-05
Ans qso search qrz

abarrau:
on QSO page, the lookup qrz return some qso with not exactly le same callsign
exemple : 
F1AB
F1AB/P
EA/F1AB/P

F1ABC is return , but only F1AB is lookup

 With this proposal PR, lookup return only qso with the real callsign (here F1AB)
OK F1AB
OK EA/F1AB

On line 988 : 
not the same fix ; but it's an other fix on same file : variable $callsign not exist in this function, only $id is know.


HB9HIL (approved):
works well

PR #2783, opened 2023-12-06
Fixed Bug at "Show/Hide Version Info for all"

int2001:


HB9HIL (approved):
thx for that fix. was my bad :)

Issue #2784, opened 2023-12-06
Issue logging past QSO when pushing reset button, clock starts running 


dk5cr:
**Describe the bug**
When opening QSO -> Log past QSO (in German "zeitversetztes QSO", second entry in the menu) you can click on the clock-icon to adjust the time to the current time. When you with to cancel the entries in the fields and click on the reset-button, then the clock starts running with hh:mm:ss just like in live-logging. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to QSO -> past QSO logging
2. Click on "cancel" (left of submit) 
3. have a look at the running clock in the time box above 

**Expected behaviour**
I expect to see the clock set to the current time in hh:mm without starting running in seconds like in live logging QSO. 

**Desktop (please complete the following information):**
 - OS: Armbian on BananaPi 
 - Browser Firefox 
 - Version 2.5.2

phl0:
Fix proposed in https://github.com/magicbug/Cloudlog/pull/2785. Please test and report.

PR #2785, opened 2023-12-06
Only reset timers in live QSO mode

phl0:
Fixes https://github.com/magicbug/Cloudlog/issues/2784.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1146,7 +1146,7 @@ $(document).on('keypress',function(e) {
 
       if ( ! manual ) {
         $(function($) {
-           resetTimers();
+           resetTimers(0);
         });
       }
     });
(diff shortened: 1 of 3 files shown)


HB9HIL (approved):
confirm the bug is fixed :)

int2001 (approved):
fixed and working

HB9HIL:
> Thanks for the incredible speedy bug fix. How can i fix it in my installation? running ./update_cloudlog.sh? Or do i have to replace some source code? Cheers and 73

If you don't want to change the branch you have to wait for the next release..

phl0:
> Thanks for the incredible speedy bug fix. How can i fix it in my installation? running ./update_cloudlog.sh? Or do i have to replace some source code? Cheers and 73

Depends on which branch you are using. If you are on dev just wait for the PR to be merged. If you are on master you have to wait for a release. And otherwise you would manually have to patch according to the PR. But this will probably break future updates then and then revert before updating.

PR #2792, opened 2023-12-07
Download QRZ-Confirmations to Cloudlog

int2001:
This one adds QRZ.COM Download to Cloudlog.

If QRZ.COM API-Key is set at station_location the "QSL" received at qrz.com will be shown in Cloudlog.
There are two ways to fetch those "QSLs".
* via cron: curl --silent http(s)://[cloudlog-url]/index.php/qrz/download
* via GUI at QRZ/Import

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 157;
+$config['migration_version'] = 158;
 
 /*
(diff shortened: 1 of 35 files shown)


magicbug:
Guess i should ask what this is fixing

magicbug:
OK just remember to add a full description to the PR so we know whats happening at a glance :)

HB9HIL (approved):
works awesome :)

PR #2795, opened 2023-12-07
[BS5] SimpleFLE and Tables

HB9HIL:
Returned to the old SimpleFLE Design before BS5


HB9HIL:
Fixed the active row backgrounds in the tables

PR #2806, opened 2023-12-10
Check file/dir permissions prior to LoTW download

phl0:
Fixes file permission issue reported by @HB9HIL (e.g. lotwreport_download.adi exists but is read-only).


AndreasK79:
Tested with download, and it works here.

HB9HIL (approved):
tested on production and works perfectly :)

PR #2807, opened 2023-12-11
Catch empty QRZ.com response when downloading qrz-confirmations

int2001:
Fixes small issue, when qrz-download is empty.


phl0 (approved):
Tested ok. 👍

PR #2812, opened 2023-12-12
Check for empty LoTW/eQSL dates and catch the error

phl0:
Catches a PHP (8) error if LoTW/eQSL date is empty but cnfm is set to 'Y'. Reported by @AndreasK79 though we do not know yet how the date can be empty ...

signal-2023-12-12-092836_002


abarrau:
Hi, i have no date on "COL_EQSL_QSLRDATE" before february 2023 ; only me ? 

Perhaps, this fix is proposal like this ; but the text of "gen_this_qso_was_confirmed_on" variable lang will be not ok ... 
```
                    <?php if($row->COL_EQSL_QSL_RCVD == "Y") { ?>
                    <h3>eQSL</h3>
                        <p><?php echo lang('gen_this_qso_was_confirmed_on'); ?> <?php if ($row->COL_EQSL_QSLRDATE != null) { $timestamp = strtotime($row->COL_EQSL_QSLRDATE); echo date($custom_date_format, $timestamp); } ?>.</p>
                    <?php } ?>
```

phl0:
> Hi, i have no date on "COL_EQSL_QSLRDATE" before february 2023 ; only me ?

I have. And a fix was provided to hide the text completely if the date is empty / null. We found out that the QSO @AndreasK79 identified was an edge case. Normally those dates are set when pulling LoTW/eQSL QSL info.

PR #2816, opened 2023-12-12
WIP: Generalize gridmaster functions

phl0:
Aim is to generalize the gridmaster stuff so that it does not need separate models for every country to add.


AndreasK79:
Tested working here.

phl0:
So undrafting here =). Tnx fer test.

PR #2818, opened 2023-12-12
New was map

AndreasK79:
Added new WAS map based on Leaflet.


phl0 (changes requested):
Works finde. Just needs a little correct re 51 states :)

AndreasK79:
Yeah, the map data includes DC, which counts for MD for WAS purposes.

PR #2821, opened 2023-12-13
Ten digit grids

phl0:
Consider up to ten digits/chars in maidenhead grid locator to calc position, bearing, distance from. Addresses https://github.com/magicbug/Cloudlog/discussions/2811. Please test and report.


int2001 (approved):
Tested:
* Gridmaps ✅
* Live-/Post-QSO ✅
* API ✅

Works as expected

PR #2825, opened 2023-12-13
Adif regex

AndreasK79:
Swapping out the old iterator to split adif line into array with a regex solution. 

Benefits: Splits on tag, speed.


int2001 (approved):
tested this with importing a 350 QSO ADIF on my dev. it's working.
tested with adding QSO via API - working too

phl0 (approved):
Tested:
- importing errorneous ADIF file ✅
- live / post QSO logging ✅
- emptied and re-imported 20k QSOs ✅

magicbug:
ADIF changes always make me nervous but on first look this seems OK

AndreasK79:
@magicbug I know. I have tested it a lot, but you never know. At least it's nothing to merge for this release. I think @phl0 and @int2001 will give it some testing on their multiuser setup.

int2001:
Tested a bit and found an issue regarding lotw.

Take a stable seat:
lotw delivers for example the state field as followed in its lotw_reports:
`<STATE:2>CA, California`
Yes. True story.

With the current parser version the "2" is ignored, and "CA, California" is pasted into COL_STATE. This violates - under certain circumstances - the maxlength at our database.

So, we need to work around that.

AndreasK79:
Thanks for testing @int2001. I have an idea for a fix. I will implement that, then we'll see.

AndreasK79:
@int2001 ready for a new test. Now fetches value according to number of chars.

int2001:
the issue with the oversized STATE Column when syncing from lotw is gone.

But another issue appeared when imporing eQSL-cfms:

`Exception: mb_substr(): Argument #3 ($length) must be of type ?int, string given application/libraries/Adif_parser.php 162`

PR #2830, opened 2023-12-14
Heal some strange behaviours of qrz-qsls

int2001:
Strange behaviours fixed with this patch:
* qrz.com replaces the slash (/) in worked calls with an underscore (_).
* qrz.com is not aware of submode. If one logs and confirms with LSB the LSB is placed in mainmode


phl0 (approved):
Tested successfully with @int2001

PR #2836, opened 2023-12-15
Prevent DataTable-Error when clicking Locator too fast

int2001:


phl0 (approved):
Works fine for me.

AndreasK79:
Tested working. Now no error, and only loads modal once.

PR #2839, opened 2023-12-15
Make consequent use of current_version instead of version at config-file

int2001:
config['app_version'] will never change from 1st installation.
So use the current_version from Database, whenever it is needed


phl0 (commented):
Works a treat on ADIF exorts. Great find.

PR #2843, opened 2023-12-15
[USER DESIGN] change design user page

abarrau:
Propose this PR for a new design for user edit page
- if ok, i report this on add user page

Changes : 
- 5 principal zone, can be elapsable
- reorganize bloc setting
- add help tooltips, click on "help icon" for show help info

full design

elapse design

help tooltips


HB9HIL (changes requested):
I'm not happy with this PR. This has several reasons:

- I like the Accordion like design and it organizes the whole page as it should. BUT you created completely new functions for that and added a huge amount of unnessesary code which is not needed. To redesign the page you don't need extra css, javascript or lines in the footer. Just use the new classes of Bootstrap 5. This is why we upgraded Cloudlog to the new version as this makes it much more simple to get a proper layout. 
For the accordion like design you can use the Bootstrap class `accordion` ([more info](https://getbootstrap.com/docs/5.2/components/accordion/#always-open)). 

- You removed the `<small>` hintes below the inputs and created tooltips. Problem here is tooltips are pretty bad for mobile use and since we would have enough space with the Accordion design I would rather like to see the hints as they were before. 

I like the way you redsigned it but the implementation doesn't make sense like this.

magicbug:
Spltting the account area up is a fab idea but as @HB9HIL says Bootstrap 5 the framework we use has suitable built in functions and tools to carry out the task, it might actually be better to turn the Account page into a more tabbed design, but the input text layout as it is now must stay.

magicbug:
If you aren't sure about bootstrap see https://getbootstrap.com/docs/5.1/getting-started/introduction/ we use a tabbed kind of layout for Analysis and some of the other pages like QSL

abarrau:
@magicbug @HB9HIL , ok i will work on this way ... (show small ... )
but for the vision between accordion or classic tab, with the classic tab we can't see on 1 page all information 
for eQsl, classic tab is normal, because it's different action
but on configuration page (here user), the user can see all information on same view .. 

other proposition, can we leave the choice to the user to choose the desired view?

73

abarrau:
So, finaly, 
- no help icon (stay the small text)
- param classified by "category" (view in accordion)
- 12 themes modified
- user only 1 file (edit) for view in add and edit
- add button on pwd field for view pwd when change it (if used can verify that he was write)

AndreasK79:
Looks good. Tested and working here.

HB9HIL:
Autoscroll commit is pending :) Standby

HB9HIL:
Ready for merge 👍

HB9HIL (approved):
Works and looks great :)

int2001:
Tested that for function and look'n'feel only.
Looks nice. i like.

magicbug:
Huge improvement thankyou

PR #2846, opened 2023-12-15
[Bugfix] HRDlog.net Upload doesn't work with Pre-/Suffixes

HB9HIL:
As the function which uploaded the qso to HRDlog sent the Station Callsign as "Callsign" in the Post Data the Login with the API key failed if the Station Call has a Pre- or Suffix. 

To get this working I had to add the HRDlog Username in the Station Location.


HB9HIL:
To prevent not working HRDlog Uploads in the User Installations and avoid an additional error message the idea is to auto populate the hrdlog_username from the station callsign if a HRDlog API Key is set since this should be in 99.99% the callsign (regarding to the logic of hrdlog).

int2001 (approved):
tested - works. good find

HB9HIL:
Fixed the Success and Error Message in HRDlog, QRZ and ADIF since they where not functional.

PR #2857, opened 2023-12-18
ADIF Import Tuning part II

int2001:
Reordered processing of incoming QSOs here. 1st check for dupes afterwards other logic


HB9HIL (approved):
more then 40% faster on adif reload on my machine. good job 👍

AndreasK79:
Purely dupechecking on import (27k adif):
1. dev 2m 5s
2. pr 1m 32s

Nice improvement.

PR #2859, opened 2023-12-18
Sometimes JA prefs are one digit only...

phl0:
DK9JC reported that WAJA award fails on his machine. Analysis showed that some QSOs had state Nos < 10 with one digit only. But the code relies on two digits (i.e. 0-padded for < 10). 

photo_2023-12-18_23-48-49

The code fails to count 8 and 08 as the same state. So we need some padding here. Please test if this is compatible among the various databases.

### DIFF
--- a/application/models/Waja.php
+++ b/application/models/Waja.php
@@ -215,7 +215,7 @@ class WAJA extends CI_Model {
      * $postdata contains data from the form, in this case Lotw or QSL are used
      */
     function getWajaWorked($location_list, $band, $postdata) {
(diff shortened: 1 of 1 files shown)


phl0:
We are not sure where the error originates. The op has imported ADIFs from other logs. So that maybe the source. Pretty sure its not the LoTW report. Will analyze though.

DK9JC:
I checked and can confirm that all issues are solved and it works as it should.

Thank you for the fix.

AndreasK79:
Tested working. Same number of prefectures in this PR as DEV.

PR #2860, opened 2023-12-19
[THEME] Standardize get_theme function

abarrau:
get_theme function not used on edit_ajax.php file

Question for Team : 
Why realised an "if get_theme()" on edit_ajax and header ?   get_theme() return always something
OK for simplify too ? 

if get_theme


abarrau:
Error of branch .. SORRY

PR #2861, opened 2023-12-19
[THEME] Standardize get_theme function + fix display arrow on select list on dark theme

abarrau:
get_theme function not used on edit_ajax.php file

Question for Team :
Why an "if get_theme()" is make on edit_ajax.php and header.php view ? get_theme() return always something
OK for simplify (remove if) ?
if gettheme()


HB9HIL (approved):

HB9HIL (commented):
seems okay for me

PR #2864, opened 2023-12-19
[THEME] convert theme "wide" in option of theme

abarrau:
Propose this PR for use only list of principal theme 
The "wide" is converted as an options of theme on user account Settings.

list of theme with options

With this way, we can add other "option" for theme and add an overrides_xxx.css file on folder theme.

Remark : i am not change the content of css class for wide

Remark 2 : during testing this PR, if you came back on other branch, dont forget to change the content of "user_stylesheet" field on table users

Stay to do : change the qso/edit_ajax header, after the #2861 will be validated.


application/migrations/165_convert_wide_theme_as_option.php
HB9HIL, line 1: down function does not contain the user_stylesheet

magicbug:
While I appreciate the time that has gone into this, I would prefer to keep themes separate for maintainability.

Therefore this won't make it into Cloudlog.

AndreasK79:
@magicbug as the themes are 99% the same, I don't understand why this is any worse to maintain. This PR just adds an option for the theme. 

The only thing that should be fixed, is that is should check if a wide override exists, else you get a 404 if you choose wide for the theme, and it does not have it.

HB9HIL:
@magicbug Maybe it isn't that bad. We could reduce a significant amount of code and the 404 issue is a simple fix (only show available options for theme)

PR #2866, opened 2023-12-20
Internal: Improve loading of models/libs/etc.

int2001:
Instead of using $CI=_get_instance and so on, make use of CodeIgniters function to load models/libs/controllers/etc.
Memorymanagement is a little better, since Codeigniter checks if model/lib is already loaded.


phl0 (approved):
Works ok for me.

magicbug:
Seems OK will merge see if it breaks after a period of usage don't think it will :)

PR #2868, opened 2023-12-20
CAT: Add checks if keys exist before updating UI

DJ3CE:
This (small) patch introduces checks for the existence of 
* satname
* satmode
* prop_mode

in the CAT-information before setting the corresponding UI elements and thus prevents resets of UI elements filled by hand if the information is not given by the CAT-response. Frequency is left unchecked on purpose, as I *guess* that field is mandatory.

Feedback appreciated, i.e. if the new checks would better just check on not-null and not on empty-string.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1455,6 +1455,7 @@ $(document).on('keypress',function(e) {
 							  if (data.prop_mode == 'SAT') {
 								  text = text+' '+data.satname;
 							  }
+							  text = text + ')';
 						  }
 						  if(data.frequency_rx != null && data.frequency_rx != 0) {
 							  text = text+'<span style="margin-left:10px"></span><b>RX:</b> '+(Math.round(parseInt(data.frequency_rx)/1000)/1000).toFixed(3)+' MHz)';
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1373,6 +1373,23 @@ $(document).on('keypress',function(e) {
     <script>
     // Javascript for controlling rig frequency.
 	  var updateFromCAT = function() {
+      var cat2UI = function(ui, cat, allow_empty, allow_zero, callback_on_update) {
+        // Check, if cat-data is available
+        if(cat == null) {
+          return;
+        } else if (typeof allow_empty !== 'undefined' && !allow_empty && cat == '') {
+          return;
+        } else if (typeof allow_zero !== 'undefined' && !allow_zero && cat == '0' ) {
+          return;
+        }
+        // Only update the ui-element, if cat-data has changed
+        if (ui.data('catValue') != cat) {
+          ui.val(cat);
+          ui.data('catValue',cat);
+          if (typeof callback_on_update === 'function') { callback_on_update(cat); }
+        }
+      }
+
 		  if($('select.radios option:selected').val() != '0') {
 			  radioID = $('select.radios option:selected').val();
 			  $.getJSON( "radio/json/" + radioID, function( data ) {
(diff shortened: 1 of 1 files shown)


phl0 (changes requested):
Works ok for me. But I guess we should add the check to frequency (and band) as well to be consistent. Just like RX band / frequency.

phl0 (approved):

DJ3CE:
May I add a very minor (not really related but not unrelated as well) fix?

There was/is a closing parentheses missing in the radio_cat_state info box, if the trx sends a prop_mode, which I include in the following commit... :)

phl0 (changes requested):
There is a closing bracket a few lines up. But that is only shown if RX frequency is set. So I would rather add an 'else' there:

```
diff --git a/application/views/interface_assets/footer.php b/application/views/interface_assets/footer.php
index 70ffb28a..2f9fb2b6 100644
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1450,6 +1450,8 @@ $(document).on('keypress',function(e) {
                                                  }
                                                  if(data.frequency_rx != null && data.frequency_rx != 0) {
                                                          text = text+'<span style="margin-left:10px"></span><b>RX:</b> '+(Math.round(parseInt(data.frequency_rx)/1000)/1000).toFixed(3)+' MHz)';
+                                                 } else {
+                                                         text = text + ')';
                                                  }
                                                  if (! $('#radio_cat_state').length) {
                                                          $('#radio_status').prepend('<div aria-hidden="true"><div id="radio_cat_state" class="alert alert-success radio_cat_state" role="alert">'+text+'</div></div>');
```

DJ3CE:
I see. Thanks for pointing that out :). That makes for another 'quirky' behaviour, I guess: Assume empty `prop_mode` and non-empty `frequency_rx`, then there's an closing bracket without an opening one now, with your fix there'd be an closing one always, as those are non-nested if's.

I'd like to discuss two possible solutions:
1. Append the opening/closing parenthesis to `text` in a separate `if`-clause, before and after the `prop_mode`- and `frequency_rx`-parts, with the `if` asking if either one was true.
2. Introduce a new variable for the text inside the parenthesis, maybe `text_p` and if that is non-empty, add it to `text` with the appropriate parenthesis (` if (text_p != '') { text = '(' + text_p + ')' }`).

I'd tend to the second solution, seems cleaner to me - but I'm happy to read about other approaches as well :).

DJ3CE:
Had a second (or third) thought on the CAT-based updates of UI elements, which resulted in a second branch (see [check-cat-ui-v2](https://github.com/DJ3CE/Cloudlog/tree/check-cat-ui-v2)), which consists of two commits, one fixing cat-to-ui-behaviour introducing a function for the handling to clean up the code, and one for the parenthesis thing.

Let me know if I should open a new PR for it or merge it into this PR.

magicbug:
@phl0 whats the feeling good for merge? just tried it here and seems OK but second set of eyes never hurts

DJ3CE:
Thanks for asking. I have added another two commits for

a) correct appearance of parentheses (correction of my correction and fix of the issue),
b) a better approach to cat based UI updates, which is now outsourced into a function.

The function does the checking of values (not null, not empty, not 0) and updates the UI element if the cat value has changed, ie. a manually entered value is only overridden if the cat value changes. For this, the cat-value is not only put into the UI's value, but also stored in an data-item.

Please test carefully if this matches your expectations.

I'm rather unhappy with the commit history, so maybe, we better close this PR and I'll open a cleaner one?

magicbug:
Sure, submit a new one and we can do a final check

PR #2874, opened 2023-12-23
[SimpleFLE] QRG Bug

HB9HIL:
This solves a bug, which made it impossible to write a line with only QRG and mode (in the same pattern as Band and Mode)

```
3.753 lsb
``` 
instead of 
``` 
80m lsb
```

Now it's possible. Important is, that `freq` set to `0` if mode changes but not band.  I tried to solve that with `prevMode` Variable. 

Here a Text to test this behaviour.

```
*example-data*
3.573 lsb
1212 m0abc okff-1234
3 hb9hil
4 ok1tn
20 dl6kva 7 8
5 dl5cw
day ++
cw
32 ok7wa ol/zl-071 5 8
33 ok1xxx  4 3
14.231 ssb
2134 2m0sql
6 la8aja 47 46
date 2021-05-14
40m 
40 dj7nt
day ++
cw
df3et
```

### DIFF
--- a/assets/js/sections/simplefle.js
+++ b/assets/js/sections/simplefle.js
@@ -109,6 +109,7 @@ function handleInput() {
 
 	var extraQsoDate = qsodate;
 	var band = "";
+	var prevMode = "";
 	var mode = "";
 	var freq = "";
 	var callsign = "";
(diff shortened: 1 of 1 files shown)


int2001 (approved):
wks - good find!

PR #2876, opened 2023-12-23
CAT: Improve update of UI values and fix parenthesis in radio-status

DJ3CE:
This PR is the successor to the unmerged-closed PR #2868 and contains two cat related fixes:

a) Correct appearance of parenthesis in the radio-status box by introducing a new variable for the contents,
b) a better approach UI updates from CAT-infos, which is now outsourced into a function.

The function does the checking of values (not null, not empty, not 0) and updates the UI element if the cat value has changed, ie. a manually entered value is only overridden if the cat value changes. For this, the cat-value is not only put into the UI's value, but also stored in an data-item. By supplying a callback function you can trigger additional actions on an performed update (the RST update on mode-change, e.g.).

Thanks to @phl0 for pointing out issues with my first solution.

Feedback appreciated.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1442,15 +1442,17 @@ $(document).on('keypress',function(e) {
 						  if(data.power != null && data.power != 0) {
 							  text = text+'<span style="margin-left:10px"></span>'+data.power+' W';
 						  }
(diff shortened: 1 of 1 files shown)


phl0 (approved):
Tested and looks fine to me :)

PR #2879, opened 2023-12-26
[OQRS] Fix and Removal

HB9HIL:
I fixed the callsign search field and

I removed the Station Name Column in the grouped results. 

This is IMHO not needed and just confuses the ham which is requesting a QSL Card. The Ham see the Callsign which corresponds to the Station Callsign.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 165;
+$config['migration_version'] = 166;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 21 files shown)


AndreasK79:
I don't agree that the name of the station is confusing. In my opinion, it helps clarify in which station you get a hit in the logsearch.

Let us say that the same callsign has been used several places, like SOTA/POTA/IOTA/WWFF, then it's good to show this information.

phl0:
So make the display optional / configurable?

AndreasK79:
I can agree on an option.

HB9HIL:
> Let us say that the same callsign has been used several places, like SOTA/POTA/IOTA/WWFF, then it's good to show this information.

@AndreasK79 I still don't see the use case because the QSL is related to Date, Time and Station Callsign. 

But I will add it as Option :) No problem

HB9HIL:
@AndreasK79 @phl0 

Now available as option :)

AndreasK79:
Tested working :)

PR #2880, opened 2023-12-27
Fixes query so that only qsos from your station_id is shown

AndreasK79:
The previous QSOs query in the logging window was broken, and fetched QSOs from other station id's


int2001:
This fixes a security-bug, we've overseen. Working HotFix. Tnx @AndreasK79

PR #2881, opened 2023-12-27
[SimpleFLE] removed unused line

HB9HIL:
Removed unused Lang Line.


int2001 (approved):
yep. it's never used. Tnx!

PR #2882, opened 2023-12-28
Sharpened the keypress-handler in footer.php

int2001:
Sharpened the keypress-handler, so it only reacts on searchfield and not - e.g. - on the NotesBox.

PS: Bug was found by @HB9HIL - so props to him ;)


HB9HIL (approved):
Thx! That was an annoying bug and prevented proper editing of Notes in Search -> QSO Edit

PR #2885, opened 2023-12-28
[QSO] Fav dropdown colorized

HB9HIL:
I colorized the Trash Icon and moved it to the right in the dropdown. It makes it a bit more intuitive. 

addresses #2884 

fav_dropdown


int2001 (approved):
Nice

PR #2886, opened 2023-12-29
Add "times worked" to QSO-Page

int2001:
At the qso-page you can see the last 5 QSOs, but you cannot see how often you worked the station.
This PR adds the information at the top of the table (See Screenshot)

image of qso-prev-window

QSO-Count is calculated on the BASE-Call.
e.g.: DF2ET counts "TF/DF2ET/p" as well as "DF2ET" and "DF2ET/mm"


HB9HIL (approved):
great add on!

PR #2889, opened 2023-12-30
Check VUCC_GRIDs and gridsquare for structual integrity at import

int2001:
Set COL_GRIDSQUARE and/or COL_VUCC_GRIDS to NULL, if field doesn't consists of the string which is !=0 at modulo 2.

Background: Some of my Users were able to  import QSOs with a gridsquare like '.' or the Call within the gridsquare field.
Thats invalid and causes problems at qrb-calculation.

perhaps @phl0 can also add a check at the qrb-calculator? This one only prevents importing.

At QSO-Page the check is already implemented. So only existing QSOs are affected

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -3275,6 +3275,32 @@ function lotw_last_qsl_date($user_id) {
 			  $cq_zone = NULL;
 		  }
 
+		  // Sanitise gridsquare input to make sure its a gridsquare
+		  if (isset($record['gridsquare'])){
+			  $a_grids=explode(',',$record['gridsquare']);	// Split at , if there are junctions
+			  foreach ($a_grids as $singlegrid) {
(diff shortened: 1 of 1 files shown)


magicbug:
Great idea to sanitise these fields, would be useful if there's an issue with either that it would show it to the user in the UI too

int2001:
> Great idea to sanitise these fields, would be useful if there's an issue with either that it would show it to the user in the UI too

in another PR. This function is also used by API-Calls and so on. And at API we simply "Null" it. Issue came up by one of my users, when logging via WSJT-X with "crappy" grids. 

i will do another PR, for the UI, when this one is working, and theres time for it ;)

magicbug:
As always thanks for the hard work!

AndreasK79:
Tested working by not inserting invalid grids to the database when importing and ADIF file.

PR #2890, opened 2023-12-30
Added time of "Worked before" in Contest-Section

int2001:
In some contest you are permitted to work a station again on the next day, or after a specified time.
To see the last-worked timediff quickly i added this patch.

Behaviour of everything else is still the same.

image of wked b4


magicbug:
Would it not make more sense as Worked before at [time]

int2001:
I think this is more flexible. Seen contests where you can work the other station after 24h and other contests where you are permitted to work them the next day. 

What about adding the last time_off, too?

phl0 (approved):

PR #2893, opened 2024-01-01
[Admin User Menu] Show Callsign in List and Option to send Passwort Reset Link

HB9HIL:
The first PR in 2024 🎉 Happy New Year to everybody 🎊

This introduces a slightly new admin user list. Not the same but similar to the Station Locations Table. 

Also new is the option to send a passwort reset link to an user. For the first time we use HTML E-Mails which are way more nicer than just plain text. 

email

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 166;
+$config['migration_version'] = 167;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/User.php
+++ b/application/controllers/User.php
@@ -804,6 +804,84 @@ class User extends CI_Controller {
 		}
 	}
 
+	function admin_send_passwort_reset() {
+
+		$this->load->model('user_model');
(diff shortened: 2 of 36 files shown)


HB9HIL:
Ready to test

HB9HIL:
I also added a "Last Login" Timestamp to the users database and in the users table. This is pretty handy to see the activity of the own users 

lastlogin

int2001:
it is working. Great feature. Makes it easier to onBoard new Users.

But: I would strongly recommend to NOT use HTML. HTML-Mails increase the Spam-Score.

Example:
`No, score=2.6 required=3.5 tests=BAYES_40,HTML_MESSAGE,`

HB9HIL:
As @int2001 mentioned I moved back to plain text E-Mail.

magicbug:
Yep plain text is always best for basic emails like resets obviously sometimes HTML does make sense when you need a table but pretty doesn't always win :)

electricEngineer:
Finally tried it out out and works like a charme!!! Thanks for your effort. Maybe the time of last PW reset would be better to know in the admin interface than last login...

PR #2899, opened 2024-01-01
fix dxcc search result

HB9HIL:


HB9HIL:
already fixed by @magicbug and in logbook_model not necessary

PR #2904, opened 2024-01-03
Rename existing HADES-D QSOs and reupload to LoTW

phl0:


DK9JC:
worked instantly and cfm all qso in lotw

thank you

PR #2906, opened 2024-01-03
WWFF Improvements

hugosilvaparagramanet:
Two minor adjustments on WWFF handling:
- on Advanced Log, show WWFF references in REF columns accounting for the possibility of both specific de/dx WWFF and sig fields are set. If both are set, priority is for the station WWFF_REF
- on ADIF export, handle both specific de/dx WWFF and sig fields. If both are set, priority is for the station WWFF_REF

### DIFF
--- a/application/libraries/AdifHelper.php
+++ b/application/libraries/AdifHelper.php
@@ -82,8 +82,6 @@ class AdifHelper {
             'SAT_MODE',
             'SAT_NAME',
             'SFI',
-            'SIG',
-            'SIG_INFO',
             'SILENT_KEY',
             'SKCC',
             'SOTA_REF',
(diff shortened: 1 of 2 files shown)


magicbug:
Looks good thanks @hugosilvaparagramanet

phl0:
I have a concern: With this "new" export we have WWFF refs twice in the export:

Screenshot from 2024-01-06 06-53-34

ADIF standard has a separate WWFF_REF field for this introduced in v3.13. See http://adif.org.uk/313/ADIF_313_annotated.htm. Where is the point to also set this in SIG/SIG_INFO (again)?

hugosilvaparagramanet:
The purpose was to support legacy apps, but you're right, there's no need for that. Apparently WWFF.co already looks for the new fields, so this is really unnecessary. What Cloudlog should do IMO is the opposite: if there's a SIG WWFF, export it as WWFF_REF.

I will confirm WWFF.co accepts the new fields and then do a new MR wich:
- removes the "also export as SIG"
- removes the "also export as MY_SIG"
- if SIG is WWFF, export as WWFF_REF
- if MY_SIG is WWFF, export as MY_WWFF_REF

PR #2907, opened 2024-01-03
Retry QRZ.com for compound callsigns

hugosilvaparagramanet:
When searching QRZ.com for a compound callsign, try again the base call if not found.

This area could use some more work, currently there are too many requests to QRZ.com for a single callsign.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1774,11 +1774,11 @@ class Logbook_model extends CI_Model {
      * Function returns all the station_id's with HRDLOG Code
      */
     function get_station_id_with_hrdlog_code() {
-        $sql = 'SELECT station_id, hrdlog_username, hrdlog_code 
+        $sql = 'SELECT station_id, hrdlog_username, hrdlog_code
                 FROM station_profile
-                WHERE coalesce(hrdlog_username, "") <> "" 
+                WHERE coalesce(hrdlog_username, "") <> ""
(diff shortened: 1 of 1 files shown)


phl0:
This is a qrz.com issue imho. The XML API does not work correctly if queried for a /P callsign. I queried: http://xmldata.qrz.com/xml/current/?s=SESSIONKEY;callsign=HB3YRZ/P and get:

Screenshot from 2024-01-06 07-42-24

Also using and URL encoded slash like http://xmldata.qrz.com/xml/current/?s=SESSIONKEY;callsign=HB3YRZ%2FP gives the same result. I did drop an email to the qrz devs back in Nov 2023 but no reply. So I just created a thread on their forum:

https://forums.qrz.com/index.php?threads/querying-api-for-p-callsigns-broken.895927/.

phl0:
This is a qrz.com issue imho. The XML API does not work correctly if queried for a /P callsign. I queried: http://xmldata.qrz.com/xml/current/?s=SESSIONKEY;callsign=HB3YRZ/P and get:

Screenshot from 2024-01-06 07-42-24


Also using and URL encoded slash like http://xmldata.qrz.com/xml/current/?s=SESSIONKEY;callsign=HB3YRZ%2FP gives the same result. I did drop an email to the qrz devs back in Nov 2023 but no reply. So I just created a thread on their forum:

https://forums.qrz.com/index.php?threads/querying-api-for-p-callsigns-broken.895927/.

PR #2920, opened 2024-01-14
In QSO logging, don't refresh callsign data if it's not changed

hugosilvaparagramanet:
When you are logging a QSO, if you put focus on the callsign field, and then blur without changing it, it refreshes the other station data (name, locator, etc.). It's redundant and it resets corrections the use has made to the callbook values.

This PR remembers the last checked value for callsign, and only does the needed deed if it's different from the last  version.


magicbug:
Thanks this caught me today

PR #2922, opened 2024-01-15
Trim VUCC_GRIDS Grids

phl0:
As discovered by @DK9JC (see https://x.com/DK9JC/status/1745134799566651543?s=20) it can happen that VUCC_GRIDS are stored with whitespace between fields. This fixes calculation of distance as well as removes inner whitespace upon storing and editing such QSOs.


magicbug:
Does indeed look like it fixes the issue

PR #2958, opened 2024-02-05
Copy xOTA refs upon QSO edit

phl0:
This addresses issue mentioned in https://github.com/magicbug/Cloudlog/discussions/2956. xOTA references are not copied to the QSO table upon editing a QSO.


magicbug:
This seems to function as expected Flo

PR #2960, opened 2024-02-05
Eqsl fix page

dg0tm:


magicbug:
Can you explain what this is fixing

dg0tm:
If no eQSL nickname is defined, a less attractive error page was rendered
301867373-986412c4-cb36-4e84-8305-153de96c712f

after that
eQSL-Nickname_not_defined
and I also add the same lines to make it clearer who belongs to what.

PR #2961, opened 2024-02-05
add stations_id to eqsl functions

dg0tm:


magicbug:
Can you explain what this is fixing

dg0tm:
I think that the check and update process should only work for the location I selected via the dropdown menu and not my entire logbook, but the underlying function ignores it so far.

PR #2967, opened 2024-02-08
visitor: Hide map when used without JavaScript (e.g. on qrz.com)

Manawyrm:
Hi,

this change makes CloudLog look a bit nicer when embedding it into a QRZ.com profile (which blocks JavaScript on iframes).
The map can't be rendered (due to the missing JS), thus let's just hide it away. 

Before:

After:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ When submitting PRs please make sure code is commented and includes one feature 
 
 ## Credits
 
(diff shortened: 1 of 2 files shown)


magicbug:
Would it not be better to use the widget feature instead of embedding the page?

Manawyrm:
Do you mean the `/widgets/qsos/[callsign]` URL? 

That doesn't seem to have any fancy CSS, no search function, no QSO and Country statistics.
I don't have any other content on my QRZ page, so having Cloudlog "in there" seems pretty nice to me 😸

PR #3009, opened 2024-03-13
Remove extra comma causing error

PartTimeLegend:


magicbug:
Out of interest what error is this triggering?

PartTimeLegend:
Updated base branch. I thought I had set it to dev so that's my fault.

When the trailing comma is there on PHP 7.4 I get an unexpected ) error. Removing the comma resolved that. Might just be my server though.

magicbug:
Interesting dont see that happening here but should be removed :D

PR #3025, opened 2024-03-24
Reduce calling scp, cache result

DJ3CE:
For callsign suggestions in logging, Cloudlog requested a list from the server with every additional character typed, from a minimum of 3 on.

This PR reduces calls to the server, by saving and filtering a (matching) list from the server for additional characters typed.

Example: You typed 'DJ3' and got a list of all callsigns. Now you add 'C' ('DJ3C') and while previously, the server was asked to give a new suggestion list with this PR the already received list is filtered for 'DJ3C'.

If the original request does not match the typed callsign anymore (Example: You replace the '3' in the previous example by a '4'), a new request is made.

Included in this PR is contest-, as well as normal-logging.

### DIFF
--- a/assets/js/sections/contesting.js
+++ b/assets/js/sections/contesting.js
@@ -10,7 +10,6 @@ $(document).ready(async function () {
 
 var scp_data = {
   request: "",
-  status: -1 , // -1 - never req, 0 - req in progress, 1 - request done
   data: [],
 };
 
@@ -214,11 +213,10 @@ $('#start_date').change(function () {
 
 // On Key up check and suggest callsigns
 $("#callsign").keyup(function () {
-	var call = $(this).val().toUpperCase();
+	let call = $(this).val().toUpperCase();
 	if (call.length >= 3) {
 
-        if (scp_data.status < 0 || scp_data.request != call.substr(0, scp_data.request.length)) {
-          scp_data.status = 0;
+        if ( scp_data.request == "" || ! call.startsWith(scp_data.request) ) {
           scp_data.request = call;
           scp_data.data = [];
           $.ajax({
@@ -228,18 +226,21 @@ $("#callsign").keyup(function () {
                   callsign: call
               },
               success: function (result) {
-                  scp_data.status = 1;
-                  scp_data.data = result.split(" ");
+                  var call_now = $("#callsign").val().toUpperCase(); // Might have changed while running the query...
+                  if (call_now.startsWith(call)) {
+                    scp_data.data = result.split(" ");
 
-                  var call = $("#callsign").val().toUpperCase(); // Might have changed while running the query...
-                  $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
-                  highlight(call);
+                    call_now = call_now.replace('0','Ø');
+                    $('.callsign-suggestions').text(filterCallsignList(call_now, scp_data.data));
+                    highlight(call_now);
+                  }
               }
           });
 
         } else {
           // Filter the already obtained list:
-          $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
+          call = call.replace('0','Ø');
+          $('.callsign-suggestions').text(filterCallsignList(call, scp_data.data));
           highlight(call);
 
         }
@@ -248,7 +249,7 @@ $("#callsign").keyup(function () {
 		var qTable = $('.qsotable').DataTable();
 		qTable.search(call).draw();
 	}
-	else if (call.length <= 2) {
+	else {
 		$('.callsign-suggestions').text("");
 	}
 });
@@ -686,4 +687,9 @@ function getUTCDateStamp(el) {
 	var localTime = now.getTime();
 	var utc = localTime + (now.getTimezoneOffset() * 60000);
 	$(el).attr('value', ("0" + now.getUTCDate()).slice(-2) + '-' + ("0" + (now.getUTCMonth() + 1)).slice(-2) + '-' + now.getUTCFullYear());
+}
+
+function filterCallsignList(call, list) {
+    let re = "(^|\/)" + call;
+    return list?.filter((el) => (el.search(re) !== -1)).join(' ') || '';
 }
--- a/assets/js/sections/contesting.js
+++ b/assets/js/sections/contesting.js
@@ -8,11 +8,6 @@ $(document).ready(async function () {
 	setRst($("#mode").val());
 });
 
-var scp_data = {
-  request: "",
-  data: [],
-};
-
 // Resets the logging form and deletes session from database
 function reset_contest_session() {
 	$('#name').val("");
@@ -212,47 +207,22 @@ $('#start_date').change(function () {
 });
 
 // On Key up check and suggest callsigns
-$("#callsign").keyup(function () {
-	let call = $(this).val().toUpperCase();
-	if (call.length >= 3) {
+$("#callsign").keyup(scp_keyup({
+  selector: $("#callsign"),
+  showSuggestions: function (call, text) {
+    $('.callsign-suggestions').text(text);
+    highlight(call);
+  }
+}));
 
-        if ( scp_data.request == "" || ! call.startsWith(scp_data.request) ) {
-          scp_data.request = call;
-          scp_data.data = [];
-          $.ajax({
-              url: 'lookup/scp',
-              method: 'POST',
-              data: {
-                  callsign: call
-              },
-              success: function (result) {
-                  var call_now = $("#callsign").val().toUpperCase(); // Might have changed while running the query...
-                  if (call_now.startsWith(call)) {
-                    scp_data.data = result.split(" ");
-
-                    call_now = call_now.replace('0','Ø');
-                    $('.callsign-suggestions').text(filterCallsignList(call_now, scp_data.data));
-                    highlight(call_now);
-                  }
-              }
-          });
-
-        } else {
-          // Filter the already obtained list:
-          call = call.replace('0','Ø');
-          $('.callsign-suggestions').text(filterCallsignList(call, scp_data.data));
-          highlight(call);
-
-        }
-		// moved to blur
-		// checkIfWorkedBefore();
-		var qTable = $('.qsotable').DataTable();
-		qTable.search(call).draw();
-	}
-	else {
-		$('.callsign-suggestions').text("");
-	}
+$("#callsign").keyup(function() {
+  const call = $(this).val().toUpperCase();
+  if (call.length >= 3) {
+	var qTable = $('.qsotable').DataTable();
+	qTable.search(call).draw();
+  }
 });
+
 
 function checkIfWorkedBefore() {
 	var call = $("#callsign").val();
@@ -689,7 +659,61 @@ function getUTCDateStamp(el) {
 	$(el).attr('value', ("0" + now.getUTCDate()).slice(-2) + '-' + ("0" + (now.getUTCMonth() + 1)).slice(-2) + '-' + now.getUTCFullYear());
 }
 
-function filterCallsignList(call, list) {
-    let re = "(^|\/)" + call;
-    return list?.filter((el) => (el.search(re) !== -1)).join(' ') || '';
+function scp_keyup(options) {
+  // options must have two keys:
+  //   * selector - element, with .val() which gives the entered callsign
+  //   * showSuggestions - function(call, text), where the text is
+  //                       the list of callsign-suggestions
+  const scp = {
+    request: "",
+    data: []
+  };
+  const callFromInput = (el) => el.val().toUpperCase().replace('0','Ø');
+  const checkCacheValid = (call) => (scp.request != "" && call.includes(scp.request));
+  const filterCallsignList = function (call) {
+      return scp.data?.filter((el) => (el.includes(call) == true)).join(' ') || '';
+  };
+  const updateSuggestions = function (call) {
+    const suggestions = filterCallsignList(call);
+    options.showSuggestions(call, suggestions);
+  }
+
+  const keyup = function(){
+	const call = callFromInput(options.selector);
+
+    if (call.length < 3) {
+      options.showSuggestions("", "");
+      return;
+    }
+
+    if ( checkCacheValid(call) ) {
+      updateSuggestions(call);
+      return;
+    }
+
+    // Cache invalid, so update it and reset suggestions
+    options.showSuggestions("");
+
+    scp.request = call;
+    scp.data = [];
+    $.ajax({
+        url: 'lookup/scp',
+        method: 'POST',
+        data: {
+            callsign: call
+        },
+        success: function (result) {
+            const call_now = callFromInput(options.selector);
+            if (checkCacheValid(call_now)) {
+              scp.data = result.split(" ");
+
+              updateSuggestions(call_now);
+            }
+        }
+    });
+
+  };
+
+  return keyup;
 }
+
--- a/assets/js/sections/contesting.js
+++ b/assets/js/sections/contesting.js
@@ -8,6 +8,12 @@ $(document).ready(async function () {
 	setRst($("#mode").val());
 });
 
+var scp_data = {
+  request: "",
+  status: -1 , // -1 - never req, 0 - req in progress, 1 - request done
+  data: [],
+};
+
 // Resets the logging form and deletes session from database
 function reset_contest_session() {
 	$('#name').val("");
@@ -208,20 +214,35 @@ $('#start_date').change(function () {
 
 // On Key up check and suggest callsigns
 $("#callsign").keyup(function () {
-	var call = $(this).val();
+	var call = $(this).val().toUpperCase();
 	if (call.length >= 3) {
 
-		$.ajax({
-			url: 'lookup/scp',
-			method: 'POST',
-			data: {
-				callsign: $(this).val().toUpperCase()
-			},
-			success: function (result) {
-				$('.callsign-suggestions').text(result);
-				highlight(call.toUpperCase());
-			}
-		});
+        if (scp_data.status < 0 || scp_data.request != call.substr(0, scp_data.request.length)) {
+          scp_data.status = 0;
+          scp_data.request = call;
+          scp_data.data = [];
+          $.ajax({
+              url: 'lookup/scp',
+              method: 'POST',
+              data: {
+                  callsign: call
+              },
+              success: function (result) {
+                  scp_data.status = 1;
+                  scp_data.data = result.split(" ");
+
+                  var call = $("#callsign").val().toUpperCase(); // Might have changed while running the query...
+                  $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
+                  highlight(call);
+              }
+          });
+
+        } else {
+          // Filter the already obtained list:
+          $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
+          highlight(call);
+
+        }
 		// moved to blur
 		// checkIfWorkedBefore();
 		var qTable = $('.qsotable').DataTable();
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -1,4 +1,9 @@
 var lastCallsignUpdated=""
+var scp_data = {
+  request: "",
+  status: -1 , // -1 - never req, 0 - req in progress, 1 - request done
+  data: [],
+};
 
 $( document ).ready(function() {
 	setTimeout(function() {
@@ -1019,18 +1024,31 @@ $("#callsign").on("keypress", function(e) {
 // On Key up check and suggest callsigns
 $("#callsign").keyup(function() {
 	if ($(this).val().length >= 3) {
-	  $('.callsign-suggest').show();
-	  $callsign = $(this).val().replace('Ø', '0');
-	  $.ajax({
-		url: 'lookup/scp',
-		method: 'POST',
-		data: {
-		  callsign: $callsign.toUpperCase()
-		},
-		success: function(result) {
-		  $('.callsign-suggestions').text(result);
-		}
-	  });
+	  $callsign = $(this).val().replace('Ø', '0').toUpperCase();
+
+      if (scp_data.status < 0 || scp_data.request != $callsign.substr(0, scp_data.request.length)) {
+        scp_data.status = 0;
+        scp_data.request = $callsign;
+        scp_data.data = [];
+        $.ajax({
+          url: 'lookup/scp',
+          method: 'POST',
+          data: {
+            callsign: $callsign
+          },
+          success: function(result) {
+            scp_data.status = 1;
+            scp_data.data = result.split(" ");
+
+            var call = $("#callsign").val().replace('Ø', '0').toUpperCase();
+            $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
+            $('.callsign-suggest').show();
+          }
+        });
+      } else {
+            $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith($callsign)).join(' '));
+            $('.callsign-suggest').show();
+      }
 	}
   });
 
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -1,9 +1,4 @@
 var lastCallsignUpdated=""
-var scp_data = {
-  request: "",
-  status: -1 , // -1 - never req, 0 - req in progress, 1 - request done
-  data: [],
-};
 
 $( document ).ready(function() {
 	setTimeout(function() {
@@ -1022,38 +1017,13 @@ $("#callsign").on("keypress", function(e) {
 });
 
 // On Key up check and suggest callsigns
-$("#callsign").keyup(function() {
-	if ($(this).val().length >= 3) {
-	  let call = $(this).val().toUpperCase();
-
-      if ( scp_data.request == "" || ! call.startsWith(scp_data.request) )  {
-        scp_data.request = call;
-        scp_data.data = [];
-        $.ajax({
-          url: 'lookup/scp',
-          method: 'POST',
-          data: {
-            callsign: call
-          },
-          success: function(result) {
-            var call_now = $("#callsign").val().toUpperCase();
-
-            if (call_now.startsWith(call)) {
-              scp_data.data = result.split(" ");
-
-              call_now = call_now.replace('0','Ø');
-              $('.callsign-suggestions').text(filterCallsignList(call_now, scp_data.data));
-              $('.callsign-suggest').show();
-            }
-          }
-        });
-      } else {
-            call = call.replace('0','Ø');
-            $('.callsign-suggestions').text(filterCallsignList(call, scp_data.data));
-            $('.callsign-suggest').show();
-      }
-	}
-  });
+$("#callsign").keyup( scp_keyup({ 
+  selector: $(this),
+  showSuggestions: function (call, text) {
+    $('.callsign-suggestions').text(text);
+    $('.callsign-suggest').show();
+  }
+}));
 
 //Reset QSO form Fields function
 function resetDefaultQSOFields() {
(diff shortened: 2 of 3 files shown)


patrickrb:
I think just adding a debounce might be a better solution:

```
// Debounce function
function debounce(func, wait) {
    var timeout;
    return function() {
        var context = this, args = arguments;
        clearTimeout(timeout);
        timeout = setTimeout(function() {
            func.apply(context, args);
        }, wait);
    };
}

// On Key up check and suggest callsigns
$("#callsign").keyup(debounce(function () {
    var call = $(this).val();
    if (call.length >= 3) {

        $.ajax({
            url: 'lookup/scp',
            method: 'POST',
            data: {
                callsign: $(this).val().toUpperCase()
            },
            success: function (result) {
                $('.callsign-suggestions').text(result);
                highlight(call.toUpperCase());
            }
        });
        // moved to blur
        // checkIfWorkedBefore();
        var qTable = $('.qsotable').DataTable();
        qTable.search(call).draw();
    }
    else if (call.length <= 2) {
        $('.callsign-suggestions').text("");
    }
}, 500)); // 500ms delay
```

DJ3CE:
Not sure if we want to solve the same problem ;). I'd go for trying to get suggestions as early and fast as possible and narrow them down as fast as possible, therefore it seems counterintuitive to do 'debouncing', but more natural to filter stored results.

patrickrb:
So based on the pr description, I thought we were trying to reduce the number of requests sent from the frontend to the backend as the user types.

> For callsign suggestions in logging, Cloudlog requested a list from the server with every additional character typed, from a minimum of 3 on.
> This PR reduces calls to the server,

The denounce solution is the standard go to way of preventing sending a request as a user types until they are done typing.

I think implementimg caching here is setting up for problems down the road.

There's an old joke in programming:

> There are only two hard problems in computer science: Cache invalidation, naming things, and off-by-one errors.

DJ3CE:
... but the cited sentence has some more words to it ;).

I'm totally fine with alternative solutions taking other priorities in server-load and user-experience, but I guess here I'd appreciate an eye on this particular approach - especially, as pointed out, since cache invalidation is not the easiest thing to do. Feel free to point out gaps I might have left.

The approach taken just uses a state-model representing if and for which characters a request is on its way.

DJ3CE:
... one thing to add: I guess there could be different solutions 'right' for 'normal' and for 'contest' logging.

Based on your screenshots you seem to focus on 'normal'-logging, while I started implementing this for contest-logging.

patrickrb:
I definitely see where you're coming from, and caching is a valid way of solving the state problem.   


Both solutions solve the problem, and can even work in tandem  together. 

With that said, in my personal experience, it would best to start with the debounce method at ~300-500ms delay (which isn't even noticeable to the end user, it's just enough to prevent multiple calls if the user mistypes and quickly fixes).

The main reasons to start with debounce over cache are:

1. Ease of Implementation: Debouncing can be implemented quickly, This makes it a low-effort, high-impact change.
2. Minimal Risk: Since debouncing is applied on the frontend and primarily affects how often the application makes requests to the backend, it introduces minimal risk to existing backend logic or data integrity.

If we implement debounce and you're happy with how it solves the problem, then I think it'd be the appropriate time to look at integrating some type of caching.

I hope this clears up where I'm coming from and the best practice for order of operations on a problem/solution like this.

magicbug:
For SCP you want to get results as fast as possible this obviously is different for the QSO box fields some debounce on those would be a great idea.

magicbug:
This PR does seem to have an issue for example with my callsign when I type 2M0 no SCP results ever appear but GM1 etc function as expected

assets/js/sections/contesting.js
```diff
@@ -208,20 +214,35 @@ $('#start_date').change(function () {
 
 // On Key up check and suggest callsigns
 $("#callsign").keyup(function () {
-	var call = $(this).val();
+	var call = $(this).val().toUpperCase();
```
patrickrb, line 217: if you use `let` instead of `var ` here you can re-use the `call` variable in the scope below instead of re-defining `call`

assets/js/sections/contesting.js
```diff
@@ -208,20 +214,35 @@ $('#start_date').change(function () {
...
-			method: 'POST',
-			data: {
-				callsign: $(this).val().toUpperCase()
-			},
-			success: function (result) {
-				$('.callsign-suggestions').text(result);
-				highlight(call.toUpperCase());
-			}
-		});
+        if (scp_data.status < 0 || scp_data.request != call.substr(0, scp_data.request.length)) {
```
patrickrb, line 220: A few things here, you should use the `!==` comparison so it doesnt try to do any type interpretation.

see here: https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

---

The logic check on `scp_data.status < 0` will only ever be true on page load?

---

> scp_data.request != call.substr(0, scp_data.request.length)

I think this can be cleaned up with `startsWith`

```
if (scp_data.request.startsWith(call)) {
  // Your code here
}
```

https://www.w3schools.com/jsref/jsref_startswith.asp

assets/js/sections/contesting.js
```diff
@@ -208,20 +214,35 @@ $('#start_date').change(function () {
...
+              url: 'lookup/scp',
+              method: 'POST',
+              data: {
+                  callsign: call
+              },
+              success: function (result) {
+                  scp_data.status = 1;
+                  scp_data.data = result.split(" ");
+
+                  var call = $("#callsign").val().toUpperCase(); // Might have changed while running the query...
```
patrickrb, line 234: I think shouldnt need to redeclare `call` here if you use `let` above`

assets/js/sections/contesting.js
```diff
@@ -208,20 +214,35 @@ $('#start_date').change(function () {
...
+                  $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
```
patrickrb, line 235: this will thow an error if `scp_data.data` is ever missing or falsey.  We could easily handle this scenario by doing this:

`$('.callsign-suggestions').text(scp_data.data?.filter((el) => el.startsWith(call)).join(' ') || []);`

We'll check if data exists on `scp_data` with an existential operator `(?)` and if it doesn't, we pass an empty arrow to the filter function.

assets/js/sections/contesting.js
```diff
@@ -208,20 +214,35 @@ $('#start_date').change(function () {
...
+                  highlight(call);
+              }
+          });
+
+        } else {
+          // Filter the already obtained list:
+          $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
```
patrickrb, line 242: same thing here, we should add an existential and a default `[]`

assets/js/sections/qso.js
```diff
@@ -1019,18 +1024,31 @@ $("#callsign").on("keypress", function(e) {
...
-		url: 'lookup/scp',
-		method: 'POST',
-		data: {
-		  callsign: $callsign.toUpperCase()
-		},
-		success: function(result) {
-		  $('.callsign-suggestions').text(result);
-		}
-	  });
+	  $callsign = $(this).val().replace('Ø', '0').toUpperCase();
```
patrickrb, line 1027: a few things:

1. isn't using `$` on the var a php thing? This is a js variable. 
2.  can it match the same variable name as the contesting.js file? 
3.  is the `.replace('Ø', '0')` only required for qso and not contesting?

assets/js/sections/qso.js
```diff
@@ -1019,18 +1024,31 @@ $("#callsign").on("keypress", function(e) {
...
+
+            var call = $("#callsign").val().replace('Ø', '0').toUpperCase();
+            $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith(call)).join(' '));
+            $('.callsign-suggest').show();
+          }
+        });
+      } else {
+            $('.callsign-suggestions').text(scp_data.data.filter((el) => el.startsWith($callsign)).join(' '));
+            $('.callsign-suggest').show();
+      }
```
patrickrb, line 1051: Same issues as above with this one.

assets/js/sections/contesting.js
DJ3CE: I chose to not reuse `call` on purpose. The moment the success-callback is executed, the value may have changed, and the 'cache' may be invalidated already, or only a subset of the 'cached callsigns' should be displayed. 

But yes, lets go with `let` here.

assets/js/sections/contesting.js
DJ3CE: :) Strangely enough, I used `startsWith` further down in the code. Thanks (or sorry for me being... well, distracted by other languages ;) ).

-1: Yes.

assets/js/sections/contesting.js
DJ3CE: As stated above and as in the comment, the value could have changed by the time, the success-callback is being executed.

assets/js/sections/contesting.js
DJ3CE: Your code returns an empty array for an empty string, where a string is expected (the `|| []` isn't passed to the `filter`). I guess `... || ""` is sufficient here.

assets/js/sections/contesting.js
DJ3CE: Actually I'm not sure anymore if this `startsWith` is correct: We may have prefixed callsigns in the result (type 'MM0' and have sth. like 'DL/MM0ASDF' in the results). I'll think about a regex.

assets/js/sections/qso.js
DJ3CE: The code just reuses existing variables. `$` is eligible in JS-vars (heavily used by jQuery, e.g. ;) ).

I'll have a look on the replace as it poses 'issues' :D.

DJ3CE:
> This PR does seem to have an issue for example with my callsign when I type 2M0 no SCP results ever appear but GM1 etc function as expected

Thanks for trying this out and pointing me to issues where '0' is part of the callsign. What is the convention about the use of '0' or 'Ø' in callsigns in Cloudlog?

assets/js/sections/contesting.js
patrickrb: youre right, but then if it has changed, shouldnt we just set the cache and not doing anything else?  I feel like if we move to the suggestions at that point there will be a mismatch from the call entered in the input and the results?

assets/js/sections/contesting.js
patrickrb: good point, maybe use `.includes`

```javascript
const string = "foo";
const substring = "oo";
console.log(string.includes(substring)); // true
```

assets/js/sections/qso.js
patrickrb: its a nit for sure, but i feel like we should try our best to be consistent with modern code patterns :)

patrickrb: @magicbug should the `.replace('Ø', '0')` function call also be moved to the contesting.js logic?  Or is it only necessary for regular QSO's?

DJ3CE: Please see @magicbug s comment about '2E0' and my response below.

patrickrb: i saw both, neither seemed directly related to my question regarding whether the logic belonged in both places

DJ3CE:
There you go. Thank you for input and testing, appreciate it.

* @magicbug callsigns including '0' now work :). The returned list hast stroked-0 in it, thus the filtering failed (miserably ;) ),
* instead of `String.includes()` a regex is used, to have prefixed callsigns in the list while not being too coarse,
* `scp_data.status` got discarded,
* 'rough' network conditions are handled better by asking if a server-response still matches the currently entered part-of-a-callsign

assets/js/sections/qso.js
DJ3CE: I considered it 'wip' with Peter's comment already... :)

assets/js/sections/contesting.js
```diff
@@ -208,26 +213,43 @@ $('#start_date').change(function () {
...
+          scp_data.request = call;
+          scp_data.data = [];
+          $.ajax({
+              url: 'lookup/scp',
+              method: 'POST',
+              data: {
+                  callsign: call
+              },
+              success: function (result) {
+                  var call_now = $("#callsign").val().toUpperCase(); // Might have changed while running the query...
```
patrickrb, line 229: 1) I still dont see the benefit of updating this since the returned data in `result` will potentially have nothing to do with the updated call sign.  I think a more nuanced approach might be to cancel the existing ajax request if the user changes the input before the request has completed.

2)  I still don't understand why we have `scp_data.request == ""` check? It only return `true` the first request, i dont see the benefit of it ever being `false`?

3) should we go ahead and update the `startsWith` to `includes`?
 

I havent tested it, but i think something like this should work:

```javascript
$("#callsign").keyup(function () {
    let call = $(this).val().toUpperCase();
    if (call.length >= 3) {
        // Check if a previous request exists and abort it
        if(scp_data.previousRequest) {
            scp_data.previousRequest.abort();
        }

        if (scp_data.request == "" || !call.startsWith(scp_data.request)) {
            scp_data.request = call;
            scp_data.data = [];
            // Store the current request in scp_data
            scp_data.previousRequest = $.ajax({
                url: 'lookup/scp',
                method: 'POST',
                data: {
                    callsign: call
                },
                success: function (result) {
                    // When the request is complete, nullify the previousRequest to clean up
                    scp_data.previousRequest = null;
                }
            });
        }
    }
});
```

DJ3CE: 1.) You type "2M0" and the request is fired. While the request takes place, you enter "S", and as "2M0S" starts with "2M0" no new request is fired. The time the success-callback is executed `call_now="2M0S"` and the correct subset of the delivered data is shown.

2.) Every string starts with "" (try it out ;) ), so we need a starting point. An alternative would be to initialize `scp_data.request` with some magic value...

3.) Why?

4.) `scp` expects '0' in the query, but returns 'Ø' in the result (have a look in the code). Therefore this should be right - maybe @magicbug can comment on this.

Code example: Sure, there are many ways to go.

patrickrb: 1) what if i quickly press ctrl + a to select everything and start typing `WA6`?
2) I dont understand, a string cant start with an empty string.   `scp_data.request == ""` is specifically checking if that string is empty, not if its undefined, not if its false, only empty string.
3) I played with it, you're right..dont need it.
4) If that's the case, why do we only have the logic on QSO.js and not contest.js?

DJ3CE: 1.) I'd suggest to try it out or to think about what-happens-when in the code, like I did in the previous example. Let me know if I can assist you, but maybe a call would be better then :).
2.) Did you try `"anything".startsWith("")` in a JS console?
4.) What differences do you mean? In dev-Branch?

patrickrb: 1) Ok, so if a user changes their input completely before the results have returned it fires off a second request, leaving 2 requests open to the server at a single time.

See here:
image showing stale requests being made unnecessarily

It can be improved so we cancel the previous request instead of letting it finish:

image showing us abort from a stale request


2) Played with and It seems to only handle the first time the event is fired, which is fine...
4) Ahh i see it was added in both spots now

I went ahead and fired up the branch and tested the code below, i think the following implementation is easier to reason about, adds some performance upgrades, and updates it to modern javascript patterns.

- We really dont need to redeclare `call` as `call_now`  in the success call back, below is a cleaned up version showing you dont need it
- use `===` in the comparison
- use const and keep variables immutable
- adds abort to the request if the user changes the input based on the start of the string only

```javascript
// On Key up check and suggest callsigns
$("#callsign").keyup(function () {
	const call = $(this).val().toUpperCase();
	const cleanedCall = call.replace('0','Ø');

	// Exit early if the users starts typing a completely new call
    if(scp_data.previousRequest && !call.startsWith(scp_data.request)) {
        scp_data.previousRequest.abort();
    }

	
	if (call.length >= 3) {
        if ( scp_data.request === "" || !call.startsWith(scp_data.request) ) {
          scp_data.request = call;
          scp_data.data = [];
          scp_data.previousRequest = $.ajax({
              url: 'lookup/scp',
              method: 'POST',
              data: {
                  callsign: call
              },
              success: function (result) {
                if (call.startsWith(call)) {
                	scp_data.data = result.split(" ");
                    $('.callsign-suggestions').text(filterCallsignList(cleanedCall, scp_data.data));
                    highlight(cleanedCall);
                  }
              },
			  complete: function () {
				scp_data.previousRequest = null;
			  }
          });

        } else {
          // Filter the already obtained list:
          $('.callsign-suggestions').text(filterCallsignList(cleanedCall, scp_data.data));
          highlight(cleanedCall);

        }
		// moved to blur
		// checkIfWorkedBefore();
		var qTable = $('.qsotable').DataTable();
		qTable.search(call).draw();
	}
	else {
		$('.callsign-suggestions').text("");
	}
});
```

DJ3CE: You are again missing out that by the time, the success-callback is executed, the user might have typed additional letters.

I see no reason for complicating the code by introducing more global-scope items.

patrickrb: ope, i see now about `call` `call_now`, updating code above in a sec

patrickrb: oh no, this isnt an issue with that at all...it seems like the caching isnt working like it was thought to?


When i did a search for "WA6BND" it grabbed "WA6" whose results did not include `WA6BND`, heres the list it did return:

```
HP3/WA6YJW VP2V/WA6O WA6AAC WA6AEE WA6ALA WA6AM WA6AMK WA6APQ WA6AQL WA6ARA WA6AUP WA6AZP WA6BEF WA6BFG WA6BGS WA6BJH WA6BJM WA6BKR WA6BXI WA6BXV WA6CAL WA6CDO WA6CGR WA6CQE WA6D WA6DBC WA6DH WA6DIL WA6DIR WA6DKD WA6DKN WA6DNR WA6DON WA6DPP WA6EBO WA6ECH WA6ED WA6EEM WA6EGG WA6EJO WA6ERA WA6EZV WA6F WA6FFJ WA6FGV WA6FIZ WA6FV WA6FWN WA6FXT WA6FXY WA6FYD WA6FYV WA6FZV WA6G WA6GBW WA6GD WA6GDU WA6GFE WA6GFR WA6GJW WA6GUN WA6GVI WA6GXQ WA6HNA WA6HQQ WA6HTP WA6HXD WA6HXM WA6HXW WA6HZT WA6HZY WA6I WA6IBU WA6IKE WA6IOQ WA6IPD WA6IQO WA6J WA6JA WA6JBZ WA6JGM WA6JJA WA6JJM WA6JKN WA6JRP WA6JRZ WA6KFI WA6KHK WA6KJN WA6KJV WA6KLK WA6KSF WA6KTK WA6KYR WA6LAU WA6LBU WA6LDI WA6LGS WA6LIE WA6LJS WA6LOC WA6LOS/DU3 WA6MBL WA6MCL WA6MDI WA6MEM WA6MHZ WA6MIZ WA6MM WA6MNV WA6MOW WA6MP WA6MPN WA6MQF WA6MRK WA6MTN WA6MTZ WA6MVN WA6N WA6NEA WA6NEQ WA6NFA WA6NFJ WA6NHC WA6NKM WA6NPA WA6NPC WA6NUO WA6NVL WA6O WA6OC WA6OCV WA6OCV/M WA6OEF WA6OEM WA6OIB WA6OJY WA6OSX WA6OT WA6OUD WA6OVP WA6OWM WA6OWM/KH6 WA6OWM/QRP WA6OXK WA6P WA6PEL WA6PHR WA6PIC WA6PKB WA6POZ WA6PPG WA6PPX WA6PSI WA6PWP WA6PXX WA6PY WA6Q WA6QBU WA6QGH WA6QHD WA6QMQ WA6QYS WA6RE WA6RKK WA6RLR WA6RSV WA6RWI WA6RYA WA6RZI WA6RZR WA6SCW WA6SDB WA6ST WA6STV WA6SWM WA6SXU WA6SZE WA6T WA6TE WA6TFZ WA6TGT WA6TLA WA6TLL WA6TMJ WA6TOW WA6TQJ WA6TQT WA6TVO WA6TWF WA6TWJ WA6TXU WA6TYQ WA6UIM WA6UMA WA6URY WA6UVW WA6UXA WA6UYO WA6VAB WA6VHU WA6VPJ WA6VWW WA6WA WA6WGS WA6WJN WA6WPG WA6WXM WA6YAW WA6YBR WA6YDQ WA6YEE WA6YOU WA6YPE WA6YSM WA6ZKM WA6ZTY WA6Ã˜87SWL YU/WA6MP
```


So, this is a huuuuge reason to implement debounce if you want caching :)

patrickrb: Ok, so they're not global scoped, they're scoped to the anonymous function that is passed into the event.
I've gone ahead and updated it to catch the users input if its changed after the results are returned.
This is a much cleaner implementation, please use it:

```javascript
// On Key up check and suggest callsigns
$("#callsign").keyup(function () {
	const call = $(this).val().toUpperCase();
	const cleanedCall = call.replace('0', 'Ø');

	// Exit early if the users starts typing a completely new call
	if (scp_data.previousRequest && !call.startsWith(scp_data.request)) {
		scp_data.previousRequest.abort();
	}

	if (call.length >= 3) {
		if (scp_data.request === "" || !call.startsWith(scp_data.request)) {
			scp_data.request = call;
			scp_data.data = [];
			scp_data.previousRequest = $.ajax({
				url: 'lookup/scp',
				method: 'POST',
				data: {
					callsign: call
				},
				success: function (result) {
					const currentCall = $("#callsign").val().replace('0', 'Ø').toUpperCase();
					if (currentCall.startsWith(call)) {
						scp_data.data = result.split(" ");
						$('.callsign-suggestions').text(filterCallsignList(currentCall, scp_data.data));
						highlight(currentCall);
					}
				},
				complete: function () {
					scp_data.previousRequest = null;
				}
			});

		} else {
			// Filter the already obtained list:
			$('.callsign-suggestions').text(filterCallsignList(cleanedCall, scp_data.data));
			highlight(cleanedCall);
		}
		var qTable = $('.qsotable').DataTable();
		qTable.search(call).draw();
	}
	else {
		$('.callsign-suggestions').text("");
	}
});
```

DJ3CE: I don't see an advantage but increased complexity, sorry.

Scoping: I guess we are not talking about the same thing.

patrickrb: Did you try the code?  Which part is too complex?  

Also, You clearly don't understand scoping.

patrickrb: Ok, the whole pr increases complexity, but 3 lines that suggested is  too complex?  

Bro, I've been trying to be nice this whole pr.  Your pr was messy, used old and outdated JavaScript patterns.   I'm not saying this to be mean, but to keep the code quality high for a project I love.

DJ3CE:
@magicbug, please let me know if there's anything which prevents merging.

assets/js/sections/contesting.js
```diff
@@ -8,6 +8,11 @@ $(document).ready(async function () {
 	setRst($("#mode").val());
 });
 
+var scp_data = {
```
patrickrb, line 11: this should be a `let`

assets/js/sections/contesting.js
```diff
@@ -208,26 +213,43 @@ $('#start_date').change(function () {
...
-			method: 'POST',
-			data: {
-				callsign: $(this).val().toUpperCase()
-			},
-			success: function (result) {
-				$('.callsign-suggestions').text(result);
-				highlight(call.toUpperCase());
-			}
-		});
+        if ( scp_data.request == "" || ! call.startsWith(scp_data.request) ) {
```
patrickrb, line 219: This should be `===`

assets/js/sections/contesting.js
patrickrb, line 229: this should be `let`

assets/js/sections/qso.js
```diff
@@ -1,4 +1,9 @@
 var lastCallsignUpdated=""
+var scp_data = {
```
patrickrb, line 2: `let`

assets/js/sections/qso.js
```diff
@@ -1019,18 +1024,34 @@ $("#callsign").on("keypress", function(e) {
...
-		data: {
-		  callsign: $callsign.toUpperCase()
-		},
-		success: function(result) {
-		  $('.callsign-suggestions').text(result);
-		}
-	  });
+	  let call = $(this).val().toUpperCase();
+
+      if ( scp_data.request == "" || ! call.startsWith(scp_data.request) )  {
```
patrickrb, line 1029: `===`

patrickrb (changes requested):
I still dont like this change, but i can add more of my changes if this get's accepted.

assets/js/sections/contesting.js
DJ3CE: `var` and `let` are equivalent here.

assets/js/sections/contesting.js
DJ3CE: As `scp_data.request` always is a string, `==` is okay here.

assets/js/sections/contesting.js
DJ3CE: `var` and `let` are equivalent here.

assets/js/sections/qso.js
DJ3CE: :) `var` and `let` are equivalent here.

assets/js/sections/qso.js
DJ3CE: See the other comment: As `scp_data.request` always is a string, `==` is okay here. In addition, if it is not a string, then `startsWith()` coerces it to string, which will give the correct behaviour.

assets/js/sections/contesting.js
patrickrb, line 11: This should be let, i will net let it go.

assets/js/sections/contesting.js
patrickrb, line 219: this needs to be `===`

assets/js/sections/contesting.js
patrickrb, line 229: this should actually be `let`,

assets/js/sections/qso.js
patrickrb, line 2: another `let`

patrickrb (commented):
In JavaScript, whether to use let, const, or var for variable declarations depends on the specific use case and the scope where the variable is needed.

var: This is the oldest keyword for declaring variables, existing since the beginning of JavaScript. Variables declared with var are function-scoped or globally-scoped if declared outside of a function. They can be re-declared and updated. However, var has some pitfalls, such as hoisting, which can lead to bugs, especially in large codebases.

let: Introduced in ES6 (ECMAScript 2015), let provides block-scoped variables. Unlike var, variables declared with let are limited in scope to the block, statement, or expression where they are used. let allows you to update the variable but not re-declare it within the same scope. This reduces the risk of bugs compared to var.

const: Also introduced in ES6, const is used to declare variables whose values are not intended to change. const provides block-scoping similar to let. The key difference is that once a variable is assigned with const, its value cannot be reassigned. Attempting to do so will result in a runtime error. However, note that if a const variable holds an object or an array, the object or array's contents can still be modified (the variable's reference to the object or array cannot change).

Best Practices
Prefer const: Use const by default for all of your variable declarations if you know their values will not change. This makes your code easier to reason about since you know these variables won't be reassigned later in your code.

Use let for variables that change: If you have variables that need to change, like counters in a loop, or values that get reassigned, use let. This signals to others that this variable will be reassigned.

Avoid var: Due to its function scope and hoisting behavior, var can lead to confusion and bugs, especially in complex codebases. Modern JavaScript codebases tend to avoid var in favor of let and const for clearer scoping rules.

In conclusion, while it's not a rule to "always use let," the modern practice leans towards using const by default and let when necessary for variables whose values need to change. Avoiding var is generally considered good practice in contemporary JavaScript development.

assets/js/sections/qso.js
patrickrb: it should be `===` always use `===`

assets/js/sections/contesting.js
patrickrb: no, they arent.  please read about javascript before submitting PR's to change javascript.

assets/js/sections/contesting.js
patrickrb: no it's not, it needs to `===`.  Please read up on javascript best practices before submitting PR's for changes to javascript.

patrickrb:
Please read up on javascript best practices before refusing changes that catch the code up to javascript best practices.

We should *ALWAYS* use `===` checks.

We should *ALWAYS* use `let` and `const`, *NEVER* `var`.

patrickrb:
Feel free not to use the `.abort` logic if you think its too complex, but the rest of the `keyUp` function should look like this:

```javascript
// On Key up check and suggest callsigns
$("#callsign").keyup(() => {
	const call = $(this).val().toUpperCase();
	const cleanedCall = call.replace('0', 'Ø');

	// Exit early if the users starts typing a completely new call
	if (scp_data.previousRequest && !call.startsWith(scp_data.request)) {
		scp_data.previousRequest.abort();
	}

	if (call.length >= 3) {
		if (scp_data.request === "" || !call.startsWith(scp_data.request)) {
			scp_data.request = call;
			scp_data.data = [];
			scp_data.previousRequest = $.ajax({
				url: 'lookup/scp',
				method: 'POST',
				data: {
					callsign: call
				},
				success: function (result) {
					const currentCall = $("#callsign").val().replace('0', 'Ø').toUpperCase();
					if (currentCall.startsWith(call)) {
						scp_data.data = result.split(" ");
						$('.callsign-suggestions').text(filterCallsignList(currentCall, scp_data.data));
						highlight(currentCall);
					}
				},
				complete: function () {
					scp_data.previousRequest = null;
				}
			});

		} else {
			// Filter the already obtained list:
			$('.callsign-suggestions').text(filterCallsignList(cleanedCall, scp_data.data));
			highlight(cleanedCall);
		}
		var qTable = $('.qsotable').DataTable();
		qTable.search(call).draw();
	}
	else {
		$('.callsign-suggestions').text("");
	}
});
```

It's inarguably cleaner than what've you've provided in your PR.

DJ3CE:
> I'll be honest the convention in Cloudlog should be Ø ...

Sounds like s.th. to work with and to have incremental improvement :). 

Now replacing 'Ø' by '0' in `controllers/Lookup.php:scp()`, as both scp-files (`clublog_scp.txt` and `masterscp.txt`) and the db-table use '0', then it makes no difference how the backend gets called. The conversion back to 'Ø' is already done in `scp()`, an exception of '0's shows up with `getSessionQsos` in contesting.

> Can confirm that it now works, but we have lost the ability to just type the suffix and show matches based on that which is super handy with the SCP box

Thanks for pointing that out, haven't observed nor used that up to now (I seem not to remember suffixes without prefixes :D), so now consistently switching to `String.includes()` for matching cache-validity and showing results. @patrickrb already mentioned `includes()`, but neither with the background @magicbug gave here nor consistent - sorry, haven't seen that before.

Is there a library-like js-file to have one place for code(-pieces) like this? I have separated the 'functional' code from its specific application to a keyup, so both, `qso.js` and `contesting.js` have the same functional code, which could be put in a single place elsewhere (... and no more global-variable is needed :) ).

PR #3026, opened 2024-03-25
Make dxped-url configurable

DJ3CE:
This PR adds a configuration item for the DXpeditions-source. Quite useful, if you'd like to have a cronjob loading the json onto your server and then using a (filesystem-)path for improved performance.

A path is relative to the cloudlog 'root' directory, but of course you can also use an absolute path, or an URI.

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -32,7 +32,6 @@ $config['datadir'] = null; // default to install directory
 $config['table_name'] = "TABLE_HRD_CONTACTS_V01";
 $config['locator'] = "";
 $config['display_freq'] = true;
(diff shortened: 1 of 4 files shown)


patrickrb (approved):
nice!

magicbug:
Great idea, I'm trying to move away from hard coded config items you'll notice there's an options table where OSM maps tile providers are installed

It has some options

To get the value from the database use 
`$this->optionslib->get_option('option_name');` 

To set/update you can use this
`$this->optionslib->update('option_name', 'value', 'yes or no')` 

Yes/No just sets whether its automatically loaded if its yes you can just use

`$config['option_name']`

It will need a migration file for the database to set the URL into the MySQL table

DJ3CE:
Sounds nice. See commit.

PR #3027, opened 2024-03-25
Contest "copy exchange to"

DJ3CE:
In contest-logging, Cloudlog featured copying the received exchange into the (rather specific) DARC DOK field.

This PR brings this valuable approach to a broader set of contests by implementing 'copy exchange to'
* age,
* name,
* US-State,
* power used by the other station ('RX_PWR'),
* locator

which could easily be enhanced for other needs (I just included those, I know about).

Some notes:
* 'locator' might be confusing. This copies the **exchange** field to locator, not the received-locator-exchange field and is rather useful for some edge cases, I guess. I'm thinking of disabling that, if an exchange type including a locator was selected
* in `models/Logbook_model.php`, an example for a more sophisticated copy is included
* some fields never appear in Cloudlog, like 'age'. But why not fill it - it's in the database. :)

As always, although not mentioning it everytime :), I'm open to discussions, wishes, enhancements, ...


magicbug:
Implementation looks good, happy for this to be merged 👍🏻

PR #3028, opened 2024-03-25
Post-Contest-Logging: Fixes and focus to time-field

DJ3CE:
This PR sets the focus to the start-time-field after submitting a qso in post-constest-logging.

Additionally a fix for the different start-time-format is included, which fixes the internal server error on checking dupes and listing the qso's so far.

That's it, i guess, for the next couple of days. As always: Feel free to reach out.


magicbug:
Thanks

PR #3034, opened 2024-03-27
Add live/post QSO markers

phl0:
This adds badges to show live/post QSO markers. Addresses https://github.com/magicbug/Cloudlog/discussions/3033


soakland5:
I'm using the web version, not the hosted version.

phl0:
> I'm using the web version, not the hosted version.

I think you will now get the update after midnight (UTC).

soakland5:
Thank you much!

PR #3042, opened 2024-03-31
Limit Grid square to 6 chars in request to df2et.de API

DanielBausset:
In the account settings, the Locator can contain up to 16 characters.
In the Sattimers controller, the API call to df2et.de uses this locator.
But the df2et.de API works only with 6-characters locators maximum.
So this patch just crops the locator to 6 characters max before calling the df2et.de API :)

73

### DIFF
--- a/application/controllers/Sattimers.php
+++ b/application/controllers/Sattimers.php
@@ -15,10 +15,10 @@ class Sattimers extends CI_Controller {
         $footerData['scripts'] = [
            'assets/js/sections/sattimers.js?'
         ];
-        $url = 'https://www.df2et.de/tevel/api2.php?grid='.strtoupper($this->stations->find_gridsquare());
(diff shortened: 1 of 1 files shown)


phl0:
This is a good find. But I rarher fixed the code upstream so that the API call works with more than 6 chars. :-)

DanielBausset:
Hi Florian, indeed that'd be even better if you could improve this :)

73

phl0:
> Hi Florian, indeed that'd be even better if you could improve this :)

I did already. Was implemented by the time I commented. Can you confirm that it is working (again)?

DanielBausset:
I confirm that it's working, I just tested it with a locator up to 10 characters.

magicbug:
@phl0 Shall we keep this on the API end rather than a change inside Cloudlog? your shout let me know

phl0 (approved):
Well ok to merge in though fix is in the API already. But guess it would not harm.

PR #3045, opened 2024-04-02
Visitor page pagination

### LINKED ISSUE #1720: Pagination in public view
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the forum see readme for details**

**Describe the bug**
No pagination in Public View

**To Reproduce**
Steps to reproduce the behaviour:
1. Make sure you've got enough QSO's  for pagination
2. Go to Public view 
3. Scroll down to  bottom
4. No pagination

**Expected behaviour**
Pagination to go through all QSO's

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser  Edge
 - Version 106.0.1370.52 (Officiële build) (64-bits)

And if it's not a bug then maybe a feature request

patrickrb:
Closes #1720 

- Added pagination to visitors page ~*only* if the public search option is enabled~
- ~Happy to update this PR if you'd rather a different config option or any other feedback, but to me it felt like if search is enabled then pagination might make sense too.  My biggest concern would be that people who have this feature enabled currently might not want the paginated results.~

Enhancements to public search functionality:

* [`application/controllers/Visitor.php`](diffhunk://#diff-e5390d77eb91cd0945b808102973347ba151fe27ed0580d54cfb99cab6a3bfb1R78-R109): In the `index` function, added a condition to load and initialize the pagination library only if public search is enabled. The pagination configuration includes settings for the base URL, total rows, per page limit, number of links, and various HTML tags and attributes for styling the pagination links.
* [`application/controllers/Visitor.php`](diffhunk://#diff-e5390d77eb91cd0945b808102973347ba151fe27ed0580d54cfb99cab6a3bfb1L111-R146): Modified the data fetching logic in the `index` function to return paginated results when public search is enabled, otherwise it continues to return the last 18 records as before.
* [`application/views/visitor/index.php`](diffhunk://#diff-c5778b0b37e488846c56144df03ce423b41eaa98b45984047357536e3900dcc1R82-L88): updated the condition and loop to use the new `results` variable instead of `last_five_qsos`. Also added a condition to display pagination links at the end of the table if public search is enabled.

Gif of feature working:

### DIFF
--- a/application/controllers/Visitor.php
+++ b/application/controllers/Visitor.php
@@ -75,6 +75,38 @@ class Visitor extends CI_Controller {
 
                 $this->load->model('logbook_model');
 
+				// Only load, config, and init pagination if public search is enabled
+				if ($this ->public_search_enabled($public_slug)) {
+					$this->load->library('pagination');
+				
+					//Pagination config
+					$config['base_url'] = base_url().'index.php/visitor/'. $public_slug . '/index';
+					$config['total_rows'] = $this->logbook_model->total_qsos($logbooks_locations_array);
+					$config['per_page'] = '25';
+					$config['num_links'] = $this->logbook_model->total_qsos($logbooks_locations_array) / 25;
+					$config['full_tag_open'] = '<ul class="pagination">';
+					$config['full_tag_close'] = '</ul>';
+					$config['attributes'] = ['class' => 'page-link'];
+					$config['first_link'] = false;
+					$config['last_link'] = false;
+					$config['first_tag_open'] = '<li class="page-item">';
+					$config['first_tag_close'] = '</li>';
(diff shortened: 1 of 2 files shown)


magicbug:
Hi Patrick 

I actually dont think it needs to be controlled just enabled by default, public search really should just allow searching the logbook and nothing else 

so if you could change it to just have pagination on all the time its ready for merge!

patrickrb:
on it! Thanks @magicbug!

patrickrb:
> Hi Patrick
> 
> I actually dont think it needs to be controlled just enabled by default, public search really should just allow searching the logbook and nothing else
> 
> so if you could change it to just have pagination on all the time its ready for merge!

Got it fixed up, thanks again!

PR #3047, opened 2024-04-02
Feature - SSTV Image Page

patrickrb:
This pull request introduces a new page to view SSTV images,  several changes aimed at refactoring the code and enhancing functionality. The main changes include the addition of a new `storage_helper.php` file, and the removal of redundant functions from multiple files, 

Refactoring and Functionality Enhancements:

* [`application/helpers/storage_helper.php`](diffhunk://#diff-413abc91cb9cb7f2cc0f661e201eea3e2c18c458a63a89b58143afcbb28d4dbcR1-R44): A new helper file was added that includes the `folderSize` and `sizeFormat` functions. These functions were previously duplicated in multiple files. This change improves code maintainability and reusability.

* `application/controllers/Eqsl.php`, `application/controllers/Qsl.php`, `application/controllers/Sstv.php`: The `folderSize` and `sizeFormat` functions were removed from these files and replaced with calls to the newly created `storage_helper` functions. This change simplifies the codebase and removes redundancy. [[1]](diffhunk://#diff-87dcfbeca1607683d0b6ed30aa83ac2b85c83295f241b8121832414b8dee8d20L752-L796) [[2]](diffhunk://#diff-7f9f9c317ce9b5b504053037c689f485ad5599d454020c3df998b4cde73c104bL185-L228)

Language Support Enhancements:

* [`application/language/*`](diffhunk://#diff-791e4059c4dde525e94aa3cb150f947202c0239c9c9fffe6ad927dea0c27413dR1-R5): New SSTV-related language files were added to support multiple languages. These changes improve the application's internationalization.

* [`application/language/*/menu_lang.php`](diffhunk://#diff-0e791651e6f13a488545218ccd7cd95715723d737af0b0438ff113690aa15fc6R20): The menu item for viewing SSTV images was added in multiple languages. This change improves the user interface and user experience for non-English speakers. [[1]](diffhunk://#diff-0e791651e6f13a488545218ccd7cd95715723d737af0b0438ff113690aa15fc6R20) [[2]](diffhunk://#diff-41c120da2febdc063e6cbe87f06a548b67a1ef047c629dcd947bacf5e53873bbR20) [[3]](diffhunk://#diff-4ebbdb1e30d8537482447dcd95c2481a38ad5eca300dd44b151fa6970625d983R20) [[4]](diffhunk://#diff-d83bedfb77f2c07768be4ac31a54a535db47ce9aba80f6e07a2751e74b62fb46R20)

Minor Changes:

[...]

PR #3031, opened 2024-03-27
Add SSTV image support

patrickrb:
resolves #3016
closes #3016 

- Adds basic SSTV image support from viewing an existing QSO in the logbook
- If the QSO mode is `SSTV` then we show a new tab named "SSTV Management", otherwise we show nothing
  - New SSTV Management tab works similarly to QSL management tab, except we only have one input and it can handle multiple images at a time.
  - Users can delete or view an SSTV Image from the SSTV Management tab
- If there are any SSTV Images associated with the selected QSO then we show a "SSTV Images" tab
  - Inside the SSTV Images tab is a carousel similar to the QSL Card carousel
- Definitely not sure that I did the localization stuff correctly.

GIF of feature:

### SSTV QSO Details (Only shows if QSO mode is SSTV):
image
image

### View SSTV Image from management tab:
image

### Delete SSTV Image from management tab:
image

### SSTV Carousel on SSTV Images tab:
image

### Non-SSTV QSO Details (not showing SSTV options):
image

PR #3061, opened 2024-04-07
Add missing German translations for SSTV stuff

phl0:
Seems that in https://github.com/magicbug/Cloudlog/pull/3031 and https://github.com/magicbug/Cloudlog/pull/3047 at least German translations are broken:

Screenshot from 2024-04-07 08-47-17

Screenshot from 2024-04-07 08-50-10

Resulting in empty menu string and half German on SSTV view. This fixes the German part. Review by @patrickrb for other languages may be applicable :)


patrickrb:
Thank you for the fix!  Was out of town yesterday

PR #3068, opened 2024-04-09
Update pota.txt

Ciemon:
After changes made to park naming in the POTA programme this update provides a sorted list of the latest park numbers from POTA.
If any other files need updating to enable this new park numbers please highlight them for me so I can look into them.


Ciemon:
Noticed that pull requests are only accepted against Dev, moving there.

PR #3069, opened 2024-04-10
Update pota.txt

Ciemon:
After changes made to park naming in the POTA programme this update provides a sorted list of the latest park numbers from POTA.


Ciemon:
The current documentation set states:

Required ADIF fields

    STATION_CALLSIGN or OPERATOR
    CALL
    QSO_DATE
    TIME_ON
    BAND
    MODE or SUBMODE

Optional ADIF fields

    MY_SIG and MY_SIG_INFO
    SIG and SIG_INFO
    MY_STATE

The new park references are in place now, the old ones are not recognised by the website. As an example park https://pota.app/#/park/G-0597 (which is linked from a QSO in my log) fails to resolve, but https://pota.app/#/park/GB-0597 does resolve as it's the new name and URL.

Currently if I use GB-0597 within Cloudlog, I'm asked to add it, which then correctly links into the POTA site for the information Cloudlog requires.

magicbug:
I'm happy to add the new list, but I think POTA will need to provide a way of mapping old to new this seems poorly thought out on their part.

phl0:
This clearly needs a migration to map the already logged QSOs to the new references. 
Once I asked for an ADIF download of the hinterlegt to be used to insert/update references. This was denied with "would not trust the info entered by activators" ... 🤷‍♂️

PR #3084, opened 2024-04-21
Jump to RSTr after callsign input

phl0:
Align tab index as per https://github.com/magicbug/Cloudlog/discussions/3083.


phl0:
@GregTroar Pls test and report.

GregTroar:
which branch should I use ?

phl0:
> which branch should I use ?

It is sitting in my branch under https://github.com/phl0/Cloudlog/tree/tabindex until @magicbug merges this one to dev.

GregTroar:
ok will try when I can pull it

PR #3101, opened 2024-05-13
Lotw fixing

phl0:
This fixes a few things in LoTW uploads:

- The OpenSSL base64 data normally needs a newline at the end to be able to verify signatures correctly. This has been worked around on LoTW side but probably be good to be compliant in the first place
- The upload filename has the full path stripped. Not really an issue but just in case ... It does not harm to hide the directory structure of the uploading system
- The unsafe cURL file upload has been removed from PHP. So this part of the code should be removed as well as PHP fails with an error because the insecure method is not allowed any more.


magicbug:
Thanks Flo agreed should be addressed

PR #3115, opened 2024-05-21
Enable API qso to accept also form data and not only Json

emics:
Added the possibility of sending data to the "qso" method API also in www-x-form-urlencoded format to allow the automatic sending of QSOs via software such as Log4OM and other.
Perhaps it is not the most elegant way to develop this feature, you probably prefer to create a new method. 
Obviously I leave the choice up to you.
73 emiliano


magicbug:
I can understand the appeal but prefer the current API methods as for Log4OM I believe it already supports Cloudlogs API :)

emics:
in reality Log4OM allows you to specify the URL of the POST call and a series of Key/Value pairs which however are sent with this syntax:
key1=value1&key2= value2&key3=value3

if I try to inject a json as a value obviously the API doesn't like it.

You choose how you think is best to proceed, in my local version I will fork the method for my use.

emics:
Hi @magicbug Please delete this pull Request, yesterday i have found the specific configuration in Log4OM for CloudLog. My changes are useless. sorry.

emics:
no more needed

PR #3169, opened 2024-07-11
Prevent also RPT QSOs from being uploaded to LoTW

phl0:
The current code tries to also upload QSOs with PROP_MODE set to RPT to LoTW. Those QSOs are denied and not processed. In addition any other valid QSOs in the same batch are ignored. So we should only try to upload QSOs whose prop mode is not INTERNET or RPT.


magicbug:
Sounds very reasonable!

PR #3188, opened 2024-09-08
Return the station list sorted by call and name

hugosilvaparagramanet:
When going through that station profile list, this appears to be sorted randomly.
Use cases are (for ex) on Import and Export ADIF, where it's difficult to find the correct profile if you have many stations.
By  returning the list of stations sorted by call and name some organization is provided, and it becomes easier to find the correct station profile for the operation.


DanielBausset:
Nice one :)

PR #3428, opened 2026-04-20
Add Config Toggle to Show Confirmations in Visitor Map

PhillipHunter:
On the public visitor gridsquare and confirmations maps at `/visitor/gridmap/<callsign>`, I noticed that grids that the station operator has confirmed (via LoTW, eQSL, paper QSL, or QRZ.com) never appear as "confirmed" on the map, and stayed their red color.

It appears that the calls to the function for this did not have any of these services enabled for confirmation when called from the visitor page, causing this erroneous behavior.

I have since changed it to be true for these fields, allowing it to properly propagate those confirmations to the public page.


PhillipHunter:
Pulling to draft and making this configurable, as I just noticed the comments regarding it being set to false, so I assume there was some intent there...

PhillipHunter:
Added config setting for this, though I'm not sure why one would ever want that off. I'm not very familiar with the details of GDPR and such though.

magicbug:
Nice idea, I am trying to avoid putting anything in config.php anymore and instead using the Global Options in Cloudlog, where the options are storeNice idea, I am trying to avoid putting anything in config.php anymore and instead using the Global Options in Cloudlog, where the options are stored in the database


This is using the OptionsLib library

// Example: set/update a global option
$this->load->library('OptionsLib');
$this->load->model('Options_model');

$optionName  = 'my_feature_enabled';
$optionValue = '1';     // options are stored as strings
$autoload    = 'yes';   // 'yes' or 'no'

if ($this->Options_model->item($optionName) === null) {
    // Create new option
    $this->optionslib->save($optionName, $optionValue, $autoload);
} else {
    // Update existing option
    $this->optionslib->update($optionName, $optionValue, $autoload);
}

Usually I create a migration to make the new option automatically exist

PhillipHunter:
Updated and tested it with your suggestions, @magicbug . I'm not sure how you handle the language translation stuff though for the additional languages.

magicbug:
Thanks for this @PhillipHunter — I've rebased the work onto `dev` (we don't merge feature PRs into `master`) and finished the remaining bits:

- Global Options via OptionsLib + a migration (now 278, since 267 was already used on `dev`)
- Language strings for the other locales
- Visitor gridmap calls now pass logbook locations after the satellite-orbit argument so public maps still query the right stations

Landed on `dev` in https://github.com/magicbug/Cloudlog/pull/3461 — closing this PR in favour of that merge.

PR #3439, opened 2026-05-16
Add QRZCALL.EU as a QSO upload target

Ron6519:
## Add QRZCALL.EU as a QSO upload target

This PR lets Cloudlog mirror QSOs to a user's [QRZCALL.EU](https://qrzcall.eu)
logbook in real time, alongside the existing QRZ.com / Club Log / HRDLog
upload targets.

It builds on the QRZCALL.EU callbook-provider PR, but is functionally
independent — the callbook lookup reads a token from `config.php`, whereas
QSO upload uses a per-station-profile PAT. The two can be reviewed and merged
in either order.

---

### How it works

QRZCALL.EU exposes a QRZ-compatible logbook endpoint
(`api.qrzcall.eu/v1/pub/logbook_api.php`) that accepts the same
`KEY=&ACTION=INSERT&ADIF=` request format Cloudlog already uses for QRZ.com.

Upload is **real-time** and happens automatically for station profiles that
have a QRZCALL.EU PAT with real-time upload enabled:

- **QSO logged** (`create_qso()`) → the new QSO is sent immediately (INSERT)
- **QSO edited** (`edit()`) → the changed QSO is sent immediately with
  `OPTION=REPLACE`, so QRZCALL.EU updates the existing record in place rather
  than creating a duplicate

There is **no bulk / cron upload route** — synchronisation is real-time only.
Bulk import/export is handled on the QRZCALL.EU website itself, not from
external loggers.

Each upload records its status and timestamp on the QSO
(`COL_QRZCALL_QSO_UPLOAD_STATUS` / `_DATE`).

---

### Changes

**7 files — 2 new, 5 modified**

| File | Change |
|---|---|
| `application/migrations/271_add_qrzcall_to_cloudlog.php` | **NEW** — adds `qrzcallapikey` / `qrzcallrealtime` to `station_profile`; adds `COL_QRZCALL_QSO_UPLOAD_STATUS` / `COL_QRZCALL_QSO_UPLOAD_DATE` to the QSO table |
| `cypress/e2e/6-qrzcall.cy.js` | **NEW** — Cypress E2E test for the QRZCALL.EU station-profile fields |
| `application/models/Logbook_model.php` | `push_qso_to_qrzcall()` (sends `OPTION=REPLACE` for edits; treats a `duplicate` response as success), `mark_qrzcall_qso_sent()`, `exists_qrzcall_api_key()`; real-time INSERT hook in `create_qso()`; real-time REPLACE hook in `edit()` |
| `application/config/migration.php` | Bumps `migration_version` to `271` |
| `application/models/Stations.php` | Saves `qrzcallapikey` / `qrzcallrealtime` in `create()` and `edit()` |
| `application/views/station_profile/edit.php` | QRZCALL.EU PAT field + real-time upload select |
| `application/views/station_profile/create.php` | Same fields on the create form |

---

### Configuration

Station Profile → enter the QRZCALL.EU PAT and set real-time upload to **Yes**.
QSOs logged or edited under that station profile then sync automatically.

---

### Testing

Verified on the Docker development environment (PHP 7.4, MySQL):

- Migration 271 applies cleanly (idempotent — `field_exists` guards)
- Real-time create — QSO logged → immediately uploaded (INSERT), status `Y`
- Real-time edit — QSO edited → immediately re-sent with `OPTION=REPLACE`;
  QRZCALL.EU updates the existing record in place (verified `RESULT=REPLACE`,
  same logbook ID, no duplicate row)
- Invalid PAT → the upload call fails cleanly, the QSO's status is left
  untouched (no retry loop — the hook fires once per save)
- New Cypress test `6-qrzcall.cy.js` passes; existing suite unaffected

**Test account** — a shared account is available for reviewers:

```
Callsign:  TEST0WVLG
PAT:       pat_b3gvr7n6wyzjvs9xk4jmpgxm3r3gtgdx
```

Enter the PAT in a station profile (Station Profile → QRZCALL.EU card), enable
real-time upload, then log or edit a QSO.

---

### Screenshot

The QRZCALL.EU card on the station profile (PAT field + real-time toggle) is
attached to this PR.

---

### No breaking changes

All changes are additive. Upload only happens for station profiles that have
a QRZCALL.EU PAT set with real-time upload enabled. Existing QRZ.com / Club Log
/ HRDLog upload behaviour is unchanged.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 270;
+$config['migration_version'] = 271;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/migrations/271_add_qrzcall_to_cloudlog.php
+++ b/application/migrations/271_add_qrzcall_to_cloudlog.php
@@ -1,0 +1,68 @@
+<?php
+
+defined('BASEPATH') or exit('No direct script access allowed');
+
+/*
+ * Migration: Add QRZCALL.EU support to Cloudlog
+ *
+ * Adds two columns to station_profile for storing the user's Personal Access
+ * Token and real-time upload preference, and two columns to the QSO table
+ * for tracking upload status.
+ */
+
+class Migration_add_qrzcall_to_cloudlog extends CI_Migration
+{
+    public function up()
+    {
+        // --- station_profile: QRZCALL.EU Personal Access Token ---
+        if (!$this->db->field_exists('qrzcallapikey', 'station_profile')) {
+            $fields = [
+                'qrzcallapikey varchar(100) DEFAULT NULL',
+            ];
+            $this->dbforge->add_column('station_profile', $fields);
+        }
+
+        // --- station_profile: real-time upload toggle ---
+        if (!$this->db->field_exists('qrzcallrealtime', 'station_profile')) {
+            $fields = [
+                'qrzcallrealtime tinyint(1) DEFAULT 0',
+            ];
+            $this->dbforge->add_column('station_profile', $fields);
+        }
+
+        // --- QSO table: upload status (NULL/N = pending, Y = uploaded, M = modified/re-upload) ---
+        if (!$this->db->field_exists('COL_QRZCALL_QSO_UPLOAD_STATUS', $this->config->item('table_name'))) {
+            $fields = [
+                'COL_QRZCALL_QSO_UPLOAD_STATUS varchar(1) DEFAULT NULL',
+            ];
+            $this->dbforge->add_column($this->config->item('table_name'), $fields);
+        }
+
+        // --- QSO table: timestamp of last successful upload ---
+        if (!$this->db->field_exists('COL_QRZCALL_QSO_UPLOAD_DATE', $this->config->item('table_name'))) {
+            $fields = [
+                'COL_QRZCALL_QSO_UPLOAD_DATE datetime DEFAULT NULL',
+            ];
+            $this->dbforge->add_column($this->config->item('table_name'), $fields);
+        }
+    }
+
+    public function down()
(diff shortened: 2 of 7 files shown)


Ron6519:
Thanks for the review. Fixed the valid points; checked each against the existing QRZ.com implementation.

**Fixed in `push_qso_to_qrzcall()`**
- `curl_close()` moved after the `curl_errno()` check, and the function now always returns a result (no more possible `null`) — matches `push_qso_to_qrz()`.
- User-Agent now uses the shared `cloudlog_user_agent()` helper instead of building it inline.
- Duplicate check tightened from a bare `duplicate` substring to `RESULT=FAIL` + `duplicate`.
- Corrected the stale comment (it referred to a bulk/retry path this PR does not have).

**Checked but left as-is — already consistent with QRZ.com**
- PAT field `type="text"`: the existing `qrzapikey`, `hrdlog_code` and `webadifapikey` fields in this same station-profile view are all `type="text"` — the QRZCALL.EU field matches that convention.
- `exists_qrzcall_api_key()`: identical pattern to the existing `exists_qrz_api_key()` (returns the row; callers check the key) — kept consistent.
- Status `'M'` on edit: QRZ.com's `'M'` feeds its bulk re-upload queue; this integration is real-time only, so an edit re-uploads immediately with `OPTION=REPLACE` and there is no `'M'` queue to populate.

Verified on the Docker dev environment: real-time create, edit (REPLACE) and duplicate handling all confirmed.

magicbug:
Thanks for putting this together.

Same as the callbook PR — I'm not going to add QRZCALL.EU as an upload target at this point. It's extra schema, station profile UI and realtime hooks to maintain, and there's no user demand for it yet.

Closing this one too. ADIF export is still there if people want to get QSOs into QRZCALL.EU in the meantime.

